PR #3, opened 2023-03-28
Use the new JAX+pip wheel workflow

nouiz:


yhtang:
Let's get back to this PR in a few days when the build system gets more stabilized. A lot of things are getting changed right at the moment 😄

.github/container/Dockerfile.jax
```diff
@@ -1,10 +1,10 @@
 ARG BASE_IMAGE
 FROM ${BASE_IMAGE}
 
-RUN apt-get update && apt-get install -y git python-is-python3 python3-pip libcudnn8 libcudnn8-dev
+RUN apt-get update && apt-get install -y git python-is-python3 python3-pip
 ENV HOME=/root
 RUN git clone https://github.com/google/jax.git /opt/jax
-RUN cd /opt/jax && pip install --upgrade -e ".[cuda]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
+RUN cd /opt/jax && pip install --upgrade -e ".[cuda12_pip]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
```
yhtang, line 7: Is it better if we install the pip wheels in the [base image](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/container/Dockerfile.base)?

mjsML: I think those are two separate things to test (testing the build process AND testing the upstream CUDA pipwheel ... both are important and should be done).

mjsML: @nouiz based on our discussion yesterday, I'll close this PR and we can work on standing up the PIP wheel testing flow later.

PR #7, opened 2023-04-18
Add Pax dockerfile

ashors1:
This PR adds a dockerfile which builds Paxml and Praxis from source.

### DIFF
--- a/.github/container/install-pax.sh
+++ b/.github/container/install-pax.sh
@@ -5,15 +5,15 @@
 usage() {
     echo "Usage: $0 [OPTION]..."
     echo "  -d, --dir=PATH         Path to store Pax source. Defaults to /opt"
-    echo "  -f, --from_paxml=URL   URL of the Paxml repo. Defaults to https://github.com/google-research/paxml.git"
-    echo "  -g, --from_praxis=URL  URL of the Praxis repo. Defaults to https://github.com/google-research/praxis.git"
+    echo "  --from_paxml=URL       URL of the Paxml repo. Defaults to https://github.com/google-research/paxml.git"
+    echo "  --from_praxis=URL      URL of the Praxis repo. Defaults to https://github.com/google-research/praxis.git"
     echo "  -h, --help             Print usage."
-    echo "  -r, --ref_paxml=REF    Git commit hash or tag name that specifies the version of Paxml to install. Defaults to HEAD."
-    echo "  -q, --ref_praxis=REF   Git commit hash or tag name that specifies the version of Praxis to install. Defaults to HEAD."
+    echo "  --ref_paxml=REF        Git commit hash or tag name that specifies the version of Paxml to install. Defaults to HEAD."
+    echo "  --ref_praxis=REF       Git commit hash or tag name that specifies the version of Praxis to install. Defaults to HEAD."
     exit $1
 }
 
-args=$(getopt -o d:f:hr: --long dir:,from:,help,ref: -- "$@")
+args=$(getopt -o d:h --long dir:,from_paxml:,from_praxis:,help,ref_paxml:,ref_praxis: -- "$@")
 if [[ $? -ne 0 ]]; then
     exit $1
 fi
@@ -25,22 +25,22 @@ while [ : ]; do
         INSTALL_DIR="$2"
         shift 2
         ;;
-    -f | --from_paxml)
+    --from_paxml)
         PAXML_REPO="$2"
         shift 2
         ;;
-    -g | --from_praxis)
+    --from_praxis)
         PRAXIS_REPO="$2"
         shift 2
         ;;
     -h | --help)
         usage
         ;;
-    -r | --ref_paxml)
+    --ref_paxml)
         PAXML_REF="$2"
         shift 2
         ;;
-    -q | --ref_praxis)
+    --ref_praxis)
         PRAXIS_REF="$2"
         shift 2
         ;;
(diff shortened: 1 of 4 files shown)


.github/container/install-pax.sh
```diff
@@ -0,0 +1,100 @@
...
+cd ${INSTALL_DIR}/paxml
+git checkout ${PAXML_REF}
+pip install -e .[gpu]
+
+apt-get autoremove -y
+apt-get clean
+rm -rf /var/lib/apt/lists/*
+rm -rf ~/.cache/pip/
+
+export PYTHONPATH=${INSTALL_DIR}/paxml:${INSTALL_DIR}/praxis:$PYTHONPATH
```
yhtang, line 100: To let an environment variable persists in the final docker image, it needs to be defined via the Dockerfile's ENV directive.

.github/container/install-pax.sh
```diff
@@ -0,0 +1,100 @@
...
+
+git clone ${PRAXIS_REPO} ${INSTALL_DIR}/praxis
+cd ${INSTALL_DIR}/praxis
+git checkout ${PRAXIS_REF}
+pip install -e .
+
+## Install Paxml
+
+git clone ${PAXML_REPO} ${INSTALL_DIR}/paxml
+cd ${INSTALL_DIR}/paxml
```
yhtang, line 91: Please use `pushd` and `popd` for more robust control of the working path.

.github/container/install-pax.sh
```diff
@@ -0,0 +1,100 @@
...
+apt-get install -y \
+    build-essential \
+    cmake \
+    clang \
+    git
+
+## Install Praxis
+
+git clone ${PRAXIS_REPO} ${INSTALL_DIR}/praxis
+cd ${INSTALL_DIR}/praxis
```
yhtang, line 84: Please use `pushd` and `popd` for more robust control of the working path.

.github/container/install-pax.sh
```diff
@@ -0,0 +1,100 @@
...
+
+echo "Installing Paxml $PAXML_REF from $PAXML_REPO and $PRAXIS_REF from $PRAXIS_REPO to $INSTALL_DIR"
+
+set -ex
+
+## Install dependencies
+
+apt-get update
+apt-get upgrade -y
+apt-get install -y \
```
yhtang, line 75: This appears to be inherited from the T5X installer script. Are these same packages all necessary for paxml and praxis?

.github/container/install-pax.sh
```diff
@@ -0,0 +1,100 @@
...
+    echo "  -d, --dir=PATH         Path to store Pax source. Defaults to /opt"
+    echo "  -f, --from_paxml=URL   URL of the Paxml repo. Defaults to https://github.com/google-research/paxml.git"
+    echo "  -g, --from_praxis=URL  URL of the Praxis repo. Defaults to https://github.com/google-research/praxis.git"
+    echo "  -h, --help             Print usage."
+    echo "  -r, --ref_paxml=REF    Git commit hash or tag name that specifies the version of Paxml to install. Defaults to HEAD."
+    echo "  -q, --ref_praxis=REF   Git commit hash or tag name that specifies the version of Praxis to install. Defaults to HEAD."
+    exit $1
+}
+
+args=$(getopt -o d:f:hr: --long dir:,from:,help,ref: -- "$@")
```
yhtang, line 16: This needs to be updated in accordance with the usage info above. For example, for the `-d, --dir` option, we need to add `-o d:` and `--long dir:`. The trailing colon is only added if an option takes an argument.

.github/container/install-pax.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
+
+## Parse command-line arguments
+
+usage() {
+    echo "Usage: $0 [OPTION]..."
+    echo "  -d, --dir=PATH         Path to store Pax source. Defaults to /opt"
+    echo "  -f, --from_paxml=URL   URL of the Paxml repo. Defaults to https://github.com/google-research/paxml.git"
+    echo "  -g, --from_praxis=URL  URL of the Praxis repo. Defaults to https://github.com/google-research/praxis.git"
```
yhtang, line 9: It's better if we use short option letters that are related to the long options. For example, `-h` for `--help`. If there is no good candidate for a short option, it is OK to only define the long option.

yhtang (changes requested):

yhtang:
Good job! Please also add
- a `_build_paxml.yaml` file modeled after: https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/_build_t5x.yaml
- a `nightly-paxml-build.yaml` file modeled after: https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/nightly-t5x-build.yaml
Given the similarity between the Dockerfiles, I expect most of the work should be as simple as changing the various occurrences of T5X to paxml.

.github/container/install-pax.sh
yhtang: And we shouldn't need to set it anyway if the packages were installed correctly.

.github/container/install-pax.sh
ashors1: Done. Please let me know my changes are not as expected

.github/container/install-pax.sh
ashors1: No, sorry. I have updated the dependencies in my latest commit

yhtang (approved):

yhtang:
Looks good to me now. @nluehr Do you have any thoughts before I merge it?

nluehr:
LGTM as well. Merging.

PR #12, opened 2023-04-24
Nightly build for the Transformer Engine container

yhtang:
Largely modeled after the T5X build process.

### DIFF
--- a/.github/container/Dockerfile.te
+++ b/.github/container/Dockerfile.te
@@ -1,0 +1,12 @@
+###############################################################################
+## T5X
+###############################################################################
+
+ARG BASE_IMAGE=ghcr.io/nvidia/jax:latest
+FROM ${BASE_IMAGE}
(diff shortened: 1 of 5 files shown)


yhtang:
This one is indeed the TE-only container.
However, having an array of TE+X containers where X \in {T5X, PAX, ...} in addition to the current container array sounds bloating. How about:
1. add TE to each  of the existing JAX/T5X/PAX container, or
2. create an all-in-one container with JAX+TE+T5X+PAX+...

yhtang:
@nouiz I'll go ahead and merge this one just to move faster. Please feel free to comment on this PR and/or the files any time and I'll address them in subsequent PRs. 😄

nouiz:
I would go with 1.
2 can end up being problematic, T5X and PAX would have different version requirement...
This happens frequently in other cases, so I prefer to plan as this is going to happen.

PR #24, opened 2023-04-26
Add First Pax Test

ashors1:
This PR adds a single-GPU nightly Pax test.

### DIFF
--- a/.github/container/test-pax.sh
+++ b/.github/container/test-pax.sh
@@ -197,6 +197,8 @@ class GPT126MPP(TransformerLmSpmdPipelineAdam):
     task_p = super().task()
     task_p = configure_gpt3_task(self, task_p)
 
+    task_p.train.num_train_steps = self.MAX_STEPS
+
     model_p = task_p.model
     
     ### compute layernorm reductions in fp32. Needed for stable training on GPUs
@@ -259,5 +261,8 @@ python -m paxml.main \
     --alsologtostderr \
     $([[ $MULTIPROCESS != 0 ]] && echo --multiprocess_gpu)
     
+
+rm -r ${OUTPUT}/checkpoints
+
 set +x
 echo "Output at ${OUTPUT}"
--- a/.github/container/test-pax.sh
+++ b/.github/container/test-pax.sh
@@ -1,0 +1,264 @@
+#!/bin/bash
+
+# # Parse command-line arguments
+
+print_var() {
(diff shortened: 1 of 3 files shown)


.github/container/test-pax.sh
```diff
@@ -0,0 +1,264 @@
...
+args=$(getopt -o b:s:o:h --long batch-per-gpu:,dtype:,steps:,help,multiprocess,output:,data-parallel:,tensor-parallel:,pipeline-parallel: -- "$@")
+if [[ $? -ne 0 ]]; then
+    exit $1
+fi
+
+# Default arguments
+MULTIPROCESS=0
+OUTPUT=$(mktemp -d)
+
+export BATCH_PER_GPU=4
```
yhtang, line 36: It is not necessary to `export` the variables for use in the heredoc (`EOF ... EOF`). We can directly use variable substitutions (`${VARIABLE}`).

.github/container/test-pax.sh
```diff
@@ -0,0 +1,264 @@
...
+import os
+import math
+from paxml import tasks_lib, experiment_registry
+from paxml.contrib.gpu.scripts_gpu.configs import GPT126M, configure_gpt3_task
+from paxml.tasks.lm.params.c4 import TransformerLmSpmdPipelineAdam
+from paxml.tasks.lm.params.lm_cloud import SyntheticDataset
+from praxis import base_layer
+from praxis import layers
+
+dp = int(os.getenv("DP"))
```
yhtang, line 126: Directly embedding the value using variable substitution may be better for archiving purposes. Otherwise, we won't know the value of `dp` without knowing the value of the environment variable `DP`, which is not readily available from this script.

yhtang (changes requested):

.github/container/test-pax.sh
ashors1: Done. Note that paxml expects `VOCAB_PATH` to be set as an environment variable, so we still export `VOCAB_PATH` in the script.

yhtang: This script will run on a GitHub Actions runner, which will submit a SLURM job to a remote cluster. Hence, the variable will not be available even if exported here.

yhtang: We will need to add an `#SBATCH --export=...` directive around line 61-67 to pass in this variable.

yhtang (approved):
Let's merge and see how it works in real life.

PR #26, opened 2023-04-27
cleaning up exit codes in scripts

terrykong:


yhtang (approved):
Thanks for the changes :)

PR #27, opened 2023-04-28
Install OFED for RoCE and IB support

yhtang:


.github/container/install-ofed.sh
```diff
@@ -0,0 +1,42 @@
...
+set -ex
+
+export DEBIAN_FRONTEND=noninteractive
+export TZ=America/Los_Angeles
+
+# Install libnl (Netlink Protocol Library Suite), which provides a low-level API
+# for communication between kernel and user space processes in Linux. Essential for managing
+# networking components such as routing tables, network interfaces, and address resolution.
+
+apt-get update
```
nouiz, line 12: We end up doing many apt-get update and the clean up in a full docker build.
This allow us to more easily split componentize the installation and this is great.
Any idea how much time we loose during the build due to that?

yhtang: I also thought about it recently. Since `apt-get update` only downloads the package list, I feel it is not adding too much time. However, one problem it introduces is a higher chance of build failure because the update process would occasionally hang indefinitely.

PR #33, opened 2023-05-02
Adds initial rosetta CI workflows - but steps are no-ops

terrykong:


.github/workflows/_build_rosetta.yaml
```diff
@@ -0,0 +1,60 @@
...
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set docker metadata
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: |
+            ${{ env.UPLD_IMAGE }}
```
yhtang, line 54: Since these are all single-line arguments, why not just make them denser by keeping them in-line?

.github/workflows/_test_rosetta.yaml
```diff
@@ -0,0 +1,42 @@
+name: ~test and lint Rosetta
+
+on:
+  # Manual trigger
+  workflow_dispatch:
+  # Called from another workflow
+  workflow_call:
+    inputs:
+      ROSETTA_IMAGE_TAG:
```
yhtang, line 9: I recently decided to use the complete image address rather than only the tag part for the workflow input. See [for example here](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/_test_jax.yaml). This will make it easier for us to switch between the public facing `nvidia/t5x` and internal `nvidia/jax-toolbox-internal` repos. Could you please update this input argument as in the `_test_jax.yaml` example?

.github/workflows/ci.yaml
```diff
@@ -59,4 +59,12 @@ jobs:
     uses: ./.github/workflows/_test_jax.yaml
     with:
       JAX_IMAGE: ${{ needs.build-jax.outputs.DOCKER_TAGS }}
-    secrets: inherit
\ No newline at end of file
+    secrets: inherit
+
+  rosetta:
```
yhtang, line 64: `rosetta` -> `build-rosetta`

.github/workflows/nightly-rosetta-test.yaml
```diff
@@ -0,0 +1,45 @@
+name: Nightly Rosetta test
+
+on:
+  workflow_run:
+    workflows: [Nightly Rosetta build]
+    types: [completed]
+  workflow_dispatch:
+    inputs:
+      ROSETTA_IMAGE_TAG:
```
yhtang, line 9: Same as above, we will migrate to use complete image URLs instead of only the tag part.

yhtang (changes requested):
LGTM, but please fix a few nitpicks. Thanks! 😄

.github/workflows/_build_rosetta.yaml
terrykong: done

.github/workflows/_test_rosetta.yaml
terrykong: makes sense. done

.github/workflows/ci.yaml
terrykong: done

.github/workflows/nightly-rosetta-test.yaml
terrykong: done

yhtang (approved):

PR #36, opened 2023-05-03
Add MGMN Pax Tests

ashors1:
This PR extends the Pax test suite to include multi-GPU and multi-node performance tests.


.github/container/test-pax.sh
```diff
@@ -90,7 +96,8 @@ done
 
 # # Set derived variables
 
-NGPUS=$(nvidia-smi -L | grep -c '^GPU')
+GPUS_PER_NODE=$(nvidia-smi -L | grep -c '^GPU')
+NGPUS=$(( $GPUS_PER_NODE * $NODES ))
```
yhtang, line 100: `$((GPUS_PER_NODE * NODES))` should do since we are already using `$((...))`. There is no need for the extra dollar signed in front of individual variables.

yhtang (changes requested):
Great job!

I noticed that the three jobs `single-process-single-gpu`, `single-process-single-gpu`, and `multi-gpu-multi-node` are nearly identical except for a few lines. Can we unify them to avoid duplicating the boilerplates?

For example, I see that the three jobs differ at the `SBATCH` directives:
```
  single-process-single-gpu:
          ...
          #SBATCH --nodes=1
          #SBATCH --gpus-per-node=1
          #SBATCH --tasks=1
          #SBATCH --tasks-per-node=1

  multi-gpu-single-node:
          ...
          #SBATCH --nodes=1
          #SBATCH --gpus-per-node=${{ steps.meta.outputs.TOTAL_TASKS }}
          #SBATCH --tasks=${{ steps.meta.outputs.TOTAL_TASKS }}
          #SBATCH --tasks-per-node=${{ steps.meta.outputs.TOTAL_TASKS }}

  multi-gpu-multi-node:
          ...
          #SBATCH --nodes=2
          #SBATCH --gpus-per-node=8
          #SBATCH --tasks=16
          #SBATCH --tasks-per-node=8
```
but it seems that the task counts etc. can be computed in a unified way in the `meta` job step.

The `test-pax.sh` also takes in slightly different arguments in the three jobs that could be handled using a few lines of `if-else`.

This should allow us to significantly reduce the size of this workflow file and help make it more maintainable. 🌻

.github/workflows/_test_pax.yaml
```diff
@@ -73,11 +87,13 @@ jobs:
...
+              --batch-per-gpu 4 \
               --steps 500 \
-              --data-parallel ${{ matrix.DP }} \
-              --tensor-parallel ${{ matrix.TP }} \
-              --pipeline-parallel ${{ matrix.PP }} 
+              --data-parallel ${{ matrix.PARALLEL_CONFIG[0] }} \
+              --tensor-parallel ${{ matrix.PARALLEL_CONFIG[1] }} \
+              --pipeline-parallel ${{ matrix.PARALLEL_CONFIG[2] }} \
+              --nodes ${{ steps.meta.outputs.NODES }} \
+              --multiprocess
```
yhtang, line 96: I'm not sure for Pax, but for T5X, the behavior is different with and without `--multiprocess` for single-node-multi-GPU runs. How does Pax behave here?

.github/workflows/_test_pax.yaml
```diff
@@ -41,11 +45,20 @@ jobs:
         id: meta
         shell: bash -x -e {0}
         run: |
-          TEST_CASE_NAME=1P${{ matrix.N_GPU }}G${{ matrix.DP }}DP${{ matrix.TP }}TP${{ matrix.PP }}PP
+          TEST_CASE_NAME=${{ matrix.PARALLEL_CONFIG[0] }}DP${{ matrix.PARALLEL_CONFIG[1] }}TP${{ matrix.PARALLEL_CONFIG[2] }}PP
+          TOTAL_TASKS=$((${{ matrix.PARALLEL_CONFIG[0] }} * ${{ matrix.PARALLEL_CONFIG[1] }} * ${{ matrix.PARALLEL_CONFIG[2] }}))
+          GPUS_PER_NODE=$TOTAL_TASKS
```
yhtang, line 50: How about consolidating lines 50-56 into
```
MAX_GPUS_PER_NODE=8
NODES=$((TOTAL_TASKS / MAX_GPUS_PER_NODE))
GPUS_PER_NODE=$((TOTAL_TASKS / NODES))
```
?

yhtang (changes requested):

.github/workflows/_test_pax.yaml
ashors1: Yes, the behavior is different for the two. I think it is desirable to do the single-gpu run without multiprocess. Would you prefer that I split that run into a separate job?

.github/workflows/_test_pax.yaml
ashors1: I don't think this will quite do what we want because `NODES` needs to be the ceiling of `TOTAL_TASKS / MAX_GPUS_PER_NODE`. But I think that `NODES=$((result=(TOTAL_TASKS+MAX_GPUS_PER_NODE-1)/MAX_GPUS_PER_NODE))` should suffice.

yhtang: Good catch 👍

.github/workflows/_test_pax.yaml
yhtang: We don't have to split the job. Something along the line `$([[ <SOME-CONDITION> ]] && echo --multiprocess)` should do.

yhtang (approved):
Ready to go :)

yhtang:
Hold on. There is a degeneracy of test case names. The [latest run](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/4878352284) ended up creating 6 artifacts out of 7 jobs. We need to encode the test names properly to reflect the difference between single-process and multi-process runs.

yhtang (approved):
LGTM

PR #43, opened 2023-05-03
[rosetta] Adds CI steps for building and testing rosetta

terrykong:


.github/workflows/_test_rosetta.yaml
```diff
@@ -14,16 +14,52 @@ on:
...
+
+      - name: Pull Rosetta image
+        shell: bash -x -e {0}
+        run: |
+          docker pull ${{ inputs.ROSETTA_IMAGE }}
+
+      - name: Run Rosetta tests w/ docker
+        run: |
+          set -x
+          docker run --gpus '"device=0,1"' ${{ inputs.ROSETTA_IMAGE }} sh -c "pip install '/opt/rosetta[test]' && pytest /opt/rosetta ${{ matrix.MARKERS }}"
```
yhtang, line 45: Is there a particular reason for only using 2 GPUs? Preferably, we don't hardcode the number of GPUs since the compute nodes may have a variable number of GPUs available.

.github/workflows/_test_rosetta.yaml
```diff
@@ -14,16 +14,52 @@ on:
...
+
   rosetta-lint-check:
```
yhtang, line 47: It just occurred to me that we should lint Rosetta before even the build, right?

yhtang: We can create a separate reusable workflow just for linting either rosetta or even the entire repo, and call it at the beginning of the entire CI pipeline.

yhtang (changes requested):

.github/workflows/_test_rosetta.yaml
yhtang: How about we leave linting out of the scope of this PR, and do the whole-repo linting via another PR?

terrykong: > It just occurred to me that we should lint Rosetta before even the build, right?

Good point, it could be good to do it before since we'd avoid wasting compute on improperly formatted code. I originally created it as a parallel step since I would imagine we'd be interested in the functionality even if the code wasn't to our style guidelines.

I can remove it for now since it's not a hard change to re-introduce (and there's nothing substantive to lint yet anyways 😄)

yhtang: Sounds good. We can enforce style guide later.

.github/workflows/_test_rosetta.yaml
terrykong: So my assumption here was that the tests could always assume there were 2 GPUs so that we wouldn't have to create additional logic to select the right devices. The example in mind, was if someone pjitt-ed a function for DP, the batch size would need to be divisible by the number of devices. So if the tests had arrays with BS=2, we'd be okay if num_devices<=2, but if num_devices>2 we'd error.

My understanding was that after jax was imported, the "visible devices" was fixed, so to run a test on a subset of GPUs, we'd need to run in a subprocess. Is there a way to make devices after jax has been imported? If so, I agree with what you're suggesting.

.github/workflows/_test_rosetta.yaml
```diff
@@ -14,16 +14,52 @@ on:
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Pull Rosetta image
+        shell: bash -x -e {0}
+        run: |
+          docker pull ${{ inputs.ROSETTA_IMAGE }}
+
+      - name: Run Rosetta tests w/ docker
+        run: |
+          set -x
```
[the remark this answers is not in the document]
terrykong: Good point. Done!

.github/workflows/_test_rosetta.yaml
terrykong: Wellll, right now some tests do have hard coded batch sizes, but I suppose they could be refactored so that:
1. odd number of devices > 1 throws an error and
2. batchsize should be a multiple of 8

and then we should be robust to most common configurations.

I can remove the explicit device specification for now since there aren't any "real" tests anyways, but does the above sound good?

yhtang: Sounds good.

terrykong: Done

yhtang (approved):

yhtang:
Done! 🎆

PR #53, opened 2023-05-08
Allow Rosetta's distribution creation script to target multiple upstream/base projects

terrykong:
Change should make it easier to base rosetta on a different base library


yhtang:
Could you please share a bit about what is 'library independence'?

terrykong:
> Could you please share a bit about what is 'library independence'?

The change here was to make it so that the `create-distribution.sh` script (formerly `create-t5x-distribution.sh`) could be used to create a different "build" of rosetta. In the future, when someone needs to build a "paxml" or other build of rosetta, they should be able to re-use that script now and just provide different args/patchlist.

yhtang:
Now I understand that. It required a bit of change in the mindset 😄

terrykong:
> Now I understand that. It required a bit of change in the mindset 😄

Sorry, I think the phrasing could've been better 😅

rosetta/create-distribution.sh
```diff
@@ -45,27 +56,26 @@ done
...
+INSTALLED_DIR=${INSTALLED_DIR:-/opt/t5x}
 DISTRIBUTION_BASE_REF=${DISTRIBUTION_BASE_REF:-main}
-INTERNAL_T5X_SUBMODULE_DIR=$SCRIPT_DIR/t5x
-UPSTREAM_GIT_URL=https://github.com/google-research/t5x.git
+UPSTREAM_GIT_URL=${UPSTREAM_GIT_URL:-https://github.com/google-research/t5x.git}
 
-if [[ -z "$INSTALLED_T5X_DIR" ]]; then
-  echo "[ERROR]: Need to specify the env INSTALLED_T5X_DIR="
+if [[ -z "$INSTALLED_DIR" ]]; then
+  echo "[ERROR]: Need to specify the env INSTALLED_DIR="
```
yhtang, line 70: Since we have the `-d/--dir` option for specifying the installation dir, we could print the error message plus the usage guide here.

rosetta/create-distribution.sh
```diff
@@ -74,25 +84,26 @@ git branch --force distribution-base $DISTRIBUTION_BASE_REF
...
-  git -C $INTERNAL_T5X_SUBMODULE_DIR "$@"
-}
-NV_REMOTE_NAME=nv-t5x
-if [[ -d $INTERNAL_T5X_SUBMODULE_DIR ]]; then
-  for remote_branch in $(git-nv-t5x branch --remotes | grep -v 'origin/HEAD' | egrep -v 'origin/test_[0-9]+'); do
-    # Try creating a local tracking branch, but if already exists, then update it to match remote
-    git-nv-t5x checkout --track $remote_branch || git-nv-t5x reset --hard $remote_branch
+if [[ -n "${EXTRA_MIRROR_DIR+x}" ]] && [[ -d $EXTRA_MIRROR_DIR ]]; then
+  EXTRA_REMOTE_NAME=extra
+  function git-extra {
```
yhtang, line 94: Since this command appends an additional mirror to the git command, would `git+mirror` be more intuitive?

rosetta/create-distribution.sh
```diff
@@ -149,8 +157,11 @@ EOF
 done
 
 # Cleanup
-for remote in $UPSTREAM_REMOTE_NAME $NV_REMOTE_NAME; do
+for remote in $UPSTREAM_REMOTE_NAME ${EXTRA_REMOTE_NAME:-}; do
```
yhtang, line 160: Would `${EXTRA_REMOTE_NAME:-}` behave the same as `${EXTRA_REMOTE_NAME}` here?

yhtang (changes requested):

rosetta/Dockerfile.t5x
```diff
@@ -3,9 +3,12 @@ FROM ${BASE_IMAGE} AS distribution-builder
...
 COPY . /opt/rosetta
 WORKDIR /opt/rosetta
-RUN bash create-t5x-distribution.sh -p patchlist.txt -d /opt/t5x
+RUN bash create-distribution.sh \
+  -p patchlist-t5x.txt \
+  -u https://github.com/google-research/t5x.git \
+  -d /opt/t5x
 # TODO(terry): revisit whether any of these are needed
-RUN rm -rf .git $(find /opt -name "__pycache__") #&& git submodule deinit -f t5x
+RUN rm -rf .git $(find /opt -name "__pycache__")
```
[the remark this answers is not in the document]
yhtang: Thanks for the offline communication. Now I realize that the layers will get flattened later via
```
COPY --from=distribution-builder /opt/t5x /opt/t5x
COPY --from=distribution-builder /opt/rosetta /opt/rosetta
```
So we _can_ reduce final image size this way.

yhtang (approved):
Good to go 😄

PR #54, opened 2023-05-09
Adds CI for unit testing TE build

terrykong:
TE has some integration tests that we can run, but first wanted to add unit tests and then will address integration test in follow up PR. Couldn't create a badge like the jax tests do since we need to create a new gist for it. Can address the badge in a follow up PR if preferred

### DIFF
--- a/.github/workflows/_test_te.yaml
+++ b/.github/workflows/_test_te.yaml
@@ -33,7 +33,7 @@ jobs:
           docker pull ${{ inputs.JAX_TE_IMAGE }}
 
       - name: Run JAX-TE unit tests w/ docker
-        shell: docker run -v /tmp/pytest-log:/tmp/pytest-log --gpus all ${{ inputs.JAX_TE_IMAGE }} bash -x {0}
+        shell: docker run -v /tmp/pytest-log:/tmp/pytest-log --gpus all ${{ inputs.JAX_TE_IMAGE }} bash -x -c "$(cat {0})"
         run: |
           pip install pytest pytest-reportlog cuda-python
           pytest --report-log=/tmp/pytest-log/report.jsonl /opt/transformer-engine/tests/jax
--- a/.github/workflows/_test_te.yaml
+++ b/.github/workflows/_test_te.yaml
@@ -33,9 +33,10 @@ jobs:
           docker pull ${{ inputs.JAX_TE_IMAGE }}
 
       - name: Run JAX-TE unit tests w/ docker
-        shell: bash -x {0}
+        shell: docker run -v /tmp/pytest-log:/tmp/pytest-log --gpus all ${{ inputs.JAX_TE_IMAGE }} bash -x {0}
         run: |
-          docker run -v /tmp/pytest-log:/tmp/pytest-log --gpus all ${{ inputs.JAX_TE_IMAGE }} sh -c "pip install pytest pytest-reportlog cuda-python && pytest --report-log=/tmp/pytest-log/report.jsonl /opt/transformer-engine/tests/jax"
+          pip install pytest pytest-reportlog cuda-python
+          pytest --report-log=/tmp/pytest-log/report.jsonl /opt/transformer-engine/tests/jax
 
       - name: Upload test json logs
         uses: actions/upload-artifact@v3
--- a/.github/workflows/_test_te.yaml
+++ b/.github/workflows/_test_te.yaml
@@ -1,8 +1,6 @@
 name: ~test TransformerEngine
 
 on:
-  # Manual trigger
-  workflow_dispatch:
   # Called from another workflow
   workflow_call:
     inputs:
--- a/.github/workflows/_test_te.yaml
+++ b/.github/workflows/_test_te.yaml
@@ -1,0 +1,47 @@
+name: ~test TransformerEngine
+
+on:
+  # Manual trigger
+  workflow_dispatch:
+  # Called from another workflow
+  workflow_call:
+    inputs:
+      JAX_TE_IMAGE:
+        type: string
+        description: 'JAX-TE image build by NVIDIA/JAX-Toolbox'
+        required: true
+        default: 'ghcr.io/nvidia/jax-te:latest'
+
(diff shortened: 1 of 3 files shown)


yhtang:
> Can you confirm this only install the TE/JAX or does it install all TE back-end?

It's using the jax-te container that we build in this repo and only installs the JAX backend as shown here: https://github.com/NVIDIA/JAX-Toolbox/blob/b73c334aa813b9a923670fb49794034fa051e052/.github/container/install-te.sh#L70

terrykong:
> Can you confirm this only install the TE/JAX or does it install all TE back-end?

What @yhtang said + just sanity checking:

```bash
root@adbe279260ab:/opt/transformer-engine# ls -lah /opt/transformer-engine/build/**/*.so
-rwxr-xr-x 1 root root 204M May  1 18:07 /opt/transformer-engine/build/lib.linux-x86_64-3.10/libtransformer_engine.so
-rwxr-xr-x 1 root root 221K May  1 18:07 /opt/transformer-engine/build/lib.linux-x86_64-3.10/transformer_engine_jax.cpython-310-x86_64-linux-gnu.so
```
So this looks like it was only installed w/ jax support

terrykong:
Related issue: https://github.com/NVIDIA/TransformerEngine/issues/156

.github/workflows/_test_te.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Pull JAX-TE image
+        shell: bash -x -e {0}
+        run: |
+          docker pull ${{ inputs.JAX_TE_IMAGE }}
+
+      - name: Run JAX-TE unit tests w/ docker
+        shell: bash -x {0}
+        run: |
+          docker run -v /tmp/pytest-log:/tmp/pytest-log --gpus all ${{ inputs.JAX_TE_IMAGE }} sh -c "pip install pytest pytest-reportlog cuda-python && pytest --report-log=/tmp/pytest-log/report.jsonl /opt/transformer-engine/tests/jax"
```
yhtang, line 40: 👍 Given the richness of the JSON output, this should make it easier now to add a markdown table of all the failed tests to the [job summary](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) via a subsequent job step.

yhtang: I'd suggest using `docker run --name=<NAME>` to split the long command line into multiple lines for better readability.

.github/workflows/nightly-te-test.yaml
```diff
@@ -0,0 +1,105 @@
...
+          cat > endpoint.json << EOF
+          {
+            "schemaVersion": 1,
+            "label": "V100 Unit",
+            "message": "${PASSED_TESTS}/${SKIPPED_TESTS}/${FAILED_TESTS} pass/skip/fail",
+            "color": "${BADGE_COLOR}"
+          }
+          EOF
+
+      # TODO: re-enable after secret/gist_id is created for TE badge json
```
yhtang, line 86: We can use the same gist as a single gist can contain multiple files. We just need to rename the json to `te-unit-test-status.json`, and the JAX one can become `jax-unit-test-status.json`.

yhtang (changes requested):
Great work :) I only have some thoughts on small things.

.github/workflows/_test_te.yaml
terrykong: > 👍 Given the richness of the JSON output, this should make it easier now to add a markdown table of all the failed tests to the [job summary](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) via a subsequent job step.

That's a pretty cool feature!


> I'd suggest using `docker run --name=<NAME>` to split the long command line into multiple lines for better readability.

Maybe I'm missing something, but `--name` just makes it easier to find your container when you `docker ps`. Is your suggestion just to break up the shell command for readability, so something like this?

```yaml
- name: Run JAX-TE unit tests w/ docker
        shell: docker run -v /tmp/pytest-log:/tmp/pytest-log --gpus all ${{ inputs.JAX_TE_IMAGE }} sh {0}
        run: |
          pip install pytest pytest-reportlog cuda-python
          pytest --report-log=/tmp/pytest-log/report.jsonl /opt/transformer-engine/tests/jax"
```

.github/workflows/nightly-te-test.yaml
terrykong: Nice~ Easier than I thought :) 

Anything using pytest can probably re-use this badge publishing workflow (e.g., rosetta). I can address in a followup PR

.github/workflows/_test_te.yaml
terrykong: Actually on second thought, my code above probably wouldn't work b/c `{0}` would be replaced by a filepath that doesn't exist within that container, so maybe something like:

```yaml
- name: Run JAX-TE unit tests w/ docker
  shell: docker run -v /tmp/pytest-log:/tmp/pytest-log --gpus all ${{ inputs.JAX_TE_IMAGE }} bash -x -c "$(cat {0})"
  run: |
    pip install pytest pytest-reportlog cuda-python
    pytest --report-log=/tmp/pytest-log/report.jsonl /opt/transformer-engine/tests/jax
```

.github/workflows/_test_te.yaml
```diff
@@ -0,0 +1,47 @@
+name: ~test TransformerEngine
+
+on:
+  # Manual trigger
+  workflow_dispatch:
```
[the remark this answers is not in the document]
terrykong: Done

.github/workflows/_test_te.yaml
yhtang: Very cool. I didn't realize we could use the `shell:` key to solve this problem.
But I wonder if "$(`cat {0}`)" could destroy certain features in the script, such as heredoc (`EOF...EOF`) and functions. I recall that GitLab has line-ending issues when defining functions in the script section.

yhtang (approved):

PR #57, opened 2023-05-11
Made JAX-TE unit test fully functional

yhtang:


.github/workflows/_test_te.yaml
```diff
@@ -31,15 +31,17 @@ jobs:
         shell: bash -x -e {0}
         run: |
           docker pull ${{ inputs.JAX_TE_IMAGE }}
+          docker tag ${{ inputs.JAX_TE_IMAGE }} jax:te
 
-      - name: Run JAX-TE unit tests w/ docker
-        shell: docker run -v /tmp/pytest-log:/tmp/pytest-log --gpus all ${{ inputs.JAX_TE_IMAGE }} bash -x -c "$(cat {0})"
+      - name: Run JAX-TE unit tests with docker
+        shell: docker run --gpus all -v {0}:/cmd.sh -v /log:/log jax:te bash -x /cmd.sh
```
terrykong, line 37: That's quite clever! That also looks like it gets around the former impl, which probably wouldn't have worked for heredocs.

terrykong (approved):

PR #59, opened 2023-05-12
Adds TE Integration tests (and badge) + Refactors pytest badge creation to reusable workflow

terrykong:


yhtang:
Can we make `.github/workflows/_publish_pytest_badge.yaml` reusable by both JAX and TE to publish results? We can even reuse it for T5X/PAX perf result. We should be able to achieve that if we turn the postprocessing script and the endpoint file name into inputs of the workflow.

Essentially we can create something like `.github/workflows/_publish_badge.yaml` that can be used like:
```
  publish:
    with:
      BADGE_LABEL: 'V100 unit'
      BADGE_FILENAME: 'te-unit-test-status.json'
      SCRIPT: ...
      ARTIFACTS: ...
      PUBLISH: ${{ inputs.PUBLISH }} # do we still need this?
```

terrykong:
> Can we make `.github/workflows/_publish_pytest_badge.yaml` reusable by both JAX and TE to publish results? We can even reuse it for T5X/PAX perf result. We should be able to achieve that if we turn the postprocessing script and the endpoint file name into inputs of the workflow.
> 
> Essentially we can create something like `.github/workflows/_publish_badge.yaml` that can be used like:
> 
> ```
>   publish:
>     uses: ./.github/workflows/_publish_badge.yaml
>     with:
>       BADGE_LABEL: 'V100 unit'
>       BADGE_FILENAME: 'te-unit-test-status.json'
>       SCRIPT: ...
>       ARTIFACTS: ...
>       PUBLISH: ${{ inputs.PUBLISH }} # do we still need this?
> ```

That makes sense. What do you think about me addressing this in a follow up PR to separate the changes? 

Regarding: `PUBLISH: ${{ inputs.PUBLISH }} # do we still need this?`
I forwarded the "PUBLISH" input to the reusable workflow b/c I wanted the publish job to run so I could inspect the script/failures, but not update the gist.

yhtang (approved):

PR #61, opened 2023-05-15
Parametrizes badge creation for jax and TE

terrykong:
Addressing this comment: https://github.com/NVIDIA/JAX-Toolbox/pull/59#issuecomment-1546188611 

This PR parametrizes the badge creation for TE and jax

### DIFF
--- a/.github/workflows/_publish_badge.yaml
+++ b/.github/workflows/_publish_badge.yaml
@@ -1,8 +1,13 @@
-name: ~publish pytest results into a json file that can be used to make badge
+name: ~publish test results into a json file that can be used to make badge
 
 on:
   workflow_call:
     inputs:
+      SCRIPT_PATH:
+        type: string
+        description: 'Path to script relative to project root:'
+        required: true
+        default: '.github/workflows/scripts/pytest_log_badge.sh'
       BADGE_LABEL:
         type: string
         description: 'Badge label:'
@@ -11,9 +16,9 @@ on:
         type: string
         description: 'Filename to find the json in the gist for creating a badge:'
         required: true
-      LOG_GLOB:
+      ARTIFACTS:
         type: string
-        description: '(Recursive) glob to match pytest report logs assuming artifacts are downloaded in CWD:'
+        description: 'Paths of artifacts assuming they are all downloaded (globs can be used, but it is up to script to support):'
         required: true
         default: '**/*.jsonl'
       PUBLISH:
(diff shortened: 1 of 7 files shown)


yhtang:
This is generally in the right direction, but I feel the addition of `jax_test_unit_badge.sh` and `pytest_log_badge.sh` is a bit overkill here. Do you think we could simply embed the script in the `SCRIPT` input to the reusable workflow?

terrykong:
> This is generally in the right direction, but I feel the addition of `jax_test_unit_badge.sh` and `pytest_log_badge.sh` is a bit overkill here. Do you think we could simply embed the script in the `SCRIPT` input to the reusable workflow?

I see what you mean. Yea, we can do that. I will refactor

.github/workflows/_publish_badge.yaml
```diff
@@ -0,0 +1,77 @@
...
+    steps:
+      - name: Check out the repository under ${GITHUB_WORKSPACE}
+        uses: actions/checkout@v3
+
+      - name: Download all artifacts from the previous jobs
+        uses: actions/download-artifact@v3
+        with:
+          path: ${{ env.LOCAL_TEST_LOG_DIR }}
+
+      - name: Count the number of failed tests
```
yhtang, line 44: The name probably needs an update 😄

.github/workflows/_publish_badge.yaml
```diff
@@ -0,0 +1,77 @@
...
+        required: true
+      BADGE_LABEL:
+        type: string
+        description: 'Badge label:'
+        required: true
+      BADGE_FILENAME:
+        type: string
+        description: 'Filename to find the json in the gist for creating a badge:'
+        required: true
+      ARTIFACTS:
```
yhtang, line 18: I feel it is unnecessary to pass the artifacts using an input since we are currently downloading all job artifacts anyway. 
If we keep it as an input, we should use it in the download-artifact step to only download select artifacts.

The default value for it also seems to be specialized for pytest.

.github/workflows/_publish_badge.yaml
```diff
@@ -0,0 +1,77 @@
...
+        required: true
+        default: '**/*.jsnl'
+      PUBLISH:
+        type: boolean
+        description: Update status badge?
+        default: false
+        required: true
+
+env:
+  LOCAL_TEST_LOG_DIR: .test-logs
```
yhtang, line 30: Since it is a standalone (albeit reusable) workflow, it always starts in a clean docker container so we can simply use the current working folder to store all the downloaded artifacts. There is no risk of interference or clutter with previous jobs.

.github/workflows/_test_jax.yaml
```diff
@@ -8,11 +8,20 @@ on:
...
+
+env:
+  UNIT_TEST_ARTIFACT_NAME: unit-test-logs
 
 jobs:
 
   unit-test:
     runs-on: [self-hosted, compute, V100]
+    outputs:
+      UNIT_TEST_ARTIFACT_NAME: ${{ env.UNIT_TEST_ARTIFACT_NAME }}
```
yhtang, line 24: `ARTIFACT_NAME` should suffice since the workflow/jobs already has 'test' in their names.

For example, if we use it in a later job, it will be
```
needs.unit-tests.outputs.ARTIFACT_NAME
```

yhtang (changes requested):

yhtang (changes requested):

.github/workflows/_publish_badge.yaml
terrykong: > why not declare a variable ARTIFACT_PATTERN in the SCRIPT section and consume it locally?


That's a good point; not much utility added there. I will make that change

PR #62, opened 2023-05-15
Add single-process jax examples

nouiz:
I didn't add the multi-process scripts.
I suppose there is multiple GPUs on the test node.

### DIFF
--- a/.github/workflows/_test_te.yaml
+++ b/.github/workflows/_test_te.yaml
@@ -39,6 +39,12 @@ jobs:
           pip install pytest pytest-reportlog cuda-python
           TE_PATH=$(dirname $(python -c "import transformer_engine as te; print(*te.__path__)"))
           pytest --report-log=/log/report.jsonl ${TE_PATH}/tests/jax || true
+          pip install -r ${TE_PATH}/examples/jax/mnist/requirements.txt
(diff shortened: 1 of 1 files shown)


terrykong:
Hey @nouiz . I opened this [PR](https://github.com/NVIDIA/JAX-Toolbox/pull/59/files#diff-0ca60f47d99fba066bb85abe3d3922a77a48ba00de7ba1ffea5b98ab58bcb14cR126-R128) earlier for adding some of those tests, but running with pytest (+pytest-reportlog) so that way we can use the report log to extract timing info in the future.

Also FYI, the job you are adding the tests to run on a V100, which IIUC shouldn't be able to run fp8

yhtang:
Once the new OCI compartments are created, we will be then able to provision different types of GPU runners (V100, A100, etc.). I'll refactor our workflow then to accommodate the matricization of jobs across hardware types.

yhtang:
Meanwhile, @nouiz @terrykong could you please coordinate on the test items? Instead of packing every test into the current `te-unit-tests` job, we could create multiple jobs inside the `_test_te.yaml` workflow to help parallelize and parameterize the tests across hardware types, etc.

nouiz:
@terrykong I let you go with your PR.

PR #64, opened 2023-05-16
Allow customized container build in the main CI workflow

yhtang:
In addition and as a by-product, this PR properly sets the GitHub container registry token for our A100 cluster when running CI jobs.


.github/workflows/_test_pax.yaml
```diff
@@ -3,9 +3,9 @@ name: ~test Pax, MGMN
 on:
   workflow_call:
     inputs:
-      PAX_IMAGE_TAG:
+      PAX_IMAGE:
         type: string
-        description: Tag of image from https://github.com/NVIDIA/JAX-Toolbox/pkgs/container/pax
+        description: T5X image from ghcr.io/nvidia/pax
```
nouiz, line 8: t5x -> pax

yhtang: Good catch 😄 fixed

PR #65, opened 2023-05-16
Allows rosetta to be built with pax+praxis

terrykong:
* Allows rosetta to be built with pax+praxis as a base
* CI will build rosetta with pax and t5x
* Allows local directories to be provided as build-context to let the user specify custom source


terrykong:
@ashors1 for viz

.github/workflows/_build_rosetta.yaml
```diff
@@ -1,13 +1,38 @@
...
 
 on:
+  workflow_dispatch:
+    inputs:
+      BASE_IMAGE:
+        type: string
+        description: 'Base docker image that provides the foundation library:'
+        required: false
+        default: ghcr.io/nvidia/t5x:latest
+      BASE_LIBRARY:
```
yhtang, line 11: It seems that you are trying to accommodate the scenario where a user wants to build on top of an alternative image other than `ghcr.io/nvidia/${BASE_LIBRARY}:latest`. Meanwhile, we'd better perform some input validation so a user gets a meaningful error message when, say, trying to build a pax rosetta but forget to change the default image from t5x to pax.

How about we make `BASE_LIBRARY` as the first input and make `BASE_IMAGE` optional with an empty default? We can then add a job step to dynamically pick the default image, or check for the existence of t5x/pax in the user-specified container before starting the actual build.

.github/workflows/_build_rosetta.yaml
```diff
@@ -1,13 +1,38 @@
 name: ~build Rosetta container
 
 on:
+  workflow_dispatch:
```
yhtang, line 4: Just a reminder for myself to ensure workflow_dispatch is removed before merging.

.github/workflows/ci.yaml
```diff
@@ -147,16 +147,26 @@ jobs:
...
+      BASE_LIBRARY: t5x
+    secrets: inherit
+
+  build-rosetta-pax:
+    uses: ./.github/workflows/_build_rosetta.yaml
+    needs: [metadata, build-pax]
+    with:
+      BUILD_DATE: ${{ needs.metadata.outputs.BUILD_DATE }}
+      BASE_IMAGE: ${{ needs.build-pax.outputs.DOCKER_TAGS }}
+      BASE_LIBRARY: pax
```
yhtang, line 165: It looks like this will trigger the exact error that I hypothesized above :grin:

.github/workflows/ci.yaml
```diff
@@ -147,16 +147,26 @@ jobs:
...
+  build-rosetta-t5x:
     uses: ./.github/workflows/_build_rosetta.yaml
     needs: [metadata, build-t5x]
     with:
       BUILD_DATE: ${{ needs.metadata.outputs.BUILD_DATE }}
       BASE_IMAGE: ${{ needs.build-t5x.outputs.DOCKER_TAGS }}
+      BASE_LIBRARY: t5x
+    secrets: inherit
+
+  build-rosetta-pax:
```
yhtang, line 159: We can use [matrix build](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) here to build Rosetta against t5x/pax.

rosetta/Dockerfile.pax
```diff
@@ -0,0 +1,56 @@
+# syntax=docker/dockerfile:1-labs
+ARG BASE_IMAGE=ghcr.io/nvidia/paxml:latest
+ARG GIT_USER_EMAIL=jax@nvidia.com
+ARG GIT_USER_NAME=NVIDIA
+
+FROM scratch as rosetta-source
+COPY . /
+
+FROM scratch as pax-mirror-source
+ADD --keep-git-dir=true https://github.com/google/paxml.git#main /
```
yhtang, line 10: There is an `--keep-git-dir` option now? That's amazing 😄

yhtang (changes requested):
Big change set 😁

yhtang (changes requested):
Big change set 😁

yhtang (changes requested):
Big change set 😁

rosetta/Dockerfile.pax
```diff
@@ -0,0 +1,56 @@
...
+FROM ${BASE_IMAGE} AS distribution-builder
+
+ARG GIT_USER_EMAIL
+ARG GIT_USER_NAME
+RUN git config --global user.email "${GIT_USER_EMAIL}" && \
+    git config --global user.name "${GIT_USER_NAME}"
+
+COPY --from=rosetta-source / /opt/rosetta
+WORKDIR /opt/rosetta
+RUN --mount=target=/opt/pax-mirror,from=pax-mirror-source,readwrite --mount=target=/opt/praxis-mirror,from=praxis-mirror-source,readwrite <<EOF
```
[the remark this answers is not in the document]
terrykong: Yea, that's basically it. There's a good blog post about them [from docker](https://www.docker.com/blog/introduction-to-heredocs-in-dockerfiles/)

For `RUN` commands, it'll make our installation look more like a bash script so we don't need to have `&& \` on every line

The neat thing that's now possible is using heredocs for `COPY` commands. You can use a heredoc to embed something like a tiny config into a dockerfile without having to have a separate file somewhere else; especially if it's not useful outside of the context of the dockerfile

.github/workflows/_build_rosetta.yaml
terrykong: Good point. I can make that change

.github/workflows/_build_rosetta.yaml
terrykong: OOC, is there any reason not to allow workflow_dispatch on a reusable workflow? Just curious since if we "matrixify" the rosetta build/test for pax/t5x, someone could test the individual build/test workflows by dispatching the reusable _build_rosetta.yaml or _test_rosetta.yaml. But if someone needs to run the "matrixified" nightly-rosetta-build.yaml, we'd have to add logic to test one or the other OR make a concrete nightly_rosetta_t5x_build.yaml in addition to the "matrixified" nightly_rosetta_build.yaml, which would repeat a lot of the scaffolding

.github/workflows/nightly-rosetta-pax-build.yaml
```diff
@@ -0,0 +1,74 @@
...
+    if: (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success') || github.event_name == 'workflow_dispatch'
+    needs: metadata
+    uses: ./.github/workflows/_build_rosetta.yaml
+    with:
+      BUILD_DATE: ${{ needs.metadata.outputs.BUILD_DATE }}
+      BASE_LIBRARY: pax
+      BASE_IMAGE: ${{ env.DOCKER_REGISTRY }}/pax:latest
+    secrets: inherit
+
+  test:
```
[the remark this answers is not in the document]
terrykong: So this suggestion and https://github.com/NVIDIA/JAX-Toolbox/pull/65#discussion_r1196798672 are related:

The issue I ran into with matrix builds here is that I want a "matrix build" to next trigger a "matrix test". It seemed like there wasn't great GH actions support for this. I've seen a workaround where someone passes a json artifact as the output of a "matrix build" and the "matrix test" consumes it with the grammar `strategy.matrix: ${{ fromJSON(needs.build.outputs.matrix) }}`. But IIUC, the test jobs need to wait until ALL of the matrix jobs finish to start testing.

So the workaround was just to combine build and test to simplify things, since the test can start right after the build, and I wouldn't have to rewrite a lot of the same yaml file to have a concrete `nightly_rosetta_${BASE_LIBRARY}_test.yaml`. The other benefit is the one you mentioned where build/test/publish are all in the same workflow; so (in the future) we can publish the images and badges contingent on the tests succeeding. Also, I don't think this is a concern now, but it seems like this will postpone us needing to address `workflow_run`'s [3 workflow limit](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run:~:text=You%20can%27t%20use%20workflow_run%20to%20chain%20together%20more%20than%20three%20levels%20of%20workflows.)

I can re-separate build and test, but having dependent matrix builds seems like more work than they're worth. I'm curious if you know of a better solution here.

terrykong:
+ @nluehr for viz

.github/workflows/_build_rosetta.yaml
yhtang: This is meant to avoid duplication of the `inputs` section and to simplify input processing logic in the reusable workflow. I wish to keep the messy part in the 'frontpage' workflows while extracting the clean and simple common cores into reusable workflows. The mental mode is similar to a frontend/backend separation.

For your particular use case, we could run the matricized workflow using a bunch of single-element inputs, essentially launching a 1x1x1x...x1 matrix :grin:

.github/workflows/nightly-rosetta-pax-build.yaml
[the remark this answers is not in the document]
yhtang: That's a really good point. Let's keep building and testing together here, while I prioritize the badge creation process?

rosetta/Dockerfile.pax
yhtang: I've seen this `bake` command, but to me, this looks like a simplified version of `build` that is meant for ease of use. Is there a particular feature that will help us in our workflow?

rosetta/Dockerfile.pax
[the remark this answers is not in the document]
yhtang: That's really cool!

.github/workflows/ci.yaml
yhtang: Postponed due to [limitation in GitHub Actions workflow dependency](https://github.com/orgs/community/discussions/16059).

yhtang (approved):

PR #71, opened 2023-05-19
Fix te-badge to fail if no tests found; and count only calls in pytest

terrykong:


terrykong:
Closing since it's addressed in #59

PR #73, opened 2023-05-19
Fix Multiprocessing in Pax Tests

ashors1:
Fixes the `TOTAL_TASKS` variable substitution to allow Pax tests to run with multiprocessing.


yhtang (approved):
LGTM

PR #75, opened 2023-05-21
Fix rosetta build

terrykong:
Fixes issues not caught in first #65 :

* heredocs in Dockerfile need `bash -e` to exit if failure
* disable library validation b/c nvidia-container-toolkit not available in builder runner
* refactored env context usage since it is not available in under with context
* updated badges in frontpage readme  for rosetta t5x/pax

### DIFF
--- a/.github/workflows/_build_rosetta.yaml
+++ b/.github/workflows/_build_rosetta.yaml
@@ -81,7 +81,7 @@ jobs:
           if [[ $PYTHON_PACKAGE == "pax" ]]; then
             PYTHON_PACKAGE=paxml
           fi
-          docker run --rm ${{ env.BASE_IMAGE }} python -c "import $PYTHON_PACKAGE"
+          docker run --rm ${{ env.BASE_IMAGE }} python -c "import importlib,sys; sys.exit(importlib.util.find_spec('$PYTHON_PACKAGE') is None)"
 
       - name: Build docker images
         uses: docker/build-push-action@v4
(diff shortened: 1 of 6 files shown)


.github/workflows/_build_rosetta.yaml
```diff
@@ -74,15 +74,15 @@ jobs:
...
-      - name: Pull and validate BASE_IMAGE=${{ env.BASE_IMAGE }} contains BASE_LIBRARY=${{ inputs.BASE_LIBRARY }}
-        shell: bash -x -e {0}
-        run: |
-          PYTHON_PACKAGE=${{ inputs.BASE_LIBRARY }}
-          if [[ $PYTHON_PACKAGE == "pax" ]]; then
-            PYTHON_PACKAGE=paxml
-          fi
-          docker run --rm ${{ env.BASE_IMAGE }} python -c "import $PYTHON_PACKAGE"
-
+      # TODO: Can't run a cuda image yet
```
yhtang, line 77: Use `runs-on: [self-hosted, small-builder]` and it will work. I added the CUDA container runtime to our self-hosted runners 😁

yhtang (changes requested):
Just one comment regarding the docker run test 😄

.github/workflows/_build_rosetta.yaml
terrykong: Awesome! I can uncomment. Thanks for the upgrade~

terrykong: I tried, but it didn't work. Am I missing something?

https://github.com/NVIDIA/JAX-Toolbox/actions/runs/5082514197/jobs/9132336819

yhtang (approved):

PR #76, opened 2023-05-22
Use nvcr.io private registry to stage CI internal containers

yhtang:
Pros:
- No longer need to pay for private container storage on ghcr.io
- Potentially better confidentiality if builds contain NVIDIA IP not authorized for release yet.
- Potentially more reliable pulling to/from dlcluster/selene
Cons:
- (ironically) more difficult to grant access, users need to apply to internal portal first.
- docker push/pull slightly slower than ghcr.io for CI jobs. Assumed due to better colocation of ghcr.io and Actions servers.

This PR is for preview purposes and may not represent the final work to be merged.

A good compromise may be to use ghcr.io for instantaneous access in jobs, but use a periodically scheduled Actions workflow to archive containers older than a certain time to nvcr.io.

### DIFF
--- a/.github/workflows/_build_base.yaml
+++ b/.github/workflows/_build_base.yaml
@@ -19,7 +19,7 @@ on:
         value: ${{ jobs.build.outputs.DOCKER_TAGS }}
 
 env:
-  UPLD_IMAGE: ghcr.io/nvidia/jax-toolbox-internal
+  UPLD_IMAGE: nvcr.io/nvidian/jax-toolbox-internal
 
 permissions:
   contents: read  # to fetch code
@@ -42,9 +42,9 @@ jobs:
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v2
         with:
-          registry: ghcr.io
-          username: ${{ github.repository_owner }}
-          password: ${{ secrets.GITHUB_TOKEN }}
+          registry: nvcr.io
+          username: '$oauthtoken'
+          password: ${{ secrets.NVCR_TOKEN }}
 
       - name: Set docker metadata
         id: meta
--- a/.github/workflows/_build_base.yaml
+++ b/.github/workflows/_build_base.yaml
@@ -39,7 +39,7 @@ jobs:
       - name: Check out the repository under ${GITHUB_WORKSPACE}
         uses: actions/checkout@v3
 
-      - name: Login to GitHub Container Registry
+      - name: Login to NVIDIA Container Registry
         uses: docker/login-action@v2
         with:
           registry: nvcr.io
--- a/.github/workflows/_build_jax.yaml
+++ b/.github/workflows/_build_jax.yaml
@@ -39,7 +39,7 @@ on:
         value: ${{ jobs.build.outputs.DOCKER_TAGS }}
 
 env:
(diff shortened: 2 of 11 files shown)


terrykong:
Shortening the keep window to a month seems really short; especially from the perspective of someone adding a new model. They may start developing their model but then after 1.5 months, they won't be able to reproduce their numbers because the base container is gone.

If we have to make the window 1 month for ghcr, my vote would be to mirror the registry on nvcr.io and have a longer keep window (maybe 6 months).

yhtang:
To clarify, I am only proposing moving containers in the _private_ `ghcr.io/nvidia/jax-toolbox-internal` repo to `nvcr.io/nvidian`. This repo is a staging area for containers produced in our CI workflows. Separate jobs exist to copy select containers from this internal repo to the public ones for outside users to check out. These public repos, e,g. `ghcr.io/nvidia/jax` will always remain on GitHub.

Here is an example of the current container storage solution:

1. Job build-jax" creates `ghcr.io/nvidia/jax-toolbox-internal:5061628209-jax` (user needs to be member of the NVIDIA org on GitHub to access)
2. Optional: job "publish-jax" copies `ghcr.io/nvidia/jax-toolbox-internal:5061628209-jax` to `ghcr.io/nvidia/jax:nightly-2023-05-23` (anyone can access)

Two candidate solutions to use `nvcr.io` for container storage as discussed above:

1. Use `nvcr.io` for private image repo, and `ghcr.io` for all public image repos. For example:
    - Job "build-jax" creates `nvcr.io/nvidian/jax-toolbox-internal:5061628209-jax` (needs NGC registration + nvidia*n* membership to access)
    - Optional job "publish-jax" copies `nvcr.io/nvidia/jax-toolbox-internal:5061628209-jax` to `ghcr.io/nvidia/jax:nightly-2023-05-23` (anyone can access)
2. Use `ghcr.io` for all repos, but only retains private images for 30 days before transferring them to `nvcr.io` for archive.
    - Job "build-jax" creates `ghcr.io/nvidian/jax-toolbox-internal:5061628209-jax`
    - Optional job "publish-jax" copies `ghcr.io/nvidia/jax-toolbox-internal:5061628209-jax` to `ghcr.io/nvidia/jax:nightly-2023-05-23`
    - Cron job "archive" moves `ghcr.io/nvidia/jax-toolbox-internal:*` tags older than 30 days to `nvcr.io/nvidian/jax-toolbox-internal:*`

yhtang:
and I agree with @nouiz and @terrykong that solution 2 seems a better choice.

yhtang:
This is the [cost of private repo storage](https://docs.github.com/en/billing/managing-billing-for-github-packages/about-billing-for-github-packages#about-billing-for-github-packages) on `ghcr.io`:
> GitHub charges usage to the account that owns the repository where the package is published. If your account's usage surpasses these limits and you have set a spending limit above $0 USD, you will pay $0.008 USD per GB of storage per day and $0.50 USD per GB of data transfer.

So if we accumulate 10 TB of containers after 6 months, the monthly bill will be 10000 GB x $0.008/GB/day x 31days/month = $2480/month.

nouiz:
I'm good with both of those options. 1 seem simpler. 2 a little better in that it will speed up a little bit the jobs. But I'm not sure it will make an impact in the end.

PR #79, opened 2023-05-24
Make t5x/pax and all downstream builds depend on jax-te instead of jax

terrykong:
This change switches the default build of t5x, pax, and downstream rosetta builds to depend on jax-te instead of jax as a base build.

This will simplify the number of images we need to maintain since we won't have to maintain a t5x and t5x+te; or rosetta-t5x and rosetta-t5x+te.

@nouiz and @ashors1


nouiz:
LGTM.

yhtang (approved):
LGTM

PR #82, opened 2023-05-30
Fixes badge errors for te and jax

terrykong:
Few things this PR fixes:
- the `inputs` context isn't allowed in the `with` context for reusable workflows
- the PUBLISH input was boolean, but b/c it was passed thru the concrete workflow, it was converted to a string and caused errors when calling the reusable workflow
- `_publish_badge.yml` inserted step outputs inline which added an extra single quote and messed with the badge color parsing. Using shell variables works around the insertion of the extra single quotes


yhtang (approved):
Cool! Thanks!

PR #85, opened 2023-06-02
Flax distribution

terrykong:
In preparation to support some jax models in rosetta, we also need flax to be a distribution as well. This PR adds it to both the t5x and pax builds so that rosetta can build upon the install.

FWIW, pax builds can install flax in any order b/c it doesn't install flax from source, but t5x installs from source so the flax install has to happen after t5x.


yhtang (approved):
LGTM :)

PR #90, opened 2023-06-13
Propagate script failure

maanug-nv:
`test-pax.sh` should propagate failure of tests to the return value of the script, this can be done with a bash option '-e'.


yhtang (approved):
LGTM

PR #91, opened 2023-06-13
PAX Get Slurm Job status

maanug-nv:
- Get Slurm job status with `sacct`
- Prevent workflow failure if checkpoint and artifacts are not found

### DIFF
--- a/.github/workflows/_test_pax.yaml
+++ b/.github/workflows/_test_pax.yaml
@@ -137,6 +137,12 @@ jobs:
           name: ${{ steps.meta.outputs.JOB_NAME }}
           path: output/*
 
+      - name: Check failure conditions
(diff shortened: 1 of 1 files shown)


maanug-nv:
I plan to add a badge with a count of successful tests similar to TE in another PR. I can add an additional step at the end of `_test_pax.yaml` to exit with the `SLURM_EXITCODE` so that the job fails if the slurm job failed, if that is helpful but I didn't know if we want both.

terrykong:
Other than that small nit, LGTM!

ashors1 (approved):
LGTM. Thanks! If we're already reporting that the slurm job fails, I don't think it's necessary to have the full job fail as well.

PR #94, opened 2023-06-20
Pax new completion badge

maanug-nv:
- write job status to a json file
- run publish badge workflow using that status


maanug-nv:
I've replaced the badge url for pax. This badge is just tracking if the jobs completed, next steps will be to check the results of passing tests against a loss and iter time baseline.

ashors1 (dismissed):

terrykong:
Thanks for fixing this! Now we'll have a better idea what's failing

terrykong (approved):

Issue #98, opened 2023-06-22
Need rosetta build badge separated from workflow status now that rosetta's workflow file contains build+test


terrykong:
Because of limitations in GH Action workflows that make chaining matrix stages difficult, rosetta builds were consolidated so that the upstream pax build triggers the downstream rosetta-pax build+test workflow. This means that if the testing stage fails (which it currently does), the rosetta-pax badge shows fail because it is based on the worfklow status as opposed to a img shield badge.

We should create a custom badge for the build and use that instead of relying on the GH workflow status

PR #99, opened 2023-06-27
Adds proper build/test badges for rosetta && fixes gist upload for new files

terrykong:
Previously rosetta used the workflow status for the build status, but now that rosetta has build and test in one workflow, it sends the wrong signal if the test fails but build succeeds.

This change makes it so that the rosetta build and test both have proper badges regardless of if one fails.

Notably, if rosetta build fails, the test badge is still created, but will just show "n/a".

**Note**: This commit also has a fix for gist uploading since I was encountering it while adding this change. I'm actually not sure how gist uploading worked previously, but the change filters out keys that aren't needed since only the "content" key is needed ([api ref](https://docs.github.com/en/rest/gists/gists?apiVersion=2022-11-28#update-a-gist))

Addresses #98

### DIFF
--- a/.github/workflows/_publish_badge.yaml
+++ b/.github/workflows/_publish_badge.yaml
@@ -79,6 +79,15 @@ jobs:
               gist_id: gistId,
               files: {
                 [filename]: { content },
-                ...Object.fromEntries(Object.entries(gist.files).filter(([name]) => name !== filename))
-              }
+                ...Object.fromEntries(
+                  Object.entries(gist.files)
+                    .filter(([name]) => name !== filename)
+                    .map(([name, value]) => [
+                      name,
+                      Object.fromEntries(
+                        Object.entries(value).filter(([nestedKey]) => nestedKey === "content")
+                      ),
+                    ])
+                ),
+              },
             });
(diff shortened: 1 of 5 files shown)


.github/workflows/_publish_badge.yaml
```diff
@@ -79,6 +79,15 @@ jobs:
...
-                ...Object.fromEntries(Object.entries(gist.files).filter(([name]) => name !== filename))
-              }
-            });
+                ...Object.fromEntries(
+                  Object.entries(gist.files)
+                    .filter(([name]) => name !== filename)
+                    .map(([name, value]) => [
+                      name,
+                      Object.fromEntries(
+                        Object.entries(value).filter(([nestedKey]) => nestedKey === "content")
```
yhtang, line 88: Why do we only keep `content`? What are the attributes that we are trying to get rid of?

terrykong: I stumbled upon this b/c my job was failing. Here's an example: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/5383094321

Then I looked up the gist api for updating a gist and apparently it says it only supports "content" and an optional "filename". So I just filtered out content, since the filename I assumed was inferred from the key.

I'm actually not sure how this worked in the first place, maybe the API was more permissive when the files within the gist were already there, but I ran into it b/c I was creating new files?

yhtang (approved):

PR #110, opened 2023-07-07
Fixes issue in rosetta dockerfiles where distributions were copied, but deletions weren't being respected. Switched away from multi-stage builder pattern to fix

terrykong:
@ashors1 reported this issue where distributions had files that were supposed to be deleted in commits in the patchlist. The issue was that `COPY` was used to copy from the builder to the `rosetta` stage, which would be fine for everything except deletions or for files that were renamed. The simple fix is to remove the distribution-builder for now. The image size of building the distribution is still very small (<2MB), so we're not losing much

### DIFF
--- a/rosetta/Dockerfile.pax
+++ b/rosetta/Dockerfile.pax
@@ -15,7 +15,7 @@ ADD --keep-git-dir=true https://github.com/google/praxis.git#main /
 FROM scratch as flax-mirror-source
 ADD --keep-git-dir=true https://github.com/google/flax.git#main /
 
-FROM ${BASE_IMAGE} AS distribution-builder
+FROM ${BASE_IMAGE} AS rosetta
 
 ARG GIT_USER_EMAIL
 ARG GIT_USER_NAME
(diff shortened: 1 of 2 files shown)


terrykong:
Moving to draft. Addressing issue with straggling mirror dirs

yhtang:
`COPY` should honor files that are deleted. For example, using this Dockerfile
```
FROM ubuntu:22.04 as stage1
RUN mkdir /test && touch /test/file1 && touch /test/file2
RUN rm /test/file1

FROM ubuntu:22.04 as stage2
COPY --from=stage1 /test /test
```
We will find in the `/test` folder of the resulting image that:
```
root@ad72916bf497:/# ls /test/
file2
```
So only `file2` is left.

yhtang:
@terrykong Could it be another reason that was causing the files to remain? For example, the wrong path is passed to `rm`?

terrykong:
I think it's more like:

```
FROM ubuntu:22.04 as stage1
# Assume there's a files already in ubuntu:22.04
RUN bash -c "[[ -e /etc/os-release ]]"
RUN rm -f /etc/os-release

FROM ubuntu:22.04 as stage2
COPY --from=stage1 /etc /etc
# And after this the /etc/os-release is still there even though I wanted it deleted.
```

Where `/etc` is a stand-in for the distribution directory. Basically the file was deleted in the builder, but the secondary "rosetta" stage had the same directory, so when the directory was copied over, the file that was deleted in the builder did not overwrite the one in the "rosetta" stage since the file existed there already. I looked it up but unfortunately there's not a flag like `--delete` like there is for `rsync`

yhtang (approved):

PR #111, opened 2023-07-07
PAX Check loss and perf

maanug-nv:
- Adds baselines for 4 of the PAX mgmn tests
- Adds pytest files to compare tensorboard file result of mgmn tests against baselines
- Adds job to run above pytest in workflow


terrykong:
It would be neat to surface the results of this either in a GITHUB_STEP_SUMMARY or a badge somewhere. Any thoughts about adding that?

maanug-nv:
Agreed, that's my next step 😄.  Just wanted to split into two PRs.

.github/workflows/baselines/pytest/test_pax_mgmn_metrics.py
```diff
@@ -0,0 +1,49 @@
...
+def test_step_time(baseline_filename):
+    baseline_filepath = os.path.join(baselines_dir, baseline_filename)
+    test_config = baseline_filename.split(".")[0]
+    event_file = os.path.join(results_dir, test_config, "summaries/train/events*")
+    event_file = glob.glob(event_file)[0]
+    with open(baseline_filepath, "r") as baseline_file:
+        step_time_avg_expected = json.load(baseline_file)["step_time_avg"]
+        step_time_values = test_utils.read_tb_tag(event_file, step_time_summary_name).values()
+        step_time_avg_actual = mean(step_time_values)
+        assert step_time_avg_expected + STEP_TIME_DELTA > step_time_avg_actual
```
terrykong, line 38: Is there some advantage of an additive threshold vs. a multiplicative one?

yhtang:
How are the baseline numbers in the JSON files derived? We need to document the source of the data so that when a significant change happens, we know where to look for the diffs.

If they are from a particular Workflow run, maybe we can simply reference it and checkout the associated artifacts? This way, the source and methodology of the data is self-explanatory.

terrykong:
@yhtang Raises a good point about baseline provenance.

What about in addition to the python file that create the baseline jsons from artifacts, you add a script that takes jobids and pulls the artifacts and calls this python script? That way from the jobid, we can encode the URL of the workflow run and the date it ran in the JSON for inspection and repeatability?

yhtang:
> What about in addition to the python file that create the baseline jsons from artifacts, you add a script that takes jobids and pulls the artifacts and calls this python script? That way from the jobid, we can encode the URL of the workflow run and the date it ran in the JSON for inspection and repeatability?

I like this idea. Essentially, we can wrap this up as a reusable workflow, say `_compare_metrics.yaml`, which we can mentally regard as a function. This workflow/function will accept 2 workflow/job IDs, download the associated artifacts, and determine if the differences between certain metrics (can be specified via an input) is large.

This way, we can reuse this workflow for comparing results for PAX, T5X, JAX, TE, etc...

terrykong (approved):

Issue #109, opened 2023-07-06
XLA_FLAGS to enable LHS should be exported in JAX container


nluehr:
Similar to the environment variables in issue https://github.com/NVIDIA/JAX-Toolbox/issues/105, the following XLA flags should be enabled by default to provide out-of-the-box performance at least for paxml models (and probably for all JAX-based containers).

```
XLA_FLAGS=--xla_gpu_enable_latency_hiding_scheduler=true --xla_gpu_enable_async_all_gather=true --xla_gpu_enable_async_reduce_scatter=true --xla_gpu_enable_triton_gemm=false
```

terrykong:
@nluehr Given that these are going to be tested for paxml models, should we set these for the pax containers first, and move them to jax after we test them for other models/frameworks?

> (and probably for all JAX-based containers)

Do you expect the perf to be the same or better? Or is it possible that by enabling these, other models can become less performant?

nouiz:
Maybe we should check with Google why they aren't enabled by default?

Issue #105, opened 2023-06-29
Adding new environment variables to speed up nccl collectives in Jax.


abhinavgoel95:
Could we add these environment variables in the JAX containers? They help with some issues we are facing with unexpectedly long nccl kernels.

```
CUDA_DEVICE_MAX_CONNECTIONS=1
NCCL_AVOID_RECORD_STREAMS=1
NCCL_IB_SL=1
NCCL_PROTO=LL128
```
```[tasklist]
### Tasks
```

terrykong:
Two questions:
1. Are these variables model specific, or are they helpful across all models? 
2. Is this pax or t5x specific, or helpful for all?
This will help us narrow down where these should live

abhinavgoel95:
1. Helpful across all models.
2. Should help all frameworks/libraries.

yhtang:
Just out of curiosity, if the variables are helpful across all frameworks, why aren't these set as the default behavior of NCCL?

nluehr:
For the CUDA_DEVICE_MAX_CONNECTIONS=1, it will generally be helpful in JAX/XLA cases because the kernel launches have already been ordered during compilation/scheduling. In the general case, setting MAX_CONNECTIONS=1 can result a ready-to-run kernels getting stuck behind other kernels waiting on execution dependencies.

PR #114, opened 2023-07-11
Adding default XLA/GPU env vars for all JAX-based containers

terrykong:
Addresses #105 and #109 

*  `--xla_gpu_enable_latency_hiding_scheduler=true`: Allows XLA:GPU to move communication collectives to increase overlap with compute kernels.
* `--xla_gpu_enable_async_all_gather=true`: Allows XLA:GPU to run All Gather NCCL kernels on a separate CUDA stream to allow overlap with compute kernels.
* `--xla_gpu_enable_async_reduce_scatter=true`: Allows XLA:GPU to run Reduce Scatter NCCL kernels on a separate CUDA stream to allow overlap with compute kernels.
* `--xla_gpu_enable_triton_gemm=false`: Disallows Trition GeMM kernels; uses CUBLAS GeMM kernels instead. CUBLAS kernels are currently better tuned for GPUs and thus provide better performance.
* `CUDA_DEVICE_MAX_CONNECTIONS=1`: Use a single queue for GPU work, lowers latency of each stream operation. OK since XLA already orders launches.
* `NCCL_IB_SL=1`: defines the InfiniBand Service Level ([1](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html#nccl-ib-sl))

@nluehr @ashors1


.github/container/Dockerfile.jax
```diff
@@ -47,6 +47,11 @@ ARG SRC_PATH_JAX
 ARG SRC_PATH_XLA
 ARG BUILD_DATE
 ENV BUILD_DATE=${BUILD_DATE}
+ENV XLA_FLAGS="--xla_gpu_enable_latency_hiding_scheduler=true --xla_gpu_enable_async_all_gather=true --xla_gpu_enable_async_reduce_scatter=true --xla_gpu_enable_triton_gemm=false"
```
nluehr, line 50: @abhinavgoel95 should we also include `xla_gpu_enable_pipelined_all_reduce` and `xla_gpu_enable_pipelined_all_gather` to this list?

abhinavgoel95: No, we see perf issues with those flags as of today. Hoping they get fixed by early next week.

yhtang:
Could someone please edit the PR description to document the rationale behind each flag?

terrykong:
> Could someone please edit the PR description to document the rationale behind each flag?

@nluehr @abhinavgoel95 ?

yhtang:
- `NCCL_PROTO=LL128`:
  [doc](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html#nccl-proto) says:
  > The default is LL,LL128,Simple on platforms which support LL128, LL,Simple otherwise.
  
  Related issue: https://github.com/NVIDIA/nccl/issues/281

- NCCL_AVOID_RECORD_STREAMS=1: this is undocumented in https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html.

- NCCL_IB_SL=1
  [doc](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html#nccl-ib-sl) says:
  > Defines the InfiniBand Service Level.
  > For more information, see the InfiniBand specification Volume 1 (https://www.infinibandta.org/ibta-specifications-download) or vendor documentation.
  
  and the [link](https://www.infinibandta.org/ibta-specifications-download) in the quoted text above is **dead**.

nluehr:
@abhinavgoel95 How much speedup do we see from setting `NCCL_PROTO=LL128` in practice?
The combiner threshold may come into play here as for larger collectives, I think LL128 could reduce effective bandwidth by ~6%.

nluehr:
From my reading of the NCCL code, the only safe way to enable LL128 is to not specify NCCL_PROTO at all.
LL128 is valid only on Volta, Ampere, and Hopper with NVLink. PCIE can cause packet re-ordering that results in silent corruption.
Since we expect users will run our models on a variety of clusters, I think we shouldn't define NCCL_PROTO either in the JAX dockerfile nor in the paxml run scripts. If removing this results in worse performance, we should open a bug with the NCCL team.

ashors1:
We're also sometimes interested in setting `XLA_PYTHON_CLIENT_MEM_FRACTION=0.9` to increase the amount of memory allocated to XLA. Is this something we'd want to consider adding to the base container as well?

nluehr:
Whether we want to set XLA_PYTHON_CLIENT_MEM_FRACTION depends on the workload (particularly whether any other GPU libraries are being used outside of XLA).

My opinion is that this would be OK to set in the paxml container, but for the JAX container in general we shouldn't set it because there we expect users to extend it in varied workflows.

.github/container/Dockerfile.jax
```diff
@@ -47,6 +47,11 @@ ARG SRC_PATH_JAX
...
+ENV CUDA_DEVICE_MAX_CONNECTIONS=1
+ENV NCCL_AVOID_RECORD_STREAMS=1
+ENV NCCL_IB_SL=1
+ENV NCCL_PROTO=LL128
```
nluehr, line 54: Lets drop the NCCL_PROTO option.

terrykong: Done

nluehr:
As best I can tell, NCCL_AVOID_RECORD_STREAMS is a feature of pytorch rather than NCCL. (It seems it's now been renamed to TORCH_NCCL_AVOID_RECORD_STREAMS [here](https://github.com/pytorch/pytorch/commit/0781188e64c7065dc7c3a9607ebb4cf0ee0e3d16)).
So I think it's safe to drop it entirely.

abhinavgoel95 (approved):

yhtang (changes requested):
Did we figure out a source for the undocumented `NCCL_AVOID_RECORD_STREAMS` variable?

terrykong:
@abhinavgoel95

abhinavgoel95:
@yhtang @terrykong it is safe to drop NCCL_AVOID_RECORD_STREAMS. We do not need to upstream this. It is a PyTorch specific feature.

terrykong:
I removed mention of that env var. Can I re-request your review @yhtang ?

yhtang (approved):

PR #116, opened 2023-07-12
Allow PR refs for jax or xla

terrykong:
We already allow standard git-refs to be configurable in the build for xla and jax.

This change makes it possible to specify a PR git-ref so we can test an experimental change.


yhtang:
Could you please provide an example of how PRs are specified for a build?

terrykong:
Sure. So for this PR: https://github.com/openxla/xla/pull/4044/files

You'd specify it in the `CI` workflow as `https://github.com/openxla/xla.git#pull/4044/head`

yhtang:
We can make it more intuitive. `pull/<ID>/heads` feels a bit magical here, and users would have to guess the correct format from the git refs.

Given that a PR is referenced in GitHub Flavored Markdown by `user/repo#<PR_ID>`, which coincides with that `ci.yaml` currently accepts, how about we determine if the content after `#` is a PR if it is:
- purely numeric
- decimal

There is a small chance of conflict if a branch is named by a decimal integer, which is extremely rare in practice. The chance that it collides with a commit hash is also negligible.

Moreover, [this job](https://github.com/NVIDIA/JAX-Toolbox/blob/c9caba27fd70f2e9671c153f103e9e9623769f3c/.github/workflows/ci.yaml#L80) might be a better place to host such logic. The idea is to keep the Dockerfile clean and independent of the CI workflow.

terrykong:
> GitHub Flavored Markdown by user/repo#<PR_ID>

That's cool! I didn't realize that was that syntax in GH markdown. 

> Moreover, [this job](https://github.com/NVIDIA/JAX-Toolbox/blob/c9caba27fd70f2e9671c153f103e9e9623769f3c/.github/workflows/ci.yaml#L80) might be a better place to host such logic. The idea is to keep the Dockerfile clean and independent of the CI workflow. We can detect if the content after # is a PR, then convert it to the corresponding git ref and pass it into subsequent reusable workflows/Dockerfiles.

We can definitely move the logic that converts integers to PR git-refs `1320` -> `pull/1320/head` in that job. But, I'm not sure of a better way to reference the PR other than what's already in the Dockerfile. It is a bit magical that that's the git-ref, but it is what is documented in the [GH's docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally#modifying-an-inactive-pull-request-locally)

I can make the change to accept a decimal instead.

terrykong:
Closing this as it complicates the dockerfile. Better ways of handling this were discussed offline like potentially moving the logic of fetching github git refs outside of the dockerfile

PR #117, opened 2023-07-13
pax and t5x builds now depend on jax instead of te and install te in their own build step

terrykong:
This is to address a dependency issue that just recently cropped up:

- T5x required seqio required TF-text required TF required typing_extensions
- TE required pydantic required typing_extensions

When installed sequentially, typing_extensions is downgraded by t5x and is in conflict with the one in TE. There exists a working combination of versions, but it is only achievable if pip installed altogether. This change allows deferring installation of a library so that it can be collectively installed later.


yhtang:
Shall we deprecate the `ghcr.io/nvidia/te` containers if we adopt this PR?

terrykong:
Yea, that makes sense, After all, if someone wanted to start from the TE image and try to install t5x, they'd run into the same issue as us.

yhtang:
I see. Meanwhile, maybe we should really ship TE even in the `jax` container.

terrykong:
I don't see any issues with that.

I think even if you do move it into the jax container, the pax/t5x builds will still need to add TE as a parallel dependency (`pip install paxml transformer_engine` vs transitive `pip install paxml` and paxml depends on TE) since we want pip to search for versions that satisfy both libraries. 

It is better to have it as a transitive dependency, but that assumes our TE code changes are accepted as well.

yhtang:
I agree with your points. My thoughts are more from a product perspective; the concern is balancing the integration of NVIDIA software versus bloating the image.

yhtang (approved):

PR #118, opened 2023-07-13
Adds slurm scripts for t5x and updates t5x instructions to use latest nightlies

terrykong:
SLURM scripts were moved to rosetta outside of t5x since it is inconvenient for an end-user to pull an image just to get sbatch submit scripts.


yhtang:
I would sweat if I were asked to run these SLURM scripts for the first time 😂 
Is there any way that we can make them more readable and less lengthy?

sharathts:
@yhtang Unfortunately not. These are our minimal and standard slurm scripts used throughout. We use the same ones for all DLE release as well as for MLPerf.

sharathts (dismissed):

yhtang:
At least we can add some comments to document the use/source/rationale of the variables and setup? This would ensure a new user can learn the thinking process behind the script so that they can adapt it rationally.

terrykong:
I plan on improving docs in subsequent PRs.

Once #117 is in, this can be merged

sharathts (approved):

PR #125, opened 2023-07-17
Reorganizing badges on front page readme

terrykong:
This is a change to the organization in the front page readme.

The readme will go from 
image
 to 
image

This change does not add functionality, but merely restructures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,105 @@
 # JAX Toolbox
 
-| Image                                                | Build                                      | Test                                   |
-| ---------------------------------------------------- | ------------------------------------------ | -------------------------------------- |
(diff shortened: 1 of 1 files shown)


terrykong:
Will wait for https://github.com/NVIDIA/JAX-Toolbox/pull/132 to be merged since that PR will remove the TE badges in the front page readme. Will rebase this change after that is in

terrykong:
After offline discussions with @yhtang, will keep the readme change, but revert the workflow changes since it prolongs build/test times since all builds are run collectively and we may overwrite a previous build. Will revert workflow changes shortly

yhtang (approved):

PR #132, opened 2023-07-19
Ship Transformer Engine in the JAX container

yhtang:
After merging this, we will no longer build new `ghcr.io/nvidia/jax-te` images. Instead, `ghcr.io/nvidia/jax` will always ship with TE.


.github/container/Dockerfile.jax
```diff
@@ -45,15 +53,24 @@ ARG BASE_IMAGE
...
+COPY --from=jax-builder ${SRC_PATH_TE}-no-git ${SRC_PATH_TE}
 
-RUN pip --disable-pip-version-check install ${SRC_PATH_JAX}/dist/*.whl && \
-    pip --disable-pip-version-check install -e ${SRC_PATH_JAX} && \
-    rm -rf ~/.cache/pip/
+RUN <<EOF
+# Transformer Engine installation dependencies
+pip install --no-cache-dir pybind11 ninja packaging
+# Install JAX + Transformer Engine
+NVTE_FRAMEWORK=jax pip --disable-pip-version-check --no-cache-dir install -e \
```
terrykong, line 68: I'm wondering what you think about setting `ENV NVTE_FRAMEWORK=jax`. Since we must set that environment variable to correctly install TE, it seems more friendly for re-installs. For example, when installing t5x next, it'll look something like `pip install -e /opt/transformer-engine -e /opt/t5x` and if `NVTE_FRAMEWORK` is set in the base container, we can save a few characters when installing

yhtang: Good point.

nouiz: The env variable isn't needed anymore if jax is already installed.
By default, the setup.py file try to import all the fw and will build for those installed.

yhtang:
Completed in #371.

PR #133, opened 2023-07-19
Prevent existing jax installation from being replaced by google/jax:head when TE installs test dependencies

terrykong:
Installing TE test dependencies using the `[test]` extras will bring in praxis, which [hard-coded the JAX dependency to be that from the GitHub HEAD](https://github.com/google/praxis/blob/93894981f2337072d3671c4d9adf76c98db811f5/requirements.in#L10). This will cause pip to always grab the latest JAX source from GitHub to replace the existing installation baked into the container.

This PR introduces a temporary fix by manually installing praxis while suppressing the installation of dependencies from GitHub HEAD. As a by-product, fiddle need to be installed manually too (also done by this PR) due to it being installed from HEAD by praxis.

This fix should be reverted/dropped as soon as Praxis employs a better practice to version dependencies.

With the fix above, TE unit tests should now run.


terrykong:
Working run: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/5601516272

terrykong:
Following the same procedure as https://github.com/NVIDIA/TransformerEngine/pull/316

terrykong:
The failing tests should be addressed with this PR to TE and be fixed when that gets merged in and picked up by the nightlies: https://github.com/NVIDIA/TransformerEngine/pull/332

.github/workflows/_test_te.yaml
```diff
@@ -49,8 +49,14 @@ jobs:
...
         run: |
-          pip install pytest pytest-reportlog cuda-python
+          # TE test reqs
+          TEST_REQS="pytest-reportlog"
           TE_PATH=$(dirname $(python -c "import transformer_engine as te; print(*te.__path__)"))
+          if ! pip show praxis >/dev/null 2>&1; then
+            TEST_REQS+=" fiddle git+https://github.com/google/praxis"
+          fi
+          # SKIP_HEAD_INSTALL avoids having to install jax/fiddle at head which overwrites existing installations
+          SKIP_HEAD_INSTALLS=1 NVTE_FRAMEWORK=jax pip install -e ${TE_PATH}[test] $TEST_REQS
```
yhtang, line 59: Where is `SKIP_HEAD_INSTALLS` consumed? Doesn't seem like it's documented.

terrykong: There's a comment here https://github.com/google/praxis/blob/main/setup.py#L22-L24

yhtang (dismissed):

yhtang (approved):

PR #135, opened 2023-07-19
Adds t5x integration tests after rosetta build and makes publishing t5x results a reusable workflow

terrykong:
This adds t5x integration tests after the rosetta build.

Our current process is:

1. build t5x
2. integration test t5x (with `t5x:nightly`)
3. build rosetta-t5x

But after building the t5x distribution in rosetta-t5x, it's possible that the code change can break performance/convergence/functionality. So adding the same integration test after rosetta-t5x will help rule out the changes introduced in the rosetta-t5x build

So after this change the process is

1. build t5x
2. integration test t5x (with `t5x:nightly`)
3. build rosetta-t5x
4. **integration test t5x (again, but with `rosetta-t5x:nightly`)**


yhtang (approved):
~~Could you please keep the SSH~~

PR #137, opened 2023-07-21
Update SentencePiece model download instructions

ashors1:
Provide an alternate method for downloading Pax's SentencePiece model while we work to resolve the permissions issue that arises when downloading the file from Google Cloud

### DIFF
--- a/rosetta/rosetta/projects/pax/README.md
+++ b/rosetta/rosetta/projects/pax/README.md
@@ -32,7 +32,7 @@ Pax models require a pretrained SentencePiece model to tokenize the datasets. Th
 gsutil -m cp -r gs://mlperf-llm-public2/vocab/c4_en_301_5Mexp2_spm.model .
 ```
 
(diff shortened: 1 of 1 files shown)


rosetta/rosetta/projects/pax/README.md
```diff
@@ -32,6 +32,8 @@ Pax models require a pretrained SentencePiece model to tokenize the datasets. Th
 gsutil -m cp -r gs://mlperf-llm-public2/vocab/c4_en_301_5Mexp2_spm.model .
 ```
 
+<span style="background-color: lightgray">_NOTE_: We are aware of an existing permissions limitation that prevents users from downloading the SentencePiece model using the above instructions. We are actively working to resolve this, but in the meantime, the SP model file can be downloaded as part of the full C4 dataset download [here](https://cloud.mlcommons.org/index.php/s/dataset_c4_spm). Note that accessing the SP model using this method will require downloading the full C4 dataset. </span>
```
terrykong, line 35: Maybe it's just me, but that span isn't rendering correctly in github for me

terrykong (approved):
LGTM!

rosetta/rosetta/projects/pax/README.md
ashors1: You're right; I went ahead and removed the highlight

PR #138, opened 2023-07-24
update slurm logging on all slurm workflows to tail the log file and print partition and state info

terrykong:
Previously the only logging was the slurm state, which made it difficult to tell whether the job was running, but hung, or stuck in the queue.

The change also tails the slurm logs to stdout which should make debugging easier.


nouiz:
Self notes, we should finish this PR.

PR #139, opened 2023-07-24
Fixes pax test badge creation step to be created even if the test step or upstream workflow fails

terrykong:


maanug-nv (approved):
PAX test completion badge should now also show failure if build fails

PR #151, opened 2023-08-05
Prevent tagging of latest rosetta images if tests fail

maanug-nv:


.github/workflows/_test_rosetta.yaml
```diff
@@ -61,4 +64,36 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: ${{ env.TEST_ARTIFACT_NAME }}
-          path: /log/${{ matrix.TEST_TYPE }}-report.jsonl
\ No newline at end of file
+          path: /log/${{ matrix.TEST_TYPE }}-report.jsonl
+
+  test_status:
+    needs: rosetta-tests
+    runs-on: ubuntu-22.04
```
terrykong, line 71: should this also be `if: always()`?

maanug-nv: yes I think so. added

maanug-nv:
The same check needs to be applied to pax rosetta, correct?

terrykong:
Yes, that's true. We should gate both images on the test status. If you think this will take a while to test/clean up; let's merge this in first since rosetta-t5x is already public. If it's relatively easy to add/test it for rosetta-pax, please go ahead and add it here.

maanug-nv:
rosetta pax only tests with `_test_rosetta.yaml` which I already added the output to, since it was needed for rosetta t5x, so should be straightforward to add the tag gating logic.

.github/workflows/_test_t5x.yaml
```diff
@@ -257,6 +261,31 @@ jobs:
...
+    
+    steps:
+      - name: Download artifacts
+        uses: actions/download-artifact@v3
+
+      - name: Check status of all tests
+        shell: bash -x -e {0}
+        id: status
+        run: |
+          EXIT_STATUSES="${GITHUB_RUN_ID}-*/*-status.json"
```
yhtang, line 279: Could this reuse `_publish_badge.yaml`?

.github/workflows/_test_rosetta.yaml
```diff
@@ -61,4 +64,37 @@ jobs:
...
+          }
+          cnt_type() {
+            cat $ARTIFACTS | jq '. | select((.["$report_type"] == "TestReport") and (.when == "call") and (.outcome | contains("'${1}'"))) | .outcome' | wc -l
+          }
+          SKIPPED_TESTS=$(cnt_type skipped)
+          FAILED_TESTS=$(cnt_type failed)
+          PASSED_TESTS=$(cnt_type passed)
+          TOTAL_TESTS=$(all_outcomes | wc -l)
+          if [[ $FAILED_TESTS -eq 0 ]] && [[ $TOTAL_TESTS -gt 0 ]]; then
+            BADGE_COLOR=brightgreen
```
terrykong, line 96: is this needed?

.github/workflows/_test_t5x.yaml
terrykong: What about this: we can reuse `_publish_badge.yaml` in `_test_*.yaml` and have `_publish_badge.yaml` output the "status". We can still keep _publish_badge.yaml's usage in `nightly-*.yaml`, and in the future that can be updated to eventually use the status output of each `_test_*.yaml`

nouiz:
Just a question, would it be useful to have latest-pass-tests, latest_build? Or something equivalent?
Just I idea, I'm not sure if it is useful.

terrykong (approved):

PR #159, opened 2023-08-11
Specify CUDA base image default at GitHub Actions level only

andportnoy:
Also bump CUDA version to 12.2.0.

### DIFF
--- a/.github/container/Dockerfile.base
+++ b/.github/container/Dockerfile.base
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE
+ARG BASE_IMAGE=nvidia/cuda:12.2.0-devel-ubuntu22.04
 FROM ${BASE_IMAGE}
 
 ###############################################################################
--- a/.github/container/Dockerfile.base
+++ b/.github/container/Dockerfile.base
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE=nvidia/cuda:12.2.0-devel-ubuntu22.04
+ARG BASE_IMAGE
 FROM ${BASE_IMAGE}
 
 ###############################################################################
--- a/.github/workflows/_build_base.yaml
+++ b/.github/workflows/_build_base.yaml
@@ -70,8 +70,8 @@ jobs:
       - name: Print BASE_IMAGE
         run: |
           echo ${{ inputs.BASE_IMAGE }}
-          echo ${{ inputs.BASE_IMAGE == 'NOT SPECIFIED' }}
-          echo ${{ inputs.BASE_IMAGE != 'NOT SPECIFIED' }}
+          echo ${{inputs.BASE_IMAGE}} equals 'NOT SPECIFIED' is ${{ inputs.BASE_IMAGE == 'NOT SPECIFIED' }}
+          echo ${{inputs.BASE_IMAGE}} does not equal 'NOT SPECIFIED' is ${{ inputs.BASE_IMAGE != 'NOT SPECIFIED' }}
           echo ${{
                 inputs.BASE_IMAGE != 'NOT SPECIFIED'
                 && format('BASE_IMAGE={0}', inputs.BASE_IMAGE)
--- a/.github/workflows/_build_base.yaml
+++ b/.github/workflows/_build_base.yaml
@@ -3,6 +3,11 @@ name: ~build CUDA+Python base container
 on:
   workflow_call:
     inputs:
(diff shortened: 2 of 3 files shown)


.github/container/Dockerfile.base
```diff
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE=nvidia/cuda:12.2.0-devel-ubuntu22.04
+ARG BASE_IMAGE
```
terrykong, line 1: My two cents:

I prefer leaving the default in as that makes this Dockerfile self-contained. Now, if a user finds this Dockerfile and wants to build the base container themselves, they need to know where to look to inspect the "default value" that we assign to `BASE_IMAGE`. If the concern is so that we don't repeat ourselves, we could leave the default in the Dockerfile, and have the default of `on.workflow_call.inputs.BASE_IMAGE` be empty to signify to use the default, OR, if specified, it will overwrite

andportnoy: @nouiz had a similar feeling. I moved the explicit version back into the Dockerfile and removed the variable from Actions. If I figure out how to do this,
> leave the default in the Dockerfile, and have the default of on.workflow_call.inputs.BASE_IMAGE be empty to signify to use the default, OR, if specified, it will overwrite

I'll push an update.

.github/workflows/_build_base.yaml
```diff
@@ -67,6 +67,16 @@ jobs:
           driver-opts: |
             image=moby/buildkit:v0.12.1
 
+      - name: Print BASE_IMAGE
```
yhtang, line 70: No need to specifically print it out, as it will be part of the log of the next step.

.github/workflows/_build_base.yaml
```diff
@@ -77,8 +87,12 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           build-args: |
-            BASE_IMAGE=${{ inputs.BASE_IMAGE }}
             BUILD_DATE=${{ inputs.BUILD_DATE }}
+            ${{
```
yhtang, line 93: Give that this is an reusable workflow, i.e. an internal function called by other 'frontend' workflows, it is better to just pass in `inputs.BASE_IMAGE` and let the job error out if the value is invalid

.github/workflows/ci.yaml
```diff
@@ -10,7 +10,7 @@ on:
         type: string
         description: 'Base CUDA image:'
         required: false
-        default: 'nvidia/cuda:12.1.1-devel-ubuntu22.04'
+        default: 'NOT SPECIFIED' # use the Dockefile.base default
```
yhtang, line 13: The reason I left this default here is to help users quickly find the `nvidia/cuda` container registry and thus locate their preferred image version. If we change it to 'NOT SPECIFIED', users may not be aware of where to find these images.

yhtang: Of course, the problem, as you have correctly identified, is that we now have 2 defaults to manage, and it is easy to forget about updating one of them.

terrykong: Maybe we can fold the "default" into the description

.github/workflows/ci.yaml
```diff
@@ -77,6 +78,13 @@ jobs:
           BUILD_DATE=$(TZ='US/Los_Angeles' date '+%Y-%m-%d')
           echo "BUILD_DATE=${BUILD_DATE}" >> $GITHUB_OUTPUT
 
+      - name: Set base image
+        id: base
+        shell: bash -x -e {0}
+        run: |
+          IMG=${{ inputs.CUDA_IMAGE }}
+          echo "BASE_IMAGE=${IMG:-'NOT SPECIFIED'}" >> $GITHUB_OUTPUT
```
yhtang, line 86: If we already made `inputs.CUDA_IMAGE` default to 'NOT SPECIFIED', then we don't need to assign defaults again here.

yhtang (changes requested):

.github/workflows/ci.yaml
yhtang: > Maybe we can fold the "default" into the description

Good idea

yhtang: Also, 'LATEST' might be better than 'NOT SPECIFIED' as the latter kind of scares me out if I were to call an API with 'unspecified' arguments :grin:

PR #162, opened 2023-08-15
Run JAX unit tests on on-demand A100 runners provisioned via SLURM

yhtang:


terrykong:
This is pretty neat!

Should we `ssh ... scancel` the runner job once the unit tests are done? Seems like if the unit tests fail early, we can return the hr back to other jobs

yhtang:
> This is pretty neat!
> 
> Should we `ssh ... scancel` the runner job once the unit tests are done? Seems like if the unit tests fail early, we can return the hr back to other jobs

Good point. Since the runners are launched as ephemeral, they should de-register and exit as soon as they have processed a single job. Let me double-check if this is the actual behavior and we should be good if so.

Issue #165, opened 2023-08-18
Some runners currently have driver issues


terrykong:
This issue was first spotted by @maanug-nv .

Here is an [example run ](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/5899102023/job/16001199515) with the error:
```text
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: Auto-detected mode as 'legacy'
nvidia-container-cli: requirement error: unsatisfied condition: cuda>=12.1, please update your driver to a newer version, or use an earlier cuda container: unknown.
time="2023-08-15T21:50:48Z" level=error msg="error waiting for container: "
Error: Process completed with exit code 125.
```

The runner it [selects](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/_test_rosetta.yaml#L28) is:
```yaml
runs-on: [self-hosted, V100]
```

ashors1:
Pax nightly tests are failing with a similar error. This is from the [latest nightly](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/5924818289) tests:
```
printing enroot log file:
slurmstepd: error: pyxis:     nvidia-container-cli: requirement error: unsatisfied condition: cuda>=12.2, please update your driver to a newer version, or use an earlier cuda container
slurmstepd: error: pyxis:     [ERROR] /etc/enroot/hooks.d/98-nvidia.sh exited with return code 1
```

terrykong:
Here's an image that had this issue for troubleshooting: ghcr.io/nvidia/t5x:nightly-2023-08-20

terrykong:
Related: https://github.com/NVIDIA/JAX-Toolbox/pull/161

yhtang:
This turns out to be a driver issue on the runners.
[...]

PR #161, opened 2023-08-14
Bump CUDA version to 12.2.0-devel-ubuntu22.04

andportnoy:

PR #176, opened 2023-08-24
Revert "Bump CUDA version to 12.2.0-devel-ubuntu22.04"

terrykong:
Reverts NVIDIA/JAX-Toolbox#161

Proposing the reversion of the 12.2 bump. The images produced from the base cuda 12.2.0 image are not functioning and all tests on the CI have been failing for a few days now. I have also seen this fail on my workstation and on selene. It might be better to revert this until we know what's going on. I have attached an image to the issue below to help debug in parallel.

Related: https://github.com/NVIDIA/JAX-Toolbox/issues/165

### DIFF
--- a/.github/workflows/_build_base.yaml
+++ b/.github/workflows/_build_base.yaml
@@ -7,7 +7,7 @@ on:
         type: string
         description: 'Base CUDA image:'
         required: false
-        default: nvidia/cuda:12.2.0-devel-ubuntu22.04
+        default: nvidia/cuda:12.1.1-devel-ubuntu22.04
       BUILD_DATE:
         type: string
         description: "Build date in YYYY-MM-DD format"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,7 @@ on:
         type: string
         description: 'Base CUDA image:'
         required: false
-        default: 'nvidia/cuda:12.2.0-devel-ubuntu22.04'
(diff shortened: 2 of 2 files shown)


yhtang (commented):
Do you have any information regarding why the failure occurs? I am trying to dig a little deeper.

terrykong:
@nouiz mentioned something about compat drivers. I see that [forward compat drivers](https://docs.nvidia.com/deploy/cuda-compatibility/index.html#forward-compatibility-title) are installed:
```sh
docker run --rm ghcr.io/nvidia/t5x:nightly-2023-08-20 dpkg -l | grep cuda-compat
ii  cuda-compat-12-2                535.54.03-1                             amd64        CUDA Compatibility Platform
```
so that checks that box.

I get an issue when we add the nvidia container runtime:
```sh
# ok
docker run -it --rm ghcr.io/nvidia/t5x:nightly-2023-08-20 bash
# errors
docker run --gpus=all -it --rm ghcr.io/nvidia/t5x:nightly-2023-08-20 bash
docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: Running hook #0:: error running hook: exit status 1, stdout: , stderr: nvidia-container-cli: requirement error: unsatisfied condition: cuda>=12.2, please update your driver to a newer version, or use an earlier cuda container: unknown.
```
So maybe the nvidia container plugin is where we start looking?

Issue #167, opened 2023-08-21
EFA Support 


abhinavgoel95:
Hello team,

I have noticed that our Jax Toolbox containers do not contain the requirements needed to install [EFA](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start.html).

When following the instructions, I get the following error:
`error while loading shared libraries: libefa.so.1: cannot open shared object file: No such file or directory`

This can be a problem because EFA is needed to maximize the bandwidth.

Could someone look into this?
Thanks!

mjsML:
@abhinavgoel95 please give @yhtang admin access to AWS account to be able to help out.

yhtang:
This is a good first issue for @DwarKapex 😃

yhtang:
Forwarding more offline comments from Abhinav:

> To more succinctly summarize the problem:
> 
> expected behavior:
> root@ipp1-0274:/workspace# find / -name libefa*
> /usr/lib/x86_64-linux-gnu/pkgconfig/libefa.pc
> /usr/lib/x86_64-linux-gnu/libefa.so.1
> /usr/lib/x86_64-linux-gnu/libefa.so.1.1.39.0
> /usr/lib/x86_64-linux-gnu/libibverbs/libefa-rdmav34.so
> 
> Current behavior in [ghcr.io/nvidia/pax:nightly-2023-07-18](http://ghcr.io/nvidia/pax:nightly-2023-07-18)
> root@ipp1-0274:/# find / -name libefa*
> root@ipp1-0274:/#

PR #179, opened 2023-08-25
Add support of Amazon EFA

DwarKapex:
Address issue: EFA Support #167

    The upstream JAX container contains only MOFED NIC support. The MOFED package
    from Mellanox that we use installs *ibverbs* libraries which do not contain
    libefa*.so which are required for AWS.
    
    A temporary solution is to provide a script as a part of base container
    (/usr/local/bin/install-efa.sh) that AWS user can run inside the container
    to handle this issue.
    The script does the following:
    1. Remove all *ibverbs* and RDMA related libraries
    2. Download Amazon EFA installer
    3. Install EFA

**How to use**: in the running container run **install-efa.sh** script:
`
root@<container-id> $> install-efa.sh
`


.github/container/install-efa.sh
```diff
@@ -0,0 +1,41 @@
...
+curl -O "$AMAZON_EFA_LINK" 
+tar -xf aws-efa-installer-${EFA_INSTALLER_VERSION}.tar.gz && cd aws-efa-installer
+./efa_installer.sh -y --skip-kmod
+
+popd
+
+# check the installation is successful
+/opt/amazon/efa/bin/fi_info --version
+
+LD_LIBRARY_PATH=/opt/amazon/efa/lib:$LD_LIBRARY_PATH
```
yhtang, line 35: Environment variables must be set using the `ENV` statement in the Dockerfile to persist.

.github/container/install-efa.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# Install required packages
+apt-get install -y curl
+
+# Hack to install Amazon EFA correctly: for some reason the base docker 
+# image has incorrecly installed <libibverbs-dev>, as a result libefa* libraries
+# cannot be found under /usr/lib/x86_64-linux-gnu/
+# Temporary workaround: reinstall libibverbs-dev
+
+apt-get remove -y libibverbs-dev ibverbs-providers libibverbs1
```
yhtang, line 16: Are the existing ibverbs packages installed by the `install-ofed.sh` script?

yhtang (changes requested):

.github/container/install-efa.sh
yhtang: Per offline communication, the answer is yes, and we need to sort out a way to reconcile MOFED and EFA.

DwarKapex: Current solution is to install required libs from Ubuntu upstream apt repository explicitly.

sbhavani:
Do we have support for all the packages required to run on AWS? Here's a reference for customizing an NGC image: https://github.com/aws-samples/aws-efa-nccl-baseami-pipeline/blob/master/nvidia-efa-docker_base/ubuntu18.04/Dockerfile-cu11-pt20.10.sagemaker

.github/container/install-efa.sh
mjsML: Are they different or just a dependency conflict ?

DwarKapex:
> Do we have support for all the packages required to run on AWS? Here's a reference for customizing an NGC image: https://github.com/aws-samples/aws-efa-nccl-baseami-pipeline/blob/master/nvidia-efa-docker_base/ubuntu18.04/Dockerfile-cu11-pt20.10.sagemaker

Not sure, need to test. NGC image that you suggested uses efa-installer from amazon. That was my initial approach, but I found a conflict between efa-installer and MOFED: both installs **libibverbs** package. Interesting, that MOFED implementation of **libibverbs** does not contain libefa*, meaning, that if the installation order is MOFED -> efa-installer, MOFED implementation will be overwritten, in other way around efa-instller -> MOFED, libefa* will be unavailable.

@yhtang what is so important in MOFED? Can we use Amazon implementation instead?

.github/container/install-efa.sh
DwarKapex: They are different

.github/container/install-efa.sh
```diff
@@ -0,0 +1,45 @@
...
+# Temporary solution:
+# 1. download libibverbs-dev and ibverbs-providers that contains required libefa*
+# 2. unpack them
+# 3. copy libefa to /usr/lib
+
+WORKDIR=$(mktemp -d)
+pushd ${WORKDIR}
+
+# just in case remove related packages that contains libefa
+apt-get remove -y libibverbs-dev ibverbs-providers
```
yhtang, line 32: This would make the execution outcome of install-efa/ofed.sh order-dependent, and this order is not explicit.
Why do we have to remove the two packages if they are already installed?

yhtang (changes requested):

.github/container/install-efa.sh
yhtang: Could we at least raise a warning if the two packages are found pre-existing in the system? Or maybe add an argument `--remove-existing-ibverbs` using shell script options?

sbhavani:
I think we should have a separate image or static Dockerfile example for AWS. 

I remember seeing that some of our proprietary MOFED bits conflict with Amazon's so it's been difficult to ship both in NGC. 

+ @mjsML any thoughts?

mjsML:
sgtm

sbhavani:
I don't think we need to call out IB, it's included by default in NGC containers and doesn't need a separate tag. Could we keep the extra tag just for efa?

andportnoy:
> I think we should have a separate image or static Dockerfile example for AWS.

> nitpik, it would be on the jax container, so it will propagate to all containers: {jax,t5x,pax}.IB.py3 and {jax,t5x,pax}.EFA.py3

Use [multi-stage builds](https://docs.docker.com/build/building/multi-stage/) like we do already for the JAX container?

Then we could build two separate images using the same Dockerfile:
```bash
docker build -f Dockerfile.base --target generic --tag base-generic .
docker build -f Dockerfile.base --target aws     --tag base-aws     .
```

DwarKapex:
Make changes per today's morning discussions:         
As a workaround the strategy is the following:
1. Build the base image with MOFED support installed.
2. Add an extra script to the image (/usr/local/bin/install-efa.sh)  to be able to install Amazon EFA per request.

@yhtang **module** cmd is an interesting approach, but IMHO in this case is overkill. How to add a release note on **How to use it**?

.github/container/install-efa.sh
```diff
@@ -0,0 +1,38 @@
...
+# Download Amazon EFA package and install
+EFA_INSTALLER_VERSION=latest
+WORKDIR=$(mktemp -d)
+
+pushd ${WORKDIR}
+
+AMAZON_EFA_LINK="https://efa-installer.amazonaws.com/aws-efa-installer-${EFA_INSTALLER_VERSION}.tar.gz"
+curl -O "$AMAZON_EFA_LINK" 
+tar -xf aws-efa-installer-${EFA_INSTALLER_VERSION}.tar.gz && cd aws-efa-installer
+./efa_installer.sh -y --skip-kmod
```
DwarKapex, line 28: Add full spec of flags: `-y -g -d --skip-kmod --skip-limit-conf --no-verify` @DwarKapex

sbhavani:
> Make changes per today's morning discussions: As a workaround the strategy is the following:
> 
> 1. Build the base image with MOFED support installed.
> 2. Add an extra script to the image (/usr/local/bin/install-efa.sh)  to be able to install Amazon EFA per request.
> 
> @yhtang **module** cmd is an interesting approach, but IMHO in this case is overkill. How to add a release note on **How to use it**?

for the how to use it, we should try to keep the library versions as close to what AWS supports for it's DL containers: https://github.com/aws/deep-learning-containers/blob/master/tensorflow/training/docker/2.13/py3/cu118/Dockerfile.gpu

DwarKapex:
> for the how to use it, we should try to keep the library versions as close to what AWS supports for it's DL containers: https://github.com/aws/deep-learning-containers/blob/master/tensorflow/training/docker/2.13/py3/cu118/Dockerfile.gpu

I think, they manually increment the version of Amazon EFA installer. Don't know if it is a good idea, but if it's really matter I can keep in sync the DL containers version and ours, not a big deal (@sbhavani what do you think?). As a reference: they use version 1.24.0, we use `latest`, which is 1.25.0 as of now.

yhtang:
This is at least good for unblocking people that need EFA, so I'd say let's go ahead and merge it.

In the long run, I still believe a more systematic approach, such as the `modulefiles` solution is better in terms of unifying the container bases. There are potential drawbacks of letting users run the installer script after pulling the image, including
- making the container no longer self-contained for archival and bug-repro purposes.
- requires network connection when running the container (may be mitigated by distributing the downloaded installer package, but bloats image)
- longer pull time, bigger image.
- etc...

yhtang (dismissed):

yhtang:
Before merging, @DwarKapex could you please update the PR description so that it documents the approach taken, and that it is a stopgap solution to unblock users, while further work is still needed to harden it?

yhtang (approved):

yhtang:
/assistant summarize the key takeaways of the discussion using concise bullet points

PR #186, opened 2023-08-25
Improve description for passing custom base script to "example_slurm_pile.sub"

ashors1:


rosetta/rosetta/projects/pax/scripts/example_slurm_pile.sub
```diff
@@ -57,9 +57,12 @@ GPUS_PER_NODE=${GPUS_PER_NODE}
 PERCORE_BATCH_SIZE=${PERCORE_BATCH_SIZE}
 LOG_DIR_LOCAL=${LOG_DIR_LOCAL}
 
-export BASE_SCRIPT=${BASE_SCRIPT:-"${PAXML_DIR}/paxml/contrib/gpu/scripts_gpu/run_pile_multinode.sh"}
+if [[ -z "${BASE_SCRIPT:-}" ]]; then
```
terrykong, line 60: We should probably note somewhere (maybe readme) about this functionality. It could be confusing since:
BASE_SCRIPT="" or unset uses `paxml/contrib/gpu/scripts_gpu/run_pile_multinode.sh` under the pre-install paxml dir, but `BASE_SCRIPT=paxml/contrib/gpu/scripts_gpu/run_pile_multinode.sh`, would use the one you mounted in. So if someone mounted in their own dir, it wouldn't be clear unless looking at the source, which one they were using. Maybe also an `echo` here could help? like `echo "Using BASE_SCRIPT=$BASE_SCRIPT"` under the if statement

terrykong (approved):

terrykong:
LGTM

PR #201, opened 2023-08-29
Improve ViT README

ashors1:
- add more detailed data download instructions
- fix train/eval index paths
- add an out-of-the-box interactive pretraining example
- add a `Hardware Specifications` section

### DIFF
--- a/rosetta/rosetta/data/generate_wds_indices.py
+++ b/rosetta/rosetta/data/generate_wds_indices.py
@@ -16,6 +16,7 @@ import argparse
 import os
 
 import wds2idx
+from absl import logging
 from braceexpand import braceexpand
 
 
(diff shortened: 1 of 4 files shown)


yhtang:
/assistant

This section reads a little ambiguous to me:
> Convergence and performance has been validated on NVIDIA DGX A100 (8x A100 80G) nodes. Pretraining and fine-tuning of ViT/B-16 can be performed on a single DGX A100 80G node, but we provide both singlenode and multinode support. If running on a machine with less memory, some of the default configurations may run out of memory; if this occurs, increase your GPU count and decrease your batch size per GPU.

Can we rewrite it for better clarity in terms of multi-node support and solutions for out-of-memory problems?

terrykong (approved):

PR #204, opened 2023-08-30
Fix pytest baselines path for PAX MGMN metrics

maanug-nv:
I must have forgotten to correct this when I restructured the files.


terrykong:
Good find. This change made me realize we probably do need a step in the rosetta tests that fails b/c as of now the tests look like they all passed, but the PAX perf tests failed, but that's buried in the jobs that were run.

I made an issue to track this, but that is not blocking this change: https://github.com/NVIDIA/JAX-Toolbox/issues/205

terrykong (approved):

PR #208, opened 2023-09-05
Adds a triage job that triggers if pax MGMN or t5x MGMN tests fail

terrykong:
This is a large addition so the highlights are summarized below:
* Adds a new `_triage.yaml` workflow that gets called if T5x MGMN or Pax MGMN tests fail. It relies on a `:latest-verified` tag as the last known working image.
* The `latest-verified` tag is added to the vanilla t5x and pax images if their respective tests pass
* Adds `Dockerfile.ff` which creates a new image from a `BASE_IMAGE` where all (or some) of the dependencies are fast forwarded to the versions in `BROKEN_IMAGE`. The image layer produced by this is small since no recompilation happens. Thus, this dockerfile is currently only suitable for fast-forwarding libraries with python source only (e.g., t5x, paxml, praxis).
    * Additionally, a `ff-summary` script is added to help inspect the source offline.
* To signal failure or success, a new job called `outcome` was added to _test_t5x.yaml and _test_pax.yaml in order to detect if the overall test was a success
* Adds some utility functions for workflows in this file: `.github/workflows/util.sh`; which can be reused on the command line or in other workflows. (Was helpful in debugging the triage workflow)
* Removes the endpoint artifact upload in _publish_badge.yaml since it was causing collisions on multiple re-runs of the same workflow in a matrix. The artifact isn't really needed anyway since we upload it as a gist and it's printed in the logs.
* Files a github issue if there are failures:
    * For pax: https://github.com/NVIDIA/JAX-Toolbox/issues/215
    * For t5x: https://github.com/NVIDIA/JAX-Toolbox/issues/218


nouiz:
Should we update some user doc about this?

terrykong:
@nouiz Added a user doc here: https://github.com/NVIDIA/JAX-Toolbox/blob/triage-tool/docs/triage.md

terrykong:
Here are two "sandbox" runs to demonstrate what example triaging runs look like:
* for t5x: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6087387492 
* for pax: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6089563249

terrykong:
I've also updated the triage tool to auto file a github issue if there is a failure. Here are two examples:
* For pax: https://github.com/NVIDIA/JAX-Toolbox/issues/215
* For t5x: https://github.com/NVIDIA/JAX-Toolbox/issues/218

sharathts (approved):

nouiz:
I have questions about the created github issues:
Does this create issue only for nightly and not for PR?
If the same failures happens for many days, does it creates new issues? Append info to the existing one?

nouiz:
Also, how does it select who to assign the issue to?

terrykong:
> Also, how does it select who to assign the issue to?

Right now it's hard-coded: https://github.com/NVIDIA/JAX-Toolbox/pull/208/files#diff-2ad6ab3e3b9d04131794c79a52e5f18dc271a4bcbf5d6c08694a956e1a48e287R51-R54

>Does this create issue only for nightly and not for PR?

It only creates it for the nightly. The pre-sumit CI for PRs will not trigger the triage or the github issue filing.

> If the same failures happens for many days, does it creates new issues? Append info to the existing one?

If the failure goes on for many days, it will create new issues for each day it fails. Another option is to have one issue, and then it can be closed if an engineer deems it as "fixed", and then the triaging workflow can re-open it and add a comment with the new "summary table".

I thought creating new issues would be preferred to keep conversations self-contained since each issue may be different; but the tradeoff is there's more housekeeping everyone has to do to make sure our issue page is not cluttered with these bot triages.

nouiz:
SG to me as a start. Clearly better then what we have now.
We could just check each day if the bug is the same and close the extra issue as duplicate.
If we are able to fix them fast enough, it shouldn't be too much overhead.

.github/workflows/_sandbox.yaml
```diff
@@ -2,6 +2,40 @@ name: "~Sandbox"
 
 on:
   workflow_dispatch:
+    inputs:
```
yhtang, line 5: Don't forget to reset the sandbox :)

.github/workflows/scripts/inspect_remote_img.sh
```diff
@@ -0,0 +1,18 @@
...
+  GH_TOKEN=$1
+  IMAGE=$2
+
+  TAG=$(echo $IMAGE | rev | cut -d: -f1 | rev)
+  IMAGE_REPO=$(echo $IMAGE | rev | cut -d: -f2- | rev)
+
+  PACKAGE=$(echo $IMAGE_REPO | rev | cut -d/ -f1 | rev)
+  ORG=$(echo $IMAGE_REPO | rev | cut -d/ -f2 | rev)
+
+  top_manifest_digest=$(curl -s -H "Authorization: Bearer $(echo $GH_TOKEN | base64)" "https://ghcr.io/v2/$ORG/$PACKAGE/manifests/$TAG" | jq -r .manifests[0].digest)
```
yhtang, line 13: This might not work in the future when we start to build multi-arch containers for PAX and T5X, since `manifests[]` will contain at least two manifests corresponding to the two architectures, respectively.

yhtang (dismissed):

.github/workflows/scripts/inspect_remote_img.sh
terrykong: Gotcha. Do you have an example of an image with mutliple manifests I can use as a reference to update this code to select the right manifest?

Or are you suggesting we just merge in and fix later (create GH issue)

terrykong:
The two failures can be ignored:
*  `CI / test-pax / outcome`: is not caused by this PR, but rather surfaced b/c of this PR. So it's result should not block this PR
* `CI / build-rosetta-t5x / build`: is known to be broken on main b/c the TE PR in t5x has a merge conflict which we are resolving as we speak.

yhtang (approved):

.github/workflows/scripts/inspect_remote_img.sh
yhtang: Issue created: #263.

PR #209, opened 2023-09-06
Add Dockerfile and scripts to build PAXML on ARM64

joker-eph:


.github/container/install_lingvo_aarch64.sh
```diff
@@ -0,0 +1,65 @@
...
+EOF
+
+sed -i 's/tensorflow=/#tensorflow=/'  docker/dev.requirements.txt
+sed -i 's/tensorflow-text=/#tensorflow-text=/'  docker/dev.requirements.txt
+sed -i 's/dataclasses=/#dataclasses=/'  docker/dev.requirements.txt
+pip install -r docker/dev.requirements.txt
+pip install protobuf==3.20
+pip install patchelf
+
+# Helpful if some tests are still failing for some reason.
```
nouiz, line 59: if -> as?
Do you have some tests failing and so want to skip them for now?

joker-eph: It's been a game of dependency chasing to get all the tests to pass, I got it at some point but it was still fragile, I can give it another try

nouiz: Or just update the comment to retry this when we have a more stable installation instruction?

joker-eph: So test are "mostly" passing already right now, but they are flaky:
- ` //lingvo/core:saver_test` passes 100 times when run alone, but fails when it runs concurrently with other tests...
- `//lingvo/tasks/mt:model_test                                             FAILED in 100 out of 2000 in 57.6s` it's flaky with 5% of the case and the following log on failure:

```
AssertionError: 
Not equal to tolerance rtol=1e-06, atol=0.001
Mismatched value: a is different from b. 
not close where = (array([3]), array([0]))
not close lhs = [181.64984]
not close rhs = [181.65323]
not close dif = [0.00338869]
not close tol = [0.00118165]
dtype = float32, shape = (5, 2)
Mismatched elements: 1 / 10 (10%)
Max absolute difference: 0.00338869
Max relative difference: 1.8654727e-05
 x: array([[226.91527 ,  10.373269],
       [222.4018  ,   9.463906],
       [248.7229  ,   9.899419],...
 y: array([[226.91527 ,  10.373269],
       [222.4018  ,   9.463906],
       [248.72293 ,   9.89942 ],...
```

joker-eph: I just excluded these two tests explicitly by patching the `bazel test` invocation, that we we'll catch regression for other tests.

nouiz: Great. They seems to be tests issue, not issue with ARM. So disabling them is good.

nouiz: I don't see the code that disable those tests. Where it is?

nouiz (approved):
LGTM, only questions that shouldn't block the merge.

.github/container/Dockerfile.aarch64.pax
```diff
@@ -0,0 +1,49 @@
...
+
+
+# Lingvo
+ADD install_lingvo_aarch64.sh /opt/
+RUN /opt/install_lingvo_aarch64.sh
+
+ADD install-pax.sh /usr/local/bin
+RUN install-pax.sh
+
+# Install T5 now, it needs cargo
```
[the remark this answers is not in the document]
joker-eph: Cargo is the Rust packaging system.

`pip install t5` builds the wheel on the fly, and Rust is a dep here.

.github/container/install_lingvo_aarch64.sh
joker-eph: I didn't amend before pushing... PTAL

nouiz (approved):
LGTM.

yhtang (changes requested):
Given that this PR contains many temporary fixes that will eventually go away if the related PRs are properly upstreamed, can we create a standalone `install-pax-aarch64.sh` script, and leave the original `intall-pax.sh` as-is?

.github/container/install-pax.sh
```diff
@@ -98,17 +98,18 @@ maybe_defer_pip_install() {
...
 PRAXIS_INSTALLED_DIR=${INSTALL_DIR}/praxis
 git clone ${PRAXIS_REPO} ${PRAXIS_INSTALLED_DIR}
 pushd ${PRAXIS_INSTALLED_DIR}
 git checkout ${PRAXIS_REF}
+if [[ $(uname -m) == "aarch64" ]]; then
+  # These dependencies are broken on ARM64 right now, we handle them separately
+  sed -i 's/^tensorflow/#tensorflow/' praxis/pip_package/requirements.txt  requirements.in
+  sed -i 's/^lingvo/#lingvo/' praxis/pip_package/requirements.txt  requirements.in
+  sed -i 's/^scikit-learn/#scikit-learn/' praxis/pip_package/requirements.txt  requirements.in
+fi
```
joker-eph, line 112: This will likely stay "forever"

.github/container/install-pax.sh
```diff
@@ -117,17 +118,41 @@ PAXML_INSTALLED_DIR=${INSTALL_DIR}/paxml
...
+  # These dependencies are broken on ARM64 right now, we handle them separately
+  pip install chex==0.1.7
+  sed -i 's/^tensorflow/#tensorflow/'  paxml/pip_package/requirements.txt requirements.in
+  sed -i 's/^lingvo/#lingvo/' paxml/pip_package/requirements.txt requirements.in
+  sed -i 's/^scikit-learn/#scikit-learn/' paxml/pip_package/requirements.txt requirements.in
+  sed -i 's/^t5/#t5/' paxml/pip_package/requirements.txt requirements.in
+  sed -i 's/^jax/#jax/' paxml/pip_package/requirements.txt requirements.in
+  sed -i 's/^protobuf/#protobuf/' paxml/pip_package/requirements.txt requirements.in
+  sed -i 's/^numpy/#numpy/' paxml/pip_package/requirements.txt requirements.in
+fi
```
joker-eph, line 131: This will likely stay "forever" as well

.github/container/install-pax.sh
```diff
@@ -117,17 +118,41 @@ PAXML_INSTALLED_DIR=${INSTALL_DIR}/paxml
...
+  # array-record pip package is broken on ARM64 right now, we build it from source here.
+  cd ${INSTALL_DIR}
+  pip uninstall -y array-record || true
+  git clone http://github.com/google/array_record.git
+  pushd array_record
+  oss/build_whl.sh
+  pip install  /tmp/array_record/all_dist/array_record-*.whl --force-reinstall
+  popd
+  rm -Rf array_record /tmp/array_record
+fi
```
joker-eph, line 152: This will likely stay "forever": Google does not intend to provide an arm64 wheel here.

I could pull this into another script if you'd like though!

joker-eph:
> Given that this PR contains many temporary fixes that will eventually go away if the related PRs are properly upstreamed, 

The PRs are not referenced in install-pax but in a new script install-lingvo.sh, can you take another look to be sure you didn't mix up the two files here?


> can we create a standalone install-pax-aarch64.sh script, and leave the original intall-pax.sh as-is?

Happy to fork the script, that's what I did originally, and I ended up heavily refactoring to minimize the change because I assume we wouldn't want to maintain two scripts!
I commented in the install-pax.sh on the changes: I think they will all stay (but it's just some "sed" pre-processing before install). The diff could be reduced by moving the installation of array-record into its own script though.

yhtang:
Thanks for the clarification! Let's go ahead and merge the PR.

As a side topic, do we have a plan or any ongoing discussion regarding a more systematic approach to build/host those arm64 wheels that their original maintainers do not intend to maintain?

yhtang (approved):

joker-eph:
> Thanks for the clarification! Let's go ahead and merge the PR.

OK, I'll merge as-is then.

> As a side topic, do we have a plan or any ongoing discussion regarding a more systematic approach to build/host those arm64 wheels that their original maintainers do not intend to maintain?

We haven't talked about it, we could collaborate with Google to setup an automatic builder which could push nightly wheels for example. Or just ask them to coordinate for every new release. I'm not sure what's the best option?

PR #217, opened 2023-09-08
Add hardware specifications section to t5x readme

ashors1:


yhtang (commented):
It would be great if we can link to some artifacts from the claimed successful run.

yhtang (approved):
LGTM

PR #219, opened 2023-09-08
Update patchlists to PRs created from nvjax-svc-0

terrykong:


terrykong:
Running just the rosetta build as this will only change can only affect the rosetta workflows:
* https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6193825937
* https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6193833297

Leaving the praxis patch as a TODO in #246, as it is not high priority since that patch is not causing build failures yet.

If above pipelines succeed, then change should be good to go.

ashors1 (approved):

terrykong:
T5x build is still failing, but it's failing due to merge conflicts in the distribution and not due to this PR

PR #221, opened 2023-09-11
Make pax/praxis editable again

terrykong:
The rosetta build relies on pax/praxis being installed editable, so this re-enables the editable install in the vanilla pax build

### DIFF
--- a/.github/container/install-pax.sh
+++ b/.github/container/install-pax.sh
@@ -136,6 +136,8 @@ popd
 # required by both praxis and paxml, and fiddle at head is only required by
 # praxis
 HEAD_PACKAGES="jax fiddle"
+# We currently require installing editable (-e) to build a distribution since
+# we edit the source in place and do not re-install
 SKIP_HEAD_INSTALLS=true maybe_defer_pip_install -e ${PRAXIS_INSTALLED_DIR}
 SKIP_HEAD_INSTALLS=true maybe_defer_pip_install -e ${PAXML_INSTALLED_DIR}[gpu] $HEAD_PACKAGES
 
--- a/.github/container/install-pax.sh
+++ b/.github/container/install-pax.sh
@@ -136,8 +136,8 @@ popd
 # required by both praxis and paxml, and fiddle at head is only required by
 # praxis
 HEAD_PACKAGES="jax fiddle"
-SKIP_HEAD_INSTALLS=true maybe_defer_pip_install ${PRAXIS_INSTALLED_DIR}
-SKIP_HEAD_INSTALLS=true maybe_defer_pip_install ${PAXML_INSTALLED_DIR}[gpu] $HEAD_PACKAGES
+SKIP_HEAD_INSTALLS=true maybe_defer_pip_install -e ${PRAXIS_INSTALLED_DIR}
+SKIP_HEAD_INSTALLS=true maybe_defer_pip_install -e ${PAXML_INSTALLED_DIR}[gpu] $HEAD_PACKAGES
 
 if [[ $(uname -m) == "aarch64" ]]; then
   # array-record pip package is broken on ARM64 right now, we build it from source here.
--- a/.github/container/install-flax.sh
+++ b/.github/container/install-flax.sh
@@ -76,4 +76,6 @@ set -ex
 git clone ${FLAX_REPO} ${INSTALL_DIR}
 cd ${INSTALL_DIR}
 git checkout ${FLAX_REF}
+# We currently require installing editable (-e) to build a distribution since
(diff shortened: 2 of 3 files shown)


.github/container/install-pax.sh
```diff
@@ -136,8 +136,8 @@ popd
 # required by both praxis and paxml, and fiddle at head is only required by
 # praxis
 HEAD_PACKAGES="jax fiddle"
-SKIP_HEAD_INSTALLS=true maybe_defer_pip_install ${PRAXIS_INSTALLED_DIR}
-SKIP_HEAD_INSTALLS=true maybe_defer_pip_install ${PAXML_INSTALLED_DIR}[gpu] $HEAD_PACKAGES
+SKIP_HEAD_INSTALLS=true maybe_defer_pip_install -e ${PRAXIS_INSTALLED_DIR}
```
nouiz, line 139: Add a comment why this is needed. Otherwise, it will be broken again.

joker-eph: +1, it's highly unusual to have this outside of development!

nouiz:
LGTM

yhtang (approved):

.github/container/install-pax.sh
```diff
@@ -136,8 +136,10 @@ popd
 # required by both praxis and paxml, and fiddle at head is only required by
 # praxis
 HEAD_PACKAGES="jax fiddle"
-SKIP_HEAD_INSTALLS=true maybe_defer_pip_install ${PRAXIS_INSTALLED_DIR}
-SKIP_HEAD_INSTALLS=true maybe_defer_pip_install ${PAXML_INSTALLED_DIR}[gpu] $HEAD_PACKAGES
+# We currently require installing editable (-e) to build a distribution since
+# we edit the source in place and do not re-install
```
joker-eph, line 140: Could we reinstall instead in Rosetta?

terrykong: Yea, we should eventually try to work towards that and have a "devel" and "release" container. The problem is that re-installing will trigger some of the transitive deps to re-install. Here's an example for t5x:
```
pip install --dry-run /opt/t5x
Processing /opt/t5x
  Preparing metadata (setup.py) ... done
Collecting clu@ git+https://github.com/google/CommonLoopUtils#egg=clu (from t5x==0.0.0)
  Cloning https://github.com/google/CommonLoopUtils to /tmp/pip-install-163ha6db/clu_4faa16c7ffb4401ab249a011ee67e31c
  Running command git clone --filter=blob:none --quiet https://github.com/google/CommonLoopUtils /tmp/pip-install-163ha6db/clu_4faa16c7ffb4401ab249a011ee67e31c
  Resolved https://github.com/google/CommonLoopUtils to commit 38caf2bd46abedc87a03dc77e455508949382d1a
  Preparing metadata (setup.py) ... done
Collecting flax@ git+https://github.com/google/flax#egg=flax (from t5x==0.0.0)
  Cloning https://github.com/google/flax to /tmp/pip-install-163ha6db/flax_a6d956554a9c4d36b6adc7f8180e5a86
  Running command git clone --filter=blob:none --quiet https://github.com/google/flax /tmp/pip-install-163ha6db/flax_a6d956554a9c4d36b6adc7f8180e5a86
  Resolved https://github.com/google/flax to commit 9aeb909b39a1cb0d4a356c4ae3ec4e6841211744
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
  Preparing metadata (pyproject.toml) ... done
....
Would install flax-0.7.3 numpy-1.23.1 optax-0.1.8.dev0 seqio-0.0.17 t5x-0.0.0
```

We need some way to freeze some of these deps and having an editable install means we can update the source (so long as the transitive deps don't change).

If you've got some ideas you'd like to see, there's this issue (https://github.com/NVIDIA/JAX-Toolbox/issues/223) where we can discuss how to handle this moving forward

joker-eph: I've been using the `--no-dep` option from `pip` when I need to substitute a single package while handling the dependencies manually.

terrykong: I like that idea. Until a better way to freeze deps is implemented, we could count on `install-pax.sh` installing all the transitive deps and then in rosetta just do `pip install --no-dep /opt/paxml /opt/praxis`

This should hopefully also have minimal impact to the docker layer size, but let me double check

nouiz: As the pax nightly are currently broken, can we merge today this PR?

nouiz:
As the pax nightly are currently broken, can we merge today this PR?

PR #222, opened 2023-09-12
Redesign job reporting system

yhtang:
The JAX unit test workflow is modified to serve as a model for future unification of job reporting.

# TLDR

Each job produces its own [sitrep](https://thepersimmongroup.com/situation-report-sitrep-template/) and badges as artifacts; badges from a workflow run are packaged as a gist; sitreps are compiled into a cohesive workflow summary.

# Problems to solve

- Our current CI job reporting is fragmented, and it is hard to grasp what went wrong in a particular workflow run without carefully reading the step summaries, many of which are not readable.
- It is difficult to trace badge updates and answer questions such as "Why is the badge still green while the nightly failed."
- The `_publish_badge.yaml` workflow serves two purposes: job post-processing and (optionally) publishing a badge. In fact, it is more frequently used to do post-processing rather than publishing. Hence the inconsistency between its name and functionality can be misleading.

# Key changes

1. Each job generates a `sitrep.json` file with at least a `summary` key. The file will be uploaded as an artifact.
2. Any job can optionally generate shields.io badge endpoint JSON files.
3. A helper bash function `.github/workflows/scripts/to_json.sh` can assist in 1 & 2.
4. A new reusable workflow `.github/workflows/_upload_badge.yaml` will upload badge files (currently based on file name pattern) from all jobs within a workflow run as a **new** gist.
5. A new reusable workflow `.github/workflows/_copy_gist.yaml` for selectively copying files from one gist (e.g., the job gist) to another one (e.g., the landing page badges gist).
6. A new reusable workflow `.github/workflows/_summary.yaml` will download sitreps from all jobs and concatenate the summaries into a Markdown job summary.

# Opt-in guide

To opt-in to the new paradigm, refer to [_test_jax.yaml](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/_test_jax.yaml) or follow the steps below:
1. **job-scope changes**: in each job that runs tests, add a job step that looks like:
```
      - name: Process results and generate report
        if: success() || failure()
        shell: bash -x -e {0}
        run: |
          # bring in utility functions
          source .github/workflows/scripts/to_json.sh

          # POSTPROCESSING SCRIPTS...

          # generate sitrep
          to_json \
            summary \
            errors total_tests passed_tests failed_tests \
          > sitrep.json

          # optional: generate badge endpoint
          schemaVersion=1 \
          label="${badge_label}" \
          message="${badge_message}" \
          color="${badge_color}" \
          to_json schemaVersion label message color \
          > ${{ env.BADGE_FILENAME_FULL }}

      - name: Upload artifacts
        if: success() || failure()
        with:
          name: ${{ inputs.ARTIFACT_NAME }}-${{ matrix.GPU_ARCH }}
          path: |
            sitrep.json
            ${{ env.BADGE_FILENAME_FULL }}
```
2. **workflow-scope changes**: for an entire workflow, add the following jobs. You can optionally adjust if the jobs should run regardless of whether the test job succeeded or not using the [`if`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif) key.
```
  upload-badge:
    needs: <JOBS>
    uses: ./.github/workflows/_upload_badge.yaml
    secrets: inherit

  report:
    needs: <JOBS>
    uses: ./.github/workflows/_summary.yaml
    secrets: inherit

  publish-badge:
    needs: [upload-badge]
    if: <PUBLISHING-CONDITIONS>
    uses: ./.github/workflows/_copy_gist.yaml
    with:
      SRC_GIST_ID: ${{ needs.upload-badge.outputs.GIST_ID }}
      DST_GIST_ID: ${{ vars.BADGE_ENDPOINT_GIST_ID }}
      FILES: '.*badge.*.json'
    secrets: inherit
```
3. Remove usage of the old `.github/workflows/_publish_badge.yaml` workflow.


.github/workflows/_test_jax.yaml
```diff
@@ -58,22 +63,68 @@ jobs:
...
+            errors total_tests passed_tests failed_tests \
+            badge_label badge_color badge_message \
+          > sitrep.json
+
+          schemaVersion=1 \
+          label="${badge_label}" \
+          message="${badge_message}" \
+          color="${badge_color}" \
+          to_json schemaVersion label message color \
+          > ${{ env.BADGE_FILENAME_FULL }}
```
terrykong, line 119: What do you think about `tee`ing, here and for sitrep.json, instead of redirect? I feel like it might make troubleshooting easier

.github/workflows/nightly-jax-test-unit.yaml
```diff
@@ -46,46 +52,30 @@ jobs:
...
-  publish:
-    needs: [metadata, run-jobs]
-    strategy:
-      fail-fast: false
-      matrix:
-        GPU_ARCH: [V100, A100]
-    uses: ./.github/workflows/_publish_badge.yaml
-    if: (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success') || github.event_name == 'workflow_dispatch'
+  upload-badge:
+    needs: [run-jobs]
```
terrykong, line 64: should this be `[run]`?

.github/workflows/nightly-jax-test-unit.yaml
```diff
@@ -46,46 +52,30 @@ jobs:
...
-        echo "MESSAGE='${PASSED_TESTS}/${TOTAL_TESTS} passed'" >> $GITHUB_OUTPUT
-        echo "COLOR='${BADGE_COLOR}'" >> $GITHUB_OUTPUT
 
-  if-upstream-failed:
-    runs-on: ubuntu-latest
-    if: (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'failure') && github.event_name != 'workflow_dispatch'
-    steps:
-      - run: echo 'Upstream workflow failed, aborting run' && exit 1
+  report:
+    needs: [run-jobs]
```
terrykong, line 69: should this be `[run]`?

terrykong:
IIUC, this change is not breaking and is opt-in. Would it be possible to outline the high-level steps for migration of the old way of publishing badges and this one? We can then refer to this PR as the guide and the template for how to move the remaining badges over.

.github/workflows/_test_jax.yaml
yhtang: I thought `shell: bash -x -e {0}` should already provide a similar experience. Is my understanding correct?

.github/workflows/nightly-jax-test-unit.yaml
yhtang: Good catch 😁

yhtang:
> IIUC, this change is not breaking and is opt-in. Would it be possible to outline the high-level steps for migration of the old way of publishing badges and this one? We can then refer to this PR as the guide and the template for how to move the remaining badges over.

Exactly. I'll update the PR description with a migration workflow.

yhtang:
Since the changes are introduced to the nightly instead of the CI workflow, I'll merge the PR once https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6165605890 succeeds and bypass the 'CI' one.

yhtang:
/assistant Check the opening post for ambiguity and suggest improvements.

PR #225, opened 2023-09-12
Update Paxml patchlist with TE and config patches for improved perf

ashors1:
- adds Transformer Engine support to Pax
- updates GPU configs and default XLA flags for improved performance

### DIFF
--- a/rosetta/patchlist-paxml.txt
+++ b/rosetta/patchlist-paxml.txt
@@ -5,3 +5,5 @@
 # - External Pull Requests (These are pull requests with upstream paxml and are of the form "pull/$PULLID/head")
 # - Note: Only the first column is used as a git-ref, so anything after is a comment
 
+pull/45/head # updates configs and XLA flags
+pull/46/head # adds Transformer Engine support
--- a/rosetta/patchlist-paxml.txt
+++ b/rosetta/patchlist-paxml.txt
@@ -5,5 +5,4 @@
 # - External Pull Requests (These are pull requests with upstream paxml and are of the form "pull/$PULLID/head")
 # - Note: Only the first column is used as a git-ref, so anything after is a comment
 
-pull/45/head # updates configs and XLA flags
 pull/46/head # adds Transformer Engine support


terrykong:
Rosetta pax build/test: https://github.com/NVIDIA/JAX-Toolbox/actions/workflows/nightly-rosetta-pax-build.yaml

Above workflow should be sufficient for this change

terrykong:
Looks like the build passed (yay!), but only some of the MGMN tests passed. Unit tests are expected to fail. 

Re-run just to make sure it wasn't a one-off: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6241478583

terrykong:
To address the unit test failure: https://github.com/NVIDIA/JAX-Toolbox/pull/253

ashors1:
The failures don't appear to be one-offs. I see errors like this in the logs: 
```  File "/opt/paxml/paxml/trainer_lib.py", line 1580, in initialize_partitioned_model_states
    partitioned_vars = init_fn(prng_key)
  File "/opt/paxml/paxml/trainer_lib.py", line 1707, in call
    return pjitted_fn(*args)
AttributeError: 'NoneType' object has no attribute 'empty'
````
Investigating now

ashors1:
Google reverted their commit that broke TE: [https://github.com/google/paxml/commit/16964111a3e0ef363cb6416570285146d2299bf6](https://github.com/google/paxml/commit/16964111a3e0ef363cb6416570285146d2299bf6).
Here is the test with Pax at head: [https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6354569506](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6354569506). All tests pass besides the TP-8 test, which appears to be a problem with the test itself rather than with TE. I will work on fixing this test, but in the meantime, I think we can go ahead and merge. @terrykong what do you think?

nouiz:
Merging failing test isn't great. If we do that, can we disable the test until it is fixed?
At least, the CI results will be easy to read.

terrykong:
TP=8 looks like it failed, but it's not showing up in the metrics pytest check. Let me look into why

terrykong:
So I've reminded myself that actually two tests are not measured for perf/loss tests because they were failing some time ago. The two tests were:

1. 1DP8TP1PP
2. 2DP2TP4PP

It looks like (2) is working now, so created an issue to track adding these back in: https://github.com/NVIDIA/JAX-Toolbox/issues/272

But to @nouiz 's comment, the test is actually already omitted, so I think this is okay to merge

terrykong (approved):

ashors1:
Latest build and test: [https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6399477950](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6399477950). This run looks good. Interestingly, the TP=8 run also succeeded this time.

PR #226, opened 2023-09-13
Fix JSON type issue in the `to_json` util method

yhtang:
Use `--arg` for string and `--argjson` for numbers


yhtang:
test passed: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6168024562

PR #227, opened 2023-09-13
Increase JAX unit test concurrency from 4 to 16 tasks per GPU

yhtang:


nouiz:
Do you have any evidence that this cause speed up?
In the past, I tried higher value and I only had issues and no speed up.

PR #247, opened 2023-09-16
Added warning for shm < 1GB

### LINKED ISSUE #183: Warn user to increase shmem size when starting a container
Something like `--shm-size=1g` will help to avoid errors such as
```
Thread 787 "imagenet_fake_d" received signal SIGBUS, Bus error.
[Switching to Thread 0x7f45d7dff640 (LWP 3925)]
0x00007f6b59544f3b in ncclShmOpen(char*, unsigned long, void**, void**, int, void**) () from /usr/local/lib/python3.10/dist-packages/jaxlib/xla_extension.so
(gdb) bt
#0 0x00007f6b59544f3b in ncclShmOpen(char*, unsigned long, void**, void**, int, void**) ()
  from /usr/local/lib/python3.10/dist-packages/jaxlib/xla_extension.so
#1 0x00007f6b595436d7 in ncclProxyService(void*) ()
  from /usr/local/lib/python3.10/dist-packages/jaxlib/xla_extension.so
#2 0x00007f6b5e7adb43 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:442
#3 0x00007f6b5e83ebb4 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:100
```

chaserileyroberts:
Fixes #183


Thenerdstation:
Here is what the error looks like in the terminal when triggered: 
image

yhtang (changes requested):
Nice job. Please move the script to `https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/container/Dockerfile.base` so all containers will inherit it automatically.

rosetta/systemchecks.sh
```diff
@@ -0,0 +1,28 @@
...
+shm_size=$(df /dev/shm | awk 'NR==2 {print $2}')
+minimum_shm_size=1048576 # ~1GB in KB
+
+# Yellow-ish
+YELLOW='\033[0;33m'
+
+# Stop using terminal colors.
+STOPCOLOR='\033[0m'
+
+SHM_WARNING="${YELLOW}
```
yhtang, line 14: How about using heredoc (`EOF...EOF`) for better readability?

Thenerdstation: I was able to make it look like this

```bash
SHM_WARNING="${YELLOW}
WARNING: Your shm is currenly less than 1GB. This may cause SIGBUS errors.
To avoid this problem, you can manually set the shm size in docker with: 

$ docker run ... --shm-size=1g ...
${STOPCOLOR}"
```

and it worked. I think this version is pretty readable.

Thenerdstation:
So, it turns out that docker really does not have a good way to extend an entrypoint point directly, and instead there needs to be a system built into the parent's entrypoint.

The good news is that it already exists in the nvidia base image.
https://gitlab.com/nvidia/container-images/cuda/-/blame/master/nvidia_entrypoint.sh#L53

Adding `systemchecks.sh` to `/opt/nvidia/entrypoint.d/` makes it get executed in the above linked for-loop. 

I've also confirmed it executes even when overriding the CMD. Here is an example of it working when running `ls`

```
chase@chase-linux:~/JAX-Toolbox/rosetta
$ docker run -it --shm-size=64mb --gpus all rosetta-devel:latest ls

==========
== CUDA ==
==========

CUDA Version 12.1.1

Container image Copyright (c) 2016-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.

This container image and its contents are governed by the NVIDIA Deep Learning Container License.
By pulling and using the container, you accept the terms and conditions of this license:
https://developer.nvidia.com/ngc/nvidia-deep-learning-container-license

A copy of this license is made available in this container at /NGC-DL-CONTAINER-LICENSE for your convenience.

WARNING: Your shm is currenly less than 1GB. This may cause SIGBUS errors.
To avoid this problem, you can manually set the shm size in docker with: 

$ docker run ... --shm-size=1g ...

Dockerfile.pax  README.md    create-distribution.sh  patchlist-flax.txt   patchlist-praxis.txt  rosetta           setup.py
Dockerfile.t5x  conftest.py  docs   
``

File: .github/container/systemchecks.sh [hunk 2]
```diff
@@ -0,0 +1,23 @@
...
+STOPCOLOR='\033[0m'
+
+SHM_WARNING="${YELLOW}
+WARNING: Your shm is currenly less than 1GB. This may cause SIGBUS errors.
+To avoid this problem, you can manually set the shm size in docker with:
+
+$ docker run ... --shm-size=1g ...
+${STOPCOLOR}"
+
+if (( 0 < minimum_shm_size )); then
```
yhtang, line 21: I guess the `0` here is for debugging only? :grin:

.github/container/Dockerfile.base
```diff
@@ -50,3 +50,9 @@ ENV PATH=/opt/amazon/efa/bin:${PATH}
 ###############################################################################
 
 RUN ln -s /opt/nvidia/nsight-compute/*/host/target-linux-x64/nsys /usr/local/cuda/bin
+
+###############################################################################
+## Add the systemcheck to the entrypoint.
+###############################################################################
+
+COPY systemchecks.sh /opt/nvidia/entrypoint.d/jax-systemchecks.sh
```
yhtang, line 58: Can we use the same file name here? Something like
```
COPY system-checks.sh /opt/nvidia/entrypoint.d/
```

yhtang (changes requested):

.github/container/systemchecks.sh
chaserileyroberts: woops...

.github/container/Dockerfile.base
chaserileyroberts: Yes I can make that change. Just worried about someone accidentally overwriting that name downstream.

yhtang (approved):

PR #252, opened 2023-09-19
Aarch64 t5x

nouiz:


yhtang:
For future reference, here is the reason why the docker push commands failed for this PR:
https://docs.github.com/en/actions/security-guides/automatic-token-authentication

In short, workflows triggered on a fork runs on a read-only token as a security measure.

PR #253, opened 2023-09-19
Removes flax from rosetta-pax build and unit rosetta unit tests from rosetta-pax since those are t5x specific

terrykong:
1. Previously flax was installed as a distribution for both t5x and pax, but only t5x uses the flax distribution, so removing for now b/c pax build does not need to fail if flax distribution cannot be built
2. Unit tests in rosetta were mainly for ViT, and since that requires t5x, I am removing the unit tests for pax for now. They can be reintroduced once we introduce a mechanism for testing only t5x stuff vs. only pax stuff.

CC @maanug-nv

### DIFF
--- a/.github/workflows/nightly-rosetta-pax-build.yaml
+++ b/.github/workflows/nightly-rosetta-pax-build.yaml
@@ -83,14 +83,6 @@ jobs:
         echo "MESSAGE='${MSG}'" >> $GITHUB_OUTPUT
         echo "COLOR='${BADGE_COLOR}'" >> $GITHUB_OUTPUT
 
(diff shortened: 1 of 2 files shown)


ashors1 (approved):
LGTM. Thanks!

terrykong:
Notes: Rosetta-t5x build and the t5x tests are failing, but are not related to this change:

1. the build is failing due to a known merge conflict and 
2. t5x tests are failing due to the slurm cluster timing out, but this change should have nothing to do with t5x tests so that's irrelevant

sharathts (approved):

PR #261, opened 2023-09-25
Set NCCL_NVLS_ENABLE=0 in base jax container

terrykong:


README.md
```diff
@@ -157,6 +157,7 @@ The [JAX image](ghcr.io/nvidia/jax) is embedded with the following flags and env
 | -------------------- | ----- | ----------- |
 | `CUDA_DEVICE_MAX_CONNECTIONS` | `1` | use a single queue for GPU work to lower latency of stream operations; OK since XLA already orders launches |
 | `NCCL_IB_SL` | `1` | defines the InfiniBand Service Level ([1](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html#nccl-ib-sl)) |
+| `NCCL_NVLS_ENABLE` | `0` | Disables NVLink SHARP ([1](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html#nccl-nvls-enable)). Future releases will re-enable this feature. |
```
nouiz, line 160: Do we have a readme for the jax container? It is a good idea to have one, but can be another pr.

nouiz (approved):

README.md
terrykong: We don't, but that's a good idea!

PR #275, opened 2023-10-02
Add nightly tags for PAX CUDA 12.1 builds

yhtang:


.github/workflows/pax-cuda-121.yaml
```diff
@@ -80,6 +86,18 @@ jobs:
...
+    if: github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && inputs.PUBLISH)
+    needs: [metadata, build-pax]
+    uses: ./.github/workflows/_publish_container.yaml
+    secrets: inherit
+    with:
+      SOURCE_IMAGE: ${{ needs.build-pax.outputs.DOCKER_TAGS }}
+      TARGET_IMAGE: upstream-pax
+      TARGET_TAGS: |
+        type=raw,value=latest,priority=1000
+        type=raw,value=nightly-${{ needs.metadata.outputs.BUILD_DATE }},priority=900
```
terrykong, line 99: This looks like it will overwrite the tag that's already there. Did we want to add another tag for `12.1`?

.github/container/install_lingvo_aarch64.sh
```diff
@@ -27,11 +27,11 @@ pushd ${LINGVO_INSTALLED_DIR}
...
 git fetch origin pull/326/head:pr326
 # git fetch origin pull/328/head:pr328  ## merged upstream
-git fetch origin pull/329/head:pr329
+# git fetch origin pull/329/head:pr329
 git config user.name "JAX Toolbox"
 git config user.email "jax@toolbox"
-# git cherry-pick pr326 pr328 pr329  ## pr328 merged
-git cherry-pick pr326 pr329
+# git cherry-pick pr326 pr328 pr329  ## pr328, pr329 merged
+git cherry-pick --allow-empty pr326
```
terrykong, line 34: This looks like something we can add to [create-distribution.sh](https://github.com/NVIDIA/JAX-Toolbox/blob/distribution-keep-redundant-commits/rosetta/create-distribution.sh#L195), which would let you rewrite this section to:

```sh
cat <<EOF >/tmp/patchlist-lingvo.txt
pull/326/head
pull/328/head
pull/329/head
EOF
bash $SCRIPT_DIR/../../rosetta/create-distribution.sh \
  -d ${LINGVO_INSTALLED_DIR}
  -p /tmp/patchlist-lingvo.txt
rm -rf /tmp/patchlist-lingvo.txt
pushd ${LINGVO_INSTALLED_DIR}
```

.github/workflows/pax-cuda-121.yaml
yhtang: Good catch :)

PR #280, opened 2023-10-03
[DRAFT] Make the pax ARM dockerfile works on AMD64

nouiz:
Also move some stuff to the dockerfile vs install script.
 
 Still need to rename and adjust CI scripts.


nouiz:
Closed in favor of https://github.com/NVIDIA/JAX-Toolbox/pull/291

PR #281, opened 2023-10-03
Add sitrep report to JAX build, revise JAX unit test sitrep

yhtang:


.github/workflows/_build_jax.yaml
```diff
@@ -123,6 +136,40 @@ jobs:
             REF_JAX=${{ inputs.REF_JAX }}
             REF_XLA=${{ inputs.REF_XLA }}
 
+      - name: Generate sitrep
+        if: success() || failure()
```
DwarKapex, line 140: Maybe,
```
if: always()
```
?

yhtang: Good point :) but there is a subtle difference here:
```
always(): Causes the step to always execute, and returns true, even when canceled.
```
Here I do want the step to be skipped if the job is cancelled, hence why `success() || failure()`

PR #282, opened 2023-10-04
Fixes rosetta tests from overwriting artifacts from upstream tests

terrykong:
Artifacts have been uploaded to `${GITHUB_RUN_ID}-${TEST_CASE_NAME}`, but now that the `_test_{t5x,pax}.yaml` are potentially called twice in a single run (e.g., in ci.yaml), we can have artifacts that clobber each other. This change adds a `SUFFIX` input that will prevent them from clobbering each other

### DIFF
--- a/.github/workflows/_test_pax.yaml
+++ b/.github/workflows/_test_pax.yaml
@@ -13,6 +13,11 @@ on:
         description: Extra command line args to pass to test-pax.sh
         default: ""
         required: false
+      SUFFIX:
+        type: string
+        description: If provided, will append a suffix to the arfiact name. Helpful if re-running this reusable workflow to prevent clobbering of artifacts
+        default: ""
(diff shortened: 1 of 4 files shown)


ashors1 (approved):

terrykong:
Actually upon further investigation, this doesn't seem to be a problem currently because ci.yaml only runs _test_{pax,t5x}.yaml once, (ex: https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/ci.yaml#L228). For the public facing workflows, we haven't run into this b/c they all get their own GITHUB_RUN_ID when triggered via upstream public facing workflows

Moving this to draft since ci.yaml needs to be updated too

yhtang:
Is this issue addressed by #298? Can we close this?

PR #284, opened 2023-10-05
Updates rosetta CI to build multiarch (Arm64/amd64) for pax

terrykong:
Example images from run:
* pax: `ghcr.io/nvidia/jax-toolbox-internal:6414016731-pax-multiarch`

T5x will not be be able to be built multiarch until https://github.com/NVIDIA/JAX-Toolbox/pull/252 is merged, so for now platforms is configurable and t5x is only built for x86


.github/workflows/_build_rosetta.yaml
```diff
@@ -65,32 +73,95 @@ jobs:
...
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
         with:
           driver-opts: |
-            image=moby/buildkit:v0.10.6
+            image=moby/buildkit:v0.12.1
 
-      - name: Pull and validate BASE_IMAGE=${{ env.BASE_IMAGE }} contains BASE_LIBRARY=${{ inputs.BASE_LIBRARY }}
+      - name: Pull and validate BASE_IMAGE=${{ steps.defaults.outputs.BASE_IMAGE }} contains BASE_LIBRARY=${{ inputs.BASE_LIBRARY }}
```
yhtang, line 85: Why do we need this step?

terrykong: Sanity mainly. I was thinking in the future with OCI annotations, we could just inspect the environment vars: https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/scripts/get_remote_env.sh and then avoid us having to pull the image; but didn't implement that yet.

terrykong: Actually I failed to mention why it's required; since (rosetta) pax does not re-install paxml, having paxml already installed in editable is required for the rosetta build at the moment. Once we have reproducible builds addressed, we can probably toss this step out.

yhtang: Since the need for the check is transient and will be eliminated soon with pip-compile, I'd recommend that we simply don't do it. The build process will also fail if the library is not present, right? It should be fine as long as the wrong library does not cause a silent failure.

terrykong: Done!

yhtang:
The tests failed, likely due to external factors. Rerunning to confirm,.

terrykong:
* pax: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6460591777
* t5x: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6460592264

terrykong:
* t5x (fixed build): https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6462316511

terrykong:
Not sure why this wasn't caught earlier, but rosetta-pax fails b/c of extra installs intended only for t5x based images. Am removing those sections for now and rerunning pax build: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6464588951

yhtang (approved):

PR #287, opened 2023-10-06
Fix a few issues in MGMN tests

terrykong:
- put if-upstream-failed back
- always -> success || failure
- triage now also depends on upstream job's failure
- fix container name to upstream-{t5x,pax} for triaging and verified upstream tagging


yhtang (approved):
LGTM

Issue #306, opened 2023-10-16
Fix spurious triggers of triage


terrykong:
Example: #305

PR #309, opened 2023-10-16
Avoid skipping triage jobs

olupton:
Avoid skipping triage jobs in case of failure. See [here](https://docs.github.com/en/actions/learn-github-actions/expressions#failure-with-conditions) in the docs; without this then I believe the failed status of the earlier `run-jobs.outcome` job step caused the triage job to be skipped.

### DIFF
--- a/.github/workflows/_triage.yaml
+++ b/.github/workflows/_triage.yaml
@@ -236,7 +236,7 @@ jobs:
           name_conclusion_array=()
           while IFS= read -r line; do
             name_conclusion_array+=("$line")
-          done < <(echo "$json_data" | jq -r '.jobs[] | select(.name | startswith("test-t5x-ff") and contains("outcome")) | "\(.name)\t\(.conclusion)"')
+          done < <(echo "$json_data" | jq -r '.jobs[] | select(.name | contains("test-t5x-ff") and endswith(" / outcome")) | "\(.name)\t\(.conclusion)"')
         
           TABLE_MD=$(
           cat <<EOF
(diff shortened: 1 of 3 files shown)


terrykong:
Thanks for tackling this @olupton ! Can you confirm if this would address this: https://github.com/NVIDIA/JAX-Toolbox/issues/306 ?

Two questions:
1. Do you know why it failed in the first place and why `needs.publish-completion.outputs.STATUS != 'success'` was insufficient?
1. Would [needs.<job_id>.result](https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context:~:text=The%20result%20of%20a%20job%20that%20the%20current%20job%20depends%20on) work?

olupton:
> Can you confirm if this would address this: #306 ?

I'm not sure about that. It's not obvious to me why the triage job ran in https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6516962688, I'm not optimistic that this change will have fixed that issue.

> 1. Do you know why it failed in the first place and why `needs.publish-completion.outputs.STATUS != 'success'` was insufficient?

Based on https://docs.github.com/en/actions/learn-github-actions/expressions#failure-with-conditions:
> You can include extra conditions for a step to run after a failure, but you must still include failure() to override the default status check of success() that is automatically applied to if conditions that don't contain a status check function.

I understand there was an implicit `success()` in the `if` clause of the `triage` job, so the (intentional) failure of `run-jobs.outcome` (because Slurm jobs failed) stopped the `triage` job from running.

> 2. Would [needs.<job_id>.result](https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context:~:text=The%20result%20of%20a%20job%20that%20the%20current%20job%20depends%20on) work?

I guess yes if we made the triage job explicitly depend (`needs`) on `run-jobs` (which is the job whose overall status should be red in the case that I understand is supposed to trigger the triage job). Not sure what the advantage would be.

olupton:
Outdated.

PR #315, opened 2023-10-17
Handle multi-arch container images

### LINKED ISSUE #263: Platform-aware extraction of container image digest
The utility method [`inspect_remote_img`](https://github.com/NVIDIA/JAX-Toolbox/blob/04f240bf7b2c7295f99069dbd62a8884f223b269/.github/workflows/scripts/inspect_remote_img.sh#L19) introduced in #208  extracts the digest from the first manifest in a manifest list. This will produce an undesired outcome if the manifest list is multi-arch, e.g. contains both x86 and arm64 images. A fix would be to give the method an argument `platform`, that can take values from [this list](https://github.com/docker-library/bashbrew/blob/v0.1.2/architecture/oci-platform.go#L14-L27) such as
```
var SupportedArches = map[string]OCIPlatform{
	"amd64":    {OS: "linux", Architecture: "amd64"},
	"arm32v5":  {OS: "linux", Architecture: "arm", Variant: "v5"},
	"arm32v6":  {OS: "linux", Architecture: "arm", Variant: "v6"},
	"arm32v7":  {OS: "linux", Architecture: "arm", Variant: "v7"},
	"arm64v8":  {OS: "linux", Architecture: "arm64", Variant: "v8"},
	"i386":     {OS: "linux", Architecture: "386"},
	"mips64le": {OS: "linux", Architecture: "mips64le"},
	"ppc64le":  {OS: "linux", Architecture: "ppc64le"},
	"riscv64":  {OS: "linux", Architecture: "riscv64"},
	"s390x":    {OS: "linux", Architecture: "s390x"},

	"windows-amd64": {OS: "windows", Architecture: "amd64"},
}
```
and then return the digest of the image of the corresponding platform.

olupton:
Closes #263. See also #309 for other fixes to the pipeline that relies on these helpers.


.github/workflows/scripts/get_remote_labels.sh
```diff
@@ -2,9 +2,9 @@
 
 source $( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )/inspect_remote_img.sh
 get_remote_labels() {
-  if [[ $# -ne 2 ]]; then
-    echo '[get_remote_labels](./get_remote_labels.sh) $GH_TOKEN $IMAGE'
-    echo 'Example: get_remote_labels XXXXXXXXXXXX ghcr.io/nvidia/t5x:latest'
+  if [[ $# -ne 4 ]]; then
+    echo '[get_remote_labels](./get_remote_labels.sh) $GH_TOKEN $IMAGE $OS $ARCH'
```
terrykong, line 6: this change didn't introduce this, but could we fix this in your PR?
```suggestion
    echo 'get_remote_labels $GH_TOKEN $IMAGE $OS $ARCH'
```

terrykong (dismissed):
Other than that one comment. LGTM!

terrykong (approved):

.github/workflows/_triage.yaml
```diff
@@ -117,8 +117,8 @@ jobs:
           filter_valid_tags() {
             fgrep -x -f <(echo "$ALL_BASE_TAGS")
           }
-          start_date=${BASE_IMAGE##*nightly-}
-          end_date=$(get_build_date ${{ secrets.GITHUB_TOKEN }} $BROKEN_IMAGE )
+          start_date=$(get_build_date ${{ secrets.GITHUB_TOKEN }} $BASE_IMAGE linux amd64)
```
yhtang, line 120: Please use [the `docker-login` action](https://github.com/marketplace/actions/docker-login) as a job step to authenticate to the registry, and then have the individual scripts assume that the docker CLI is already properly granted access to the given image.

.github/workflows/scripts/inspect_remote_img.sh
```diff
@@ -1,24 +1,23 @@
...
   IMAGE_REPO=$(echo $IMAGE | rev | cut -d: -f2- | rev)
 
   PACKAGE=$(echo $IMAGE_REPO | rev | cut -d/ -f1 | rev)
   ORG=$(echo $IMAGE_REPO | rev | cut -d/ -f2 | rev)
 
-  top_manifest_digest=$(curl -s -H "Authorization: Bearer $(echo $GH_TOKEN | base64)" "https://ghcr.io/v2/$ORG/$PACKAGE/manifests/$TAG" | jq -r .manifests[0].digest)
-  config_digest=$(curl -s -H "Authorization: Bearer $(echo $GH_TOKEN | base64)" "https://ghcr.io/v2/$ORG/$PACKAGE/manifests/$top_manifest_digest" | jq -r .config.digest)
-  img_metadata=$(curl -s -L -H "Authorization: Bearer $(echo $GH_TOKEN | base64)" "https://ghcr.io/v2/$ORG/$PACKAGE/blobs/$config_digest")
-  echo $img_metadata
+  config_digest=$(docker manifest inspect -v $IMAGE | jq -r ".[] | select(.Descriptor.platform == {\"architecture\": \"$ARCH\", \"os\": \"$OS\"}) | .OCIManifest.config.digest + .SchemaV2Manifest.config.digest")
```
yhtang, line 20: Is `-v` needed here? The image digest is available without adding `-v` IIRC.

.github/workflows/scripts/inspect_remote_img.sh
```diff
@@ -1,24 +1,23 @@
...
+  curl -s -L -H "Authorization: Bearer $(echo $GH_TOKEN | base64)" "https://ghcr.io/v2/$ORG/$PACKAGE/blobs/$config_digest"
```
yhtang, line 21: Can we use the docker CLI instead of GitHub RESTful API for obtaining the digest here?

yhtang (changes requested):

.github/workflows/scripts/inspect_remote_img.sh
olupton: `-v` is needed if you want the config digest, which is what we're extracting here. It's true that the image digest is available without `-v`.

.github/workflows/scripts/inspect_remote_img.sh
olupton: This isn't obtaining the digest; it's getting the JSON configuration/metadata about the image
```
{
  "architecture": "amd64",
  "config": {
    "Env": [
      "PATH=/opt/amazon/efa/bin:/usr/local/nvidia/bin:/usr/local/cuda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
      ...
    ],
    "Entrypoint": [
      "/opt/nvidia/nvidia_entrypoint.sh"
    ],
    "Labels": {
      "maintainer": "NVIDIA CORPORATION <cudatools@nvidia.com>",
      "org.opencontainers.image.created": "2023-10-16",
...
```
it wasn't clear how to get this metadata from the Docker CLI **without** downloading the image:
```
$ docker image inspect ghcr.io/nvidia/upstream-t5x:latest
[]
Error response from daemon: No such image: ghcr.io/nvidia/upstream-t5x:latest
$ docker pull ghcr.io/nvidia/upstream-t5x:latest
... slow fetch and extraction ...
$ docker image inspect ghcr.io/nvidia/upstream-t5x:latest
[
    {
        "Id": "sha256:ab3cda0001cfdbe163fe6587e7af726690100a25025591a9a42af50929e4584d",
        ...
        "Config": {
            ...
            "Env": [
                "PATH=/opt/amazon/efa/bin:/usr/local/nvidia/bin:/usr/local/cuda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
...
```
the REST API is clearly one option that doesn't need new dependencies, and which is what we already did. Other tools like `skopeo` might also do the job, but we'd need to install that...

.github/workflows/_triage.yaml
olupton: We already use that action and sometimes use the `docker` CLI. 
See https://github.com/NVIDIA/JAX-Toolbox/pull/315#discussion_r1363562867 for the case that seems non-trivial to cover with `docker`.

.github/workflows/scripts/inspect_remote_img.sh
yhtang: The first snippet that you posted appears to be the output of `docker inspect`/`docker image inspect`.

[`docker manifest inspect`](https://docs.docker.com/engine/reference/commandline/manifest_inspect/) will not download the image and will instead query the manifest from the registry. The output will look like:
```
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 529,
         "digest": "sha256:6f8686442577a7c61863e653f121aeddfc4e3624da1c308b761830e6dcbcc73f",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      ...
   ]
}
```

olupton: Yes; that is the image digest:
```
$ docker manifest inspect ghcr.io/nvidia/upstream-t5x:latest | jq .manifests[0].digest
"sha256:4b63f505184003559014ad939912c8bb2cbc0629a3ebce8053ceeca76ec2c8e1"
$ docker manifest inspect -v ghcr.io/nvidia/upstream-t5x:latest | jq .[0].Descriptor.digest
"sha256:4b63f505184003559014ad939912c8bb2cbc0629a3ebce8053ceeca76ec2c8e1"
```
what the snippet is extracting is the config digest
```
$ docker manifest inspect -v ghcr.io/nvidia/upstream-t5x:latest | jq .[0].SchemaV2Manifest.config.digest
"sha256:ab3cda0001cfdbe163fe6587e7af726690100a25025591a9a42af50929e4584d"
```
and the `curl` command is downloading the config content using this digest. The point is to get information like
```
"org.opencontainers.image.created": "2023-10-16"
```
that is contained in the config content.

yhtang: Now I got it. I was initially confused because the script's name made me think it is a wrapper around `docker manifest inspect`. 🤣

yhtang: That said, `skopeo` is included by default in the `ubuntu-22.04` runner so why not we just use that?
I just used it in a recent PR: https://github.com/NVIDIA/JAX-Toolbox/blob/add-publish-single-arch-images/.github/workflows/_publish_container.yaml#L88

yhtang: RESTful API:
- pros: standardized (?), uses generic tooling
- cons: verbose

Skopeo:
- pros: simplifies many complicated docker operations
- cons: requires specialized software (available now in GH runners by default)

yhtang: What do you think? @olupton

olupton: If we've already taken on the dependency on `skopeo` then that con doesn't really apply. So I guess we should use it here too. I'll take a look later.

.github/workflows/_triage.yaml
```diff
@@ -85,6 +85,10 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Skopeo Login to GitHub Container Registry
```
yhtang, line 88: Turns out Skopeo uses docker CLI under the hood. So the previous `docker/login-action` step already suffices and we don't need `skopeo login` for accessing ghcr.io images.

yhtang (dismissed):
Is there a workflow where the changes are tested? Looks like we need to manually trigger one?

olupton:
> Is there a workflow where the changes are tested? Looks like we need to manually trigger one?

I think the changes are only used in the triage jobs, which need more fixes to work properly. Some of those fixes are included in https://github.com/NVIDIA/JAX-Toolbox/pull/315, but I think we also need a more stable (i.e. green) baseline for running the triage jobs to make sense.

Perhaps we can trigger the triage job manually in the meantime if you want to test this PR?

.github/workflows/_triage.yaml
olupton: Removed it in da30162

olupton:
Tried launching https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7016290505 with 
Screenshot 2023-11-28 at 09 49 13

nouiz:
What this PR is doing? I didn't understand it from the description.

olupton:
> What this PR is doing? I didn't understand it from the description.

It's fixing some of the helpers for multi-architecture OCI images (when we have both amd64 and arm64 behind the same container registry tag). We weren't seeing the problem this fixed because the auto-triage is not working right now.

PR #316, opened 2023-10-17
Updates test-t5x.sh to allow the specification of contrib/gpu configs which enable TE

terrykong:
Some code paths in the TE integration were not invoked b/c the configs used in test-t5x.sh were the upstream ones, not the GPU ones in the contrib folder. Adding this flag to switch the configs allows us to use TE layers, which we previously were not.

### DIFF
--- a/.github/container/test-t5x.sh
+++ b/.github/container/test-t5x.sh
@@ -11,19 +11,19 @@ usage() {
     echo ""
     echo "Usage: $0 [OPTIONS]"
     echo ""
-    echo "  OPTIONS                DESCRIPTION"
-    echo "  -a, --additional-args  Additional gin args to pass to t5x/train.py"
-    echo "  -b, --batch-size       Global batch size (REQUIRED)"
-    echo "  -c --use-gpu-configs   If provided uses contrib/gpu configs instead of top-level configs. Notably, gpu configs use adamw instead of adafactor"
-    echo "  -d, --dtype            Data type, defaults to bfloat16."
-    echo "  -e, --epochs           Number of epochs to run, defaults to 7."
-    echo "  --multiprocess         Enable the multiprocess GPU mode."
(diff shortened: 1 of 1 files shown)


.github/container/test-t5x.sh
```diff
@@ -14,6 +14,7 @@ usage() {
     echo "  OPTIONS                DESCRIPTION"
     echo "  -a, --additional-args  Additional gin args to pass to t5x/train.py"
     echo "  -b, --batch-size       Global batch size (REQUIRED)"
+    echo "  -c --use-gpu-configs   If provided uses contrib/gpu configs instead of top-level configs. Notably, gpu configs use adamw instead of adafactor"
```
yhtang, line 17: Perhaps call it `--use-contrib-configs` to emphasize that it is 'contribute', as GPU is always implied by our container? 😁

.github/container/test-t5x.sh
```diff
@@ -31,6 +32,7 @@ fi
 
 ADDITIONAL_ARGS=""
 BATCH_SIZE=0
+USE_CONTRIB_CONFIGS=0
```
yhtang, line 35: Ha.. you are already using this name internally.

yhtang (changes requested):

.github/container/test-t5x.sh
terrykong: Done!

.github/container/test-t5x.sh
terrykong: Yea, I think I realized that half way in and missed it in the usage. Thanks for catching

yhtang:
Pax tests are failing for this PR. @ashors1 Could you please verify if this is expected?

ashors1:
The Pax failure is expected. This is the same failure we have been seeing in our nightlies recently.

PR #321, opened 2023-10-19
Publish single-arch images alongside nightly multi-arch images 

yhtang:
`_publish_container.yaml` will now by default add single-arch images using the older docker manifest v2 schema 2 format.
This is controlled by the `EXPOSE_SINGLE_ARCH_IMAGES` input, which defaults to `true`. This will create single-arch images that looks like:
```
ghcr.io/nvidia/jax:nightly-YYYY-MM-DD-linux-amd64
ghcr.io/nvidia/jax:nightly-YYYY-MM-DD-linux-arm64
```

Note:
1. Multi-arch images will remain in OCI format.
2. This currently only applies to the nightly container tags. If needed, we can also apply it to containers created in `ci.yaml`.

Additional workflow: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6571445794. Note that the created manifest type is `application/vnd.docker.distribution.manifest.v2+json` (Docker V2.2) instead of `application/vnd.oci.image.index.v1+json` (OCI).
```
$ docker manifest inspect ghcr.io/nvidia/jax-toolbox-internal:6571445794-linux-amd64
{
        "schemaVersion": 2,
        "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
        ...
}
$ docker manifest inspect ghcr.io/nvidia/jax-toolbox-internal:6571445794-linux-arm64
{
        "schemaVersion": 2,
        "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
        ...
}
```

p.s. docker has mostly abandoned the v2s2 schema in favor of OCI, as it almost cannot be found anywhere in the latest official documentation.

### DIFF
--- a/.github/workflows/_build_base.yaml
+++ b/.github/workflows/_build_base.yaml
@@ -65,7 +65,7 @@ jobs:
         uses: docker/setup-buildx-action@v2
         with:
           driver-opts: |
-            image=moby/buildkit:v0.12.1
+            image=moby/buildkit:v0.10.6
 
       - name: Build docker images
         uses: docker/build-push-action@v4
--- a/.github/workflows/_build_base.yaml
+++ b/.github/workflows/_build_base.yaml
@@ -65,7 +65,7 @@ jobs:
         uses: docker/setup-buildx-action@v2
         with:
           driver-opts: |
-            image=moby/buildkit:v0.10.6
+            image=moby/buildkit:v0.12.1
 
       - name: Build docker images
         uses: docker/build-push-action@v4
--- a/.github/workflows/_build_jax.yaml
+++ b/.github/workflows/_build_jax.yaml
@@ -112,7 +112,7 @@ jobs:
         uses: docker/setup-buildx-action@v2
         with:
           driver-opts: |
-            image=moby/buildkit:v0.10.6
+            image=moby/buildkit:v0.12.1
 
       - name: Build docker images
         id: build
--- a/.github/workflows/_build_jax.yaml
+++ b/.github/workflows/_build_jax.yaml
@@ -112,7 +112,7 @@ jobs:
         uses: docker/setup-buildx-action@v2
         with:
           driver-opts: |
-            image=moby/buildkit:v0.12.1
+            image=moby/buildkit:v0.10.6
 
       - name: Build docker images
         id: build
--- a/.github/workflows/_build_pax.yaml
+++ b/.github/workflows/_build_pax.yaml
@@ -85,7 +85,7 @@ jobs:
         uses: docker/setup-buildx-action@v2
         with:
           driver-opts: |
-            image=moby/buildkit:v0.12.1
+            image=moby/buildkit:v0.10.6
 
       - name: Build docker images
         uses: docker/build-push-action@v4
--- a/.github/workflows/_build_pax.yaml
+++ b/.github/workflows/_build_pax.yaml
@@ -85,7 +85,7 @@ jobs:
         uses: docker/setup-buildx-action@v2
         with:
           driver-opts: |
-            image=moby/buildkit:v0.10.6
+            image=moby/buildkit:v0.12.1
 
       - name: Build docker images
         uses: docker/build-push-action@v4
--- a/.github/workflows/_build_rosetta.yaml
+++ b/.github/workflows/_build_rosetta.yaml
@@ -80,7 +80,7 @@ jobs:
         uses: docker/setup-buildx-action@v2
         with:
           driver-opts: |
-            image=moby/buildkit:v0.10.6
+            image=moby/buildkit:v0.12.1
 
       - name: Build docker images
         uses: docker/build-push-action@v4
--- a/.github/workflows/_build_rosetta.yaml
+++ b/.github/workflows/_build_rosetta.yaml
@@ -80,7 +80,7 @@ jobs:
         uses: docker/setup-buildx-action@v2
         with:
           driver-opts: |
-            image=moby/buildkit:v0.12.1
+            image=moby/buildkit:v0.10.6
 
       - name: Build docker images
         uses: docker/build-push-action@v4
--- a/.github/workflows/_publish_container.yaml
+++ b/.github/workflows/_publish_container.yaml
@@ -19,7 +19,7 @@ on:
         type: boolean
         description: 'Also expose single-arch images:'
         required: false
-        default: false
+        default: true
     outputs:
       DOCKER_TAGS:
         description: "Tags of the image published"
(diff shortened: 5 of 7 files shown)


nouiz:
We also need that for the CI container for manuel build. That can be a follow up pr.

nouiz:
Also, we should document those container and ideally also document the error that is triggered when docker is too old for the multi arch container.

yhtang:
> Also, we should document those container 

Once this PR is merged and we have the first batch of single-arch nightlies created, we will add the corresponding badges to the README.

> document the error that is triggered when docker is too old for the multi arch container.

This is already documented on our landing page: https://github.com/NVIDIA/JAX-Toolbox#faq-frequently-asked-questions

yhtang:
> We also need that for the CI container for manuel build. That can be a follow up pr.

CI already creates single-arch images, e.g. `ghcr.io/nvidia/jax-toolbox-internal:6578372511-pax-amd64`, but they are still encoded as a multi-arch manifest list that contains a single platform. This is just a natural consequence of using builtkit>0.10.6. If we downgrade buildkit, we will break how we current combine single-arch images ([1](https://github.com/NVIDIA/JAX-Toolbox/blob/81b5b3ec33db52042618263f66fba3bc4e1245e4/.github/workflows/_build_base.yaml#L135), [2](https://github.com/NVIDIA/JAX-Toolbox/blob/81b5b3ec33db52042618263f66fba3bc4e1245e4/.github/workflows/_build_jax.yaml#L236), and [3](https://github.com/NVIDIA/JAX-Toolbox/blob/81b5b3ec33db52042618263f66fba3bc4e1245e4/.github/workflows/_build_pax.yaml#L162), etc.)

What do you suggest?

yhtang:
@terrykong @ashors1 @maanug-nv Could you please investigate if these errors should block the merge?
- https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6571572593/job/17851934588?pr=321#step:9:470
- https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6571572593/job/17857688031?pr=321
- https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6571572593/job/17858432019?pr=321

nouiz:
> > Also, we should document those container
> 
> Once this PR is merged and we have the first batch of single-arch nightlies created, we will add the corresponding badges to the README.

Why not add the README update in this PR?

> > document the error that is triggered when docker is too old for the multi arch container.
> 
> This is already documented on our landing page: https://github.com/NVIDIA/JAX-Toolbox#faq-frequently-asked-questions

It doesn't document the error with docker. The current docker error is related to the shared memory usage.

nouiz:
> > We also need that for the CI container for manuel build. That can be a follow up pr.
> 
> CI already creates single-arch images, e.g. `ghcr.io/nvidia/jax-toolbox-internal:6578372511-pax-amd64`, but they are still encoded as a multi-arch manifest list that contains a single platform. This is just a natural consequence of using builtkit>0.10.6. If we downgrade buildkit, we will break how we current combine single-arch images ([1](https://github.com/NVIDIA/JAX-Toolbox/blob/81b5b3ec33db52042618263f66fba3bc4e1245e4/.github/workflows/_build_base.yaml#L135), [2](https://github.com/NVIDIA/JAX-Toolbox/blob/81b5b3ec33db52042618263f66fba3bc4e1245e4/.github/workflows/_build_jax.yaml#L236), and [3](https://github.com/NVIDIA/JAX-Toolbox/blob/81b5b3ec33db52042618263f66fba3bc4e1245e4/.github/workflows/_build_pax.yaml#L162), etc.)
> 
> What do you suggest? @nouiz

Could we change the format when we make them public?
One way I changed the format is by making a new simple container. Maybe there is a simpler way?

ashors1:
> https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6571572593/job/17858432019?pr=321
This failure is expected. The pax DP+TP+PP test failed, so we would expect the workflow status to reflect this.

yhtang:
> Why not add the README update in this PR?

We'll be unable to verify if the links work without producing the first batch of images. That said, I'm fine either way.

> 
> > > document the error that is triggered when docker is too old for the multi arch container.
> > 
> > 
> > This is already documented on our landing page: https://github.com/NVIDIA/JAX-Toolbox#faq-frequently-asked-questions
> 
> It doesn't document the error with docker. The current docker error is related to the shared memory usage.

There is a second bullet point that refers to exactly the enroot vs. OCI issue:
image

yhtang:
> > > We also need that for the CI container for manuel build. That can be a follow up pr.
> > 
> > 
> > CI already creates single-arch images, e.g. `ghcr.io/nvidia/jax-toolbox-internal:6578372511-pax-amd64`, but they are still encoded as a multi-arch manifest list that contains a single platform. This is just a natural consequence of using builtkit>0.10.6. If we downgrade buildkit, we will break how we current combine single-arch images ([1](https://github.com/NVIDIA/JAX-Toolbox/blob/81b5b3ec33db52042618263f66fba3bc4e1245e4/.github/workflows/_build_base.yaml#L135), [2](https://github.com/NVIDIA/JAX-Toolbox/blob/81b5b3ec33db52042618263f66fba3bc4e1245e4/.github/workflows/_build_jax.yaml#L236), and [3](https://github.com/NVIDIA/JAX-Toolbox/blob/81b5b3ec33db52042618263f66fba3bc4e1245e4/.github/workflows/_build_pax.yaml#L162), etc.)
> > What do you suggest? @nouiz
> 
> Could we change the format when we make them public? One way I changed the format is by making a new simple container. Maybe there is a simpler way?

The old manifest format does not inherently store the OS/arch the image is associated with. If we convert it back to the old format, we have to infer the OS/arch by either:
- pull the image and inspect the config (slow)
- use tag to encode this information

The bottom line is that it isn't a good idea to walk backward towards an older technology, as we will have to upgrade to the new OCI format again in the future and then repeat the work to adapt to the new format.

How about we create another repo in addition to `jax-toolbox-internal`, say, naming it `jax-toolbox-legacy`, and put all old-format manifests there?

yhtang:
Merging as time is of essence here for this PR to help with 23.10 testing.

The failed tests also fail on the main branch: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6578998485. In addition, given that this PR only manipulates docker manifests, it is very unlikely that it causes the failure.

nouiz:
> > It doesn't document the error with docker. The current docker error is related to the shared memory usage.
> 
> There is a second bullet point that refers to exactly the enroot vs. OCI issue: image

Yes, but the error aren't exacty the same. And people using docker won't look at the hidden error for a tool that they don't know. This is why we need both.

yhtang:
> > > It doesn't document the error with docker. The current docker error is related to the shared memory usage.
> > 
> > 
> > There is a second bullet point that refers to exactly the enroot vs. OCI issue: image
> 
> Yes, but the error aren't exacty the same. And people using docker won't look at the hidden error for a tool that they don't know. This is why we need both.

Wait, are you saying that even people using docker (not pyxis/enroot) are also having difficulties with the OCI manifest? Could you please provide an example of that error with docker?

nouiz:
> Wait, are you saying that even people using docker (not pyxis/enroot) are also having difficulties with the OCI manifest? Could you please provide an example of that error with docker?

With old docker version. But yes, maybe we can wait for it to happen again before adding it to the faq.

PR #324, opened 2023-10-19
Fix pre-submit CI from building full rosetta-t5x multiarch matrix which is not possible yet

terrykong:
CI would erroneously say arm build failed, but it failued b/c there is no ARM t5x image yet. This change disables ARM builds for rosetta-t5x until https://github.com/NVIDIA/JAX-Toolbox/pull/252 is in


terrykong:
@ashors1 the only failure in the pre-submit is the TP+DP+PP pax test that is known to fail right?

terrykong:
Looked at the tests and it's indeed only that test. So the pax test failure is known. Merging this as it fixes the other presubmit failure of arm t5x build failing b/c it doesn't exist

yhtang (approved):

PR #298, opened 2023-10-11
Add rosetta tests to ci.yaml

ashors1:
Currently, the CI workflow only runs MGMN tests on 'upstream' container. This PR adds MGMN tests on Rosetta containers to `ci.yaml`.

PR #332, opened 2023-10-24
Adds rosetta-t5x TE + no-TE tests that enable the correct configs for testing

terrykong:
- [ ] Need #298 to be merged
- [ ] Add capability to retroactively test with newer test-t5x.sh like in [t5x-wget-test](https://github.com/NVIDIA/JAX-Toolbox/tree/t5x-wget-test)
- [ ] Sets `ENABLE_TE=1` in the Dockerfile.t5x which is identical to the logic from before where it was always enabled in rosetta-t5x


terrykong:
Running a manual CI run to test with a hot-fix to the TE-t5x patch: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6643625653

terrykong:
ViT 1P8G test fails, but @ashors1 confirms that this is not an alarming failure. Also, this change should not have anything to do with ViT tests

sharathts (approved):

PR #343, opened 2023-10-25
Loosens t5x loss tests relative tolerances

terrykong:
Relaxing the relative tolerance on the loss tests since it was leading to too many false positives. For reference, deviation in loss for the t5 model can sometimes be up to 15% at the start of training with real data.


terrykong:
Also affected by #347

sharathts (approved):

PR #344, opened 2023-10-25
Adds a --seed option to test-t5x.sh to ensure determinism

terrykong:
To ensure that the tests results for a particular container are reproducible between runs, this change introduces a seed argument that sets the jax seed and dataset seed to 42. It remains configurable, but now there shouldn't be variance given the same container.

- Also fixes a typo where --steps-per-epoch wasn't in the usage doc of this script


terrykong:
Also affected by #347

terrykong:
Manual run to WAR NaN issue:
* https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6697563496
* https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6697564979
(multiple to inspect reproducibility)

yhtang (dismissed):

PR #351, opened 2023-10-31
Fix random failling tests for backend_independent on V100

DwarKapex:
Fixes randomly failures in the backend-independent section of JAX unit tests:
```
Cannot find a free accelerator to run the test  on, exiting with failure
```

Changes: limit the number of concurrent test jobs even for backend-independent tests, which do create GPU contexts.

As a clarification, `--jobs` and `--local_test_jobs` do not make a difference for our particular CI pipeline, since JAX is built in a separate CI job anyway.

References (From Reed Wanderman-Milne @ Google):

> 1. In particular, you have to set NB_GPUS, JOBS_PER_ACC, and J correctly or you can get that error (I recently got the same error by not setting those correctly)
> 2. (also I think --jobs should be --local_test_jobs in that code block, no reason to restrict the number of jobs compiling JAX)

### DIFF
--- a/.github/container/test-jax.sh
+++ b/.github/container/test-jax.sh
@@ -129,11 +129,17 @@ case "${BATTERY}" in
         EXTRA_FLAGS="--local_test_jobs=${JOBS} --test_env=JAX_TESTS_PER_ACCELERATOR=${JOBS_PER_GPU} --test_env=JAX_EXCLUDE_TEST_TARGETS=PmapTest.testSizeOverflow"
         BAZEL_TARGET="${BAZEL_TARGET} //tests:image_test_gpu //tests:scipy_stats_test_gpu"
         ;;
-    gpu|backend-independent)
+    gpu)
         JOBS_PER_GPU=8
         JOBS=$((NGPUS * JOBS_PER_GPU))
         EXTRA_FLAGS="--local_test_jobs=${JOBS} --test_env=JAX_TESTS_PER_ACCELERATOR=${JOBS_PER_GPU} --test_env=JAX_EXCLUDE_TEST_TARGETS=PmapTest.testSizeOverflow"
         BAZEL_TARGET="${BAZEL_TARGET} //tests:gpu_tests"
+        ;;
+    backend-independent)
+        JOBS_PER_GPU=4
+        JOBS=$(($NGPUS * JOBS_PER_GPU))
+        EXTRA_FLAGS="--local_test_jobs=${JOBS} --test_env=JAX_TESTS_PER_ACCELERATOR=${JOBS_PER_GPU} --test_env=JAX_EXCLUDE_TEST_TARGETS=PmapTest.testSizeOverflow"
+        BAZEL_TARGET="${BAZEL_TARGET} //tests:backend_independent_tests"
         ;;
     "")
         JOBS_PER_GPU=4
--- a/.github/container/test-jax.sh
+++ b/.github/container/test-jax.sh
@@ -137,7 +137,7 @@ case "${BATTERY}" in
         ;;
     backend-independent)
         JOBS_PER_ACC=4
-        JOBS=$(($NCPUS * JOBS_PER_ACC))
+        JOBS=$(($NGPUS * JOBS_PER_ACC))
         EXTRA_FLAGS="--local_test_jobs=${JOBS} --test_env=JAX_EXCLUDE_TEST_TARGETS=PmapTest.testSizeOverflow --test_env=JAX_TESTS_PER_ACCELERATOR=${JOBS_PER_ACC}"
         BAZEL_TARGET="${BAZEL_TARGET} //tests:backend_independent_tests"
         ;;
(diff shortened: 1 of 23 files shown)


.github/container/test-jax.sh
```diff
@@ -126,24 +126,25 @@ case "${BATTERY}" in
...
         JOBS=$((NGPUS * JOBS_PER_GPU))
-        EXTRA_FLAGS="--jobs=${JOBS} --test_env=JAX_TESTS_PER_ACCELERATOR=${JOBS_PER_GPU} --test_env=JAX_EXCLUDE_TEST_TARGETS=PmapTest.testSizeOverflow"
+        EXTRA_FLAGS="--local_test_jobs=${JOBS} --test_env=JAX_TESTS_PER_ACCELERATOR=${JOBS_PER_GPU} --test_env=JAX_EXCLUDE_TEST_TARGETS=PmapTest.testSizeOverflow"
         BAZEL_TARGET="${BAZEL_TARGET} //tests:gpu_tests"
         ;;
     backend-independent)
-        JOBS=$NCPUS
-        EXTRA_FLAGS="--jobs=${JOBS} --test_env=JAX_EXCLUDE_TEST_TARGETS=PmapTest.testSizeOverflow"
+        JOBS_PER_ACC=4
+        JOBS=$(($NGPUS * JOBS_PER_ACC))
```
yhtang, line 140: Could we use `JOBS_PER_GPU` just to be consistent with the rest of the script?

.github/container/test-jax.sh
```diff
@@ -126,24 +126,25 @@ case "${BATTERY}" in
     large)
         JOBS_PER_GPU=1
         JOBS=$((NGPUS * JOBS_PER_GPU))
-        EXTRA_FLAGS="--jobs=${JOBS} --test_env=JAX_TESTS_PER_ACCELERATOR=${JOBS_PER_GPU} --test_env=JAX_EXCLUDE_TEST_TARGETS=PmapTest.testSizeOverflow"
+        EXTRA_FLAGS="--local_test_jobs=${JOBS} --test_env=JAX_TESTS_PER_ACCELERATOR=${JOBS_PER_GPU} --test_env=JAX_EXCLUDE_TEST_TARGETS=PmapTest.testSizeOverflow"
```
yhtang, line 129: Is `local_test_jobs` making a difference in our case? My understanding is that it will make a difference if 1) we run the build and test in the same Bazel session or 2) there is a remote build involved.

.github/workflows/_build_jax.yaml
```diff
@@ -63,7 +63,7 @@ jobs:
       fail-fast: false
       matrix:
         PLATFORM: [amd64, arm64]
-    runs-on: [self-hosted, "${{ matrix.PLATFORM }}", large]
+    runs-on: [self-hosted, "${{ matrix.PLATFORM }}", small]
```
yhtang, line 66: I guess this is for debugging?

.github/workflows/_build_jax.yaml
```diff
@@ -7,7 +7,7 @@ on:
         type: string
         description: 'Base docker image that provides CUDA and Python:'
         required: false
-        default: ghcr.io/nvidia/jax-toolbox:base
+        default: ghcr.io/nvidia/jax-toolbox:${{ github.run_id }}-base-multiarch
```
yhtang, line 10: I guess this is for debugging?

.github/workflows/_sandbox.yaml
```diff
@@ -1,41 +1,99 @@
-name: "~Sandbox"
+name: CI
```
yhtang, line 1: Please don't forget to reset the sandbox before making the PR ready for merge.
```
git checkout main _sandbox.yaml
```
should do

.github/workflows/_test_jax.yaml
```diff
@@ -7,7 +7,7 @@ on:
         type: string
         description: 'JAX image built by NVIDIA/JAX-Toolbox'
         required: true
-        default: 'ghcr.io/nvidia/jax:latest'
+        default: 'ghcr.io/nvidia/jax:${{ github.run_id }}-jax-multiarch'
```
yhtang, line 10: Is this for debugging?

yhtang (changes requested):

.github/container/test-jax.sh
DwarKapex: The only difference I noticed is `local_test_jobs` work with local jobs (I assume local to the current host), and `jobs` works with jobs (I assume, all jobs including jobs on remote machines). So in out case it is the same, but guys from Google advises to use `local_test_jobs`.

.github/workflows/_test_jax.yaml
DwarKapex: Correct. Revert to main.

.github/workflows/_sandbox.yaml
DwarKapex: Yeah, I know.
Done

.github/workflows/_build_jax.yaml
DwarKapex: Correct. Reverted back

.github/container/test-jax.sh
DwarKapex: Done

.github/workflows/_build_jax.yaml
DwarKapex: Reverted

yhtang (approved):

PR #354, opened 2023-11-01
[Reproducible rosetta build] Adds support for building rosetta with local patches and an already generated patch dir

terrykong:
```mermaid
flowchart TD
    B -->|./extract-patches.sh| PG1["./patches/patchlist.txt.gen"] & PG2["./patches/*.patches"]
    PO[patchlist.txt] -->|copy| PI
    subgraph main build
    direction LR
    PI[patchlist.txt]-->|--build-arg|B[docker buildx]
    PG1
    PG2
    end

    PG1 --> |copy|PG11["./patches/patchlist.txt.gen"] 
    PG2 --> |copy|PG22["./patches/*.patches"]
    PG11 --> |copy| PG111
    PG22 --> |copy| PG222

    subgraph reproducible build
    direction LR
    PG111["./patches/patchlist.txt.gen"] --> |--build-arg| B2
    PG222["./patches/*.patches"] --> |copy| B2
    B2[docker buildx]
    end
```

This introduces these new mechanisms to the rosetta build:
1. Building with local patches
2. Generating local patches for reproducibility
3. Add a --build-arg for specifying a different patchlist file
4. Save patches as github artifacts

The above is a high level flow of how the build can be reproduced after this PR. We can now build an image from an existing image or existing patchlists by doing:
```sh
scripts/extract-patches.sh <img-name>  # Extracts generated patch dir under ./patches/
docker buildx build --build-arg T5X_PATCHLIST=patches/t5x/patchlist-t5x.txt.gen --build-arg FLAX_PATCHLIST=patches/flax/patchlist-flax.txt.gen --target rosetta --tag rosetta:latest -f Dockerfile.t5x .
```
and this should yield an identical distribution.

To create local patches, this change updates `git cherry-pick` calls to `git format-patch` + `git am` to preserve metadata in patch files.

__note__: This will not guarantee that the pip packages will be reproducible as that will need to be handled by a different PR.

- [x] update rosetta workflows with new mealkit/final mechanic
- [x] update ci.yaml


terrykong:
Manual CI run: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6722919706

terrykong:
Both the t5x/pax errors are known and related to the ncclCommInitRank issue we've seen before

yhtang (changes requested):
What is the motivation behind it? I recall you mentioned that our current patch list points to PR/branches that could change over time. Could pointing to commit SHAs, or simply reusing the container solve the issue?

Overall, I wonder if build/process reproducibility truly warrants the effort that we are currently spending, and if an 'external' approach should be taken rather than complicating the current build system.

terrykong:
> Could pointing to commit SHAs

Pointing to SHAs is a little tedious since we need to extend the list of SHAs as commits get added to git refs or PRs. If we rebase, we need to figure out what the new SHAs are

> simply reusing the container solve the issue?

I'd argue this mechanism can be reused if we had a "source container", since even if we had a source container, we still need a way to freeze and reuse the state of a git-ref used for a distribution. So this work should be compatible with the rewrite for reproducible python builds. It would look something like:
```
clone_repos()
create_distribution()
build_anything_like_jax()
pip_compile()
pip_sync()
```

> 'external' approach should be taken rather than complicating the current build system.

I see. You are thinking that a mechanism like this should live elsewhere? I imagined it would live here b/c upon our next release, we'd be able to commit `./patches/...` into a release branch (here or somewhere else) and "externally" we could build from that branch with the local patches making the distribution building reproducible.

terrykong:
Converting to draft in light of #371 . Will rebase this on top of that branch and add this feature

terrykong:
Testing using the mealkit images from the CI to trigger rosetta CI:
* t5x: 
    * ghcr.io/nvidia/jax-toolbox-internal:6857094059-upstream-t5x-amd64-mealkit
    * https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6857620737
* pax
    * amd: ghcr.io/nvidia/jax-toolbox-internal:6857094059-upstream-pax-amd64-mealkit
    * arm: ghcr.io/nvidia/jax-toolbox-internal:6857094059-upstream-pax-arm64-mealkit
    * merged: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6857609228
    * rosetta-ci: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6857614715

.github/workflows/nightly-rosetta-pax-build.yaml
```diff
@@ -123,49 +150,45 @@ jobs:
         ${{ needs.arm64.outputs.DOCKER_TAG_FINAL }}
       TARGET_IMAGE: pax
       TARGET_TAGS: |
-        type=raw,value=latest,priority=1000
-        type=raw,value=nightly-${{ needs.metadata.outputs.BUILD_DATE }},priority=900  
+        ${{ needs.test-amd64.outputs.TEST_STATUS == 'success' && 'type=raw,value=latest,priority=1000' || '' }}
```
yhtang, line 153: I'd recommend to put this logic in a 'metadata' job/jobstep, and then reference the job/jobstep output for better readability.

.github/workflows/nightly-rosetta-pax-build.yaml
```diff
@@ -123,49 +150,45 @@ jobs:
...
-    with:
-      PAX_IMAGE: ${{ needs.amd64.outputs.DOCKER_TAG_FINAL }}
-    secrets: inherit
-
-  publish-test:
-    needs: [metadata, amd64, arm64, test-pax]
+  # TODO: ARM Tests
+  publish-test-badge:
+    needs: [metadata, publish-build-badge, test-amd64]
     uses: ./.github/workflows/_publish_badge.yaml
```
yhtang, line 159: Note that `_publish_badge.yaml`` will be deprecated soon.

.github/workflows/nightly-rosetta-t5x-build-test.yaml
```diff
@@ -112,73 +142,71 @@ jobs:
...
+    needs: [metadata, amd64, arm64, test-t5x-amd64, test-vit-amd64, test-unit-amd64]
+    if: needs.metadata.outputs.PUBLISH == 'true'
+    uses: ./.github/workflows/_publish_container.yaml
+    with:
+      SOURCE_IMAGE: |
+        ${{ needs.amd64.outputs.DOCKER_TAG_FINAL }}
+        ${{ needs.arm64.outputs.DOCKER_TAG_FINAL }}
+      TARGET_IMAGE: t5x
+      TARGET_TAGS: |
+        ${{ ( needs.test-t5x-amd64.outputs.TEST_STATUS == 'success' && needs.test-vit-amd64.outputs.TEST_STATUS == 'success' && needs.test-unit-amd64.outputs.TEST_STATUS == 'success' ) && 'type=raw,value=latest,priority=1000' || '' }}
```
yhtang, line 178: Same comment regarding the placement of the logic in a separate job/jobstep.

yhtang (changes requested):

.github/workflows/_build_rosetta.yaml
```diff
@@ -95,6 +95,15 @@ jobs:
           build-args: |
             BASE_IMAGE=${{ steps.defaults.outputs.BASE_IMAGE }}
 
+      - name: Extract patches
+        run: rosetta/scripts/extract-patches.sh ${{ steps.meta.outputs.tags }}
```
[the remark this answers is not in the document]
terrykong: This line pulls all of the patches, not just one. The dir copied out is pulled from a predetermined path `/opt/rosetta/patches`. I don't foresee any reason to configure this

.github/workflows/nightly-rosetta-pax-build.yaml
terrykong: This depends on test results. `metadata` is an ancestor of the test step.

.github/workflows/nightly-rosetta-pax-build.yaml
terrykong: Would it be okay to tackle this in a subsequent PR? My thinking is we can get the reproducible build/ci changes out and then circle back to sitrep. I expect the reproducible PRs to all be flushed out within a week or two :)

rosetta/scripts/extract-patches.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd $SCRIPT_DIR
```
DwarKapex, line 4: Maybe let's not jump into this folder, but use `pushd-popd` instead?
```
pushd $SCRIPT_DIR
<the rest code>
popd
```

terrykong: pushd/popd isn't needed here b/c I only enter one directory and this is in a subshell so the calling parent shell doesn't change dirs

terrykong:
@DwarKapex @yhtang 
Closing this in favor of https://github.com/NVIDIA/JAX-Toolbox/pull/405 since I undid some of the stuff in this PR to accommodate the manifest change.

PR #358, opened 2023-11-03
Add feature to build rosetta with custom patches provided thru the web UI

terrykong:
This should help test changes more quickly since an ephemeral branch is no longer needed. One pain point before this change is you would have to create a branch using a mirror-branch with the staged test, and after the test succeeds, you would need to clean up your ephemeral branch that referenced the mirror-branch.


yhtang (changes requested):
Will this contradict our reproducible build effort, in the sense that if custom patches are provided via runtime arguments then they are hard to track?

terrykong:
In this workflow:
1. patch needs to get updated
2. rebase patch in new branch
3. create jax-toolbox branch referencing new branch
4. test
5. delete jax-toolbox branch referencing new branch
6. force push new branch into original patch branch

this change would simplify (3) and (5) since you no longer need to create a branch just for testing. All patches should still be present in the patchlists in version control.

PR #359, opened 2023-11-06
Adds test cases for t5x finetuning+eval with and without TE

terrykong:


terrykong:
Manual run: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6775778313

yhtang (changes requested):
Can we hold it off until the current CI is fully fixed?

.github/container/test-t5x.sh
```diff
@@ -7,7 +7,7 @@ print_var() {
 }
 
 usage() {
-    echo "Test T5X throughput on a fake-data Wikipedia benchmark."
+    echo "Test T5X throughput on a fake-data Wikipedia benchmark or real squad."
```
yhtang, line 10: What is 'real squad'?

yhtang (commented):
Shall we revisit this PR once #401 is merged?

.github/container/test-t5x.sh
terrykong: By real squad, I mean "Stanford Question Answering Dataset (SQuAD)". I can use this casing to avoid confusion

PR #360, opened 2023-11-06
Merges rosetta unit tests and takes off the marker which spun up another matrix job

terrykong:
This should simplify the rosetta tests and save some time since another matrix job was started for one test

### DIFF
--- a/.github/workflows/_test_rosetta.yaml
+++ b/.github/workflows/_test_rosetta.yaml
@@ -11,20 +11,17 @@ on:
     outputs:
       TEST_ARTIFACT_NAME:
         description: 'Name of the unit test artifact for downstream workflows'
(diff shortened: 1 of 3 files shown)


yhtang (approved):
LGTM

terrykong:
Running manually since CI not stable: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6787945124

Will merge if no unknown error

terrykong:
@ashors1 The ViT tests in the manual run looks like an OOM error for 8G1N and 8G2N. Is that expected?

ashors1:
> @ashors1 The ViT tests in the manual run looks like an OOM error for 8G1N and 8G2N. Is that expected?

Yes, those OOMs are also caused by the known [XLA bug](https://github.com/openxla/xla/commit/8422226150968965613e6d998f0dd0cff1b9abfe).

Issue #346, opened 2023-10-26
Random NCCL failure in nightlies


terrykong:
Reproducer:
1. Start the container
```sh
docker run --shm-size=1g --rm -it --gpus '"device=0,1"' ghcr.io/nvidia/upstream-t5x:nightly-2023-10-25 bash
```

2. Run the t5x tests
```sh
for i in 0 1; do
NCCL_DEBUG=INFO CUDA_VISIBLE_DEVICES=$i test-t5x.sh -b $((32*2)) --epochs 2 --steps-per-epoch 100 --multiprocess --additional-args "--coordinator_address=localhost:12345 --process_count=2 --process_index=$i --gin.train/utils.DatasetConfig.seed=42 --gin.train_eval/utils.DatasetConfig.seed=42 --gin.train.random_seed=42" 2>&1 | tee /tmp/log_$i &
done
```

I encounter the following error randomly:
```
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/opt/t5x/t5x/train.py", line 956, in <module>
    config_utils.run(main)
  File "/opt/t5x/t5x/config_utils.py", line 214, in run
    gin_utils.run(main)
  File "/opt/t5x/t5x/gin_utils.py", line 129, in run
    app.run(
  File "/usr/local/lib/python3.10/dist-packages/absl/app.py", line 308, in run
    _run_main(main, args)
  File "/usr/local/lib/python3.10/dist-packages/absl/app.py", line 254, in _run_main
    sys.exit(main(argv))
  File "/opt/t5x/t5x/train.py", line 890, in main
    _main(argv)
  File "/opt/t5x/t5x/train.py", line 951, in _main
    train_using_gin()
  File "/usr/local/lib/python3.10/dist-packages/gin/config.py", line 1605, in gin_wrapper
    utils.augment_exception_message_and_reraise(e, err_str)
  File "/usr/local/lib/python3.10/dist-packages/gin/utils.py", line 41, in augment_exception_message_and_reraise
    raise proxy.with_traceback(exception.__traceback__) from None
  File "/usr/local/lib/python3.10/dist-packages/gin/config.py", line 1582, in gin_wrapper
    return fn(*new_args, **new_kwargs)
[...]

PR #362, opened 2023-11-07
Add WAR for XLA NCCL bug causing OOMs

terrykong:
A stopgap for https://github.com/NVIDIA/JAX-Toolbox/issues/346


nouiz (approved):
LGTM

terrykong:
The jax ARM build issue still exists. I think it would be preferred to wait for @yhtang's fix for the ARM build before merging this since this should fix many of the CI failures and it would be good to confirm that in the presubmit

nouiz:
If one arch CI pass, I would merge it. This fix stuff, so it is really useful to have it now.
If one arch CI pass, we have a strong enough signal that this small PR is good.
This PR is independent of arch normally.

terrykong:
Triggering manually a full presubmit from jax at an 11/01 commit (last known working nightly was 11/03): https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6801767976

terrykong:
All tests pass except ViT8G2N which is a strict improvement in the CI. We can chase fixing that next.

PR #366, opened 2023-11-09
1. Add xla patch 2. Refactor build to apply the patch

DwarKapex:


.github/container/xla-neon.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/xla/pjrt/transpose_kernels.h b/xla/pjrt/transpose_kernels.h
+index 26a60c122..0a4011d9b 100644
+--- a/xla/pjrt/transpose_kernels.h
++++ b/xla/pjrt/transpose_kernels.h
+@@ -33,7 +33,7 @@ limitations under the License.
+ #define XLA_HAS_SSE2
+ #endif
+ 
+-#if defined(__ARM_NEON) && !defined(__ARM_BIG_ENDIAN)
++#if defined(__ARM_NEON) && !defined(__ARM_BIG_ENDIAN) && !defined(__CUDACC__)
```
nouiz, line 10: @nluehr Why this wasn't upstreamed? Can we just upstream it and have this fixed for anyone?

nluehr: The compilation failures are due to omissions in NVCC. They are fixed in CUDA 12.3. upstreaming the fix is OK by me, but should include a version filter as well.

PR #367, opened 2023-11-09
Patch for JAX core container (ARM64)

DwarKapex:
Add patch to XLA to be able to build JAX core container for ARM64


nouiz (approved):
LGTM. Lets merge it for now and see if we can upstream it.
What will happend if we upstream this? Will this PR make it crash?
If useful, in a follow up PR, you can only skip this patch if it doesn't apply. The equivalent of
git cherry-pick --allow-empty

PR #368, opened 2023-11-09
Update the native-fp8 guide with cudnn layer norm

kaixih:
This PR updates the guide to include the new flag to enable the cudnn layer norm.

cc. @ashors1 @terrykong @nouiz

### DIFF
--- a/rosetta/docs/NATIVE_FP8.md
+++ b/rosetta/docs/NATIVE_FP8.md
@@ -113,6 +113,7 @@ In addition to the suggested XLA flags mentioned in [this section](https://githu
 export XLA_FLAGS=" \
     --xla_gpu_enable_reduction_epilogue_fusion=false \
     --xla_gpu_enable_triton_gemm=false \
+    --xla_gpu_enable_cudnn_layer_norm=true \
     --xla_gpu_enable_cublaslt=true \
     --xla_gpu_enable_latency_hiding_scheduler=true \
     --xla_gpu_enable_async_collectives=true \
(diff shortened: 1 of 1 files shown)


rosetta/docs/NATIVE_FP8.md
```diff
@@ -113,6 +113,7 @@ In addition to the suggested XLA flags mentioned in [this section](https://githu
 export XLA_FLAGS=" \
     --xla_gpu_enable_reduction_epilogue_fusion=false \
     --xla_gpu_enable_triton_gemm=false \
+    --xla_gpu_enable_cudnn_layer_norm=true \
```
nouiz, line 116: Should we add them to the container too?

kaixih: I don't think we need. The TE uses its own custom call to the layer norm function, right?

nouiz: Maybe specify in the text bellow that some only help without TE or something like that?

terrykong:
Should we add it to https://github.com/NVIDIA/JAX-Toolbox#environment-variables with the same TE disclaimer?

Thoughts @nouiz ?

rosetta/docs/NATIVE_FP8.md
kaixih: I believe whether you set this flag or not, it shouldn't impact the TE because TE replaces the entire transformer layer regardless. This document specifically addresses native FP8 support (i.e., when ENABLE_TE=0). For a concise summary of recommended TE XLA flags, I think people should refer to `rosetta/rosetta/projects/pax/README.md`. That was why we moved out this guide from there.

nouiz: SG.

nouiz (approved):

PR #369, opened 2023-11-09
Updates upstream test badges

terrykong:
We only reported "completion" badges, which mistakenly reported "n/a" if there was one failure. We also performed loss/perf tests for upstream but did not report it. 

This change simplifies the workflows by removing that extra completion badge step and re-uses the badge within `_test_t5x.yaml` and `_test_pax.yaml`. It also renames the badge labels to "Upstream tests" for upstream-{t5x,pax} and "Tests" for rosetta ones.

- [] Dry run with fake badge endpoint names

### DIFF
--- a/.github/workflows/_test_pax.yaml
+++ b/.github/workflows/_test_pax.yaml
@@ -13,6 +13,11 @@ on:
         description: Extra command line args to pass to test-pax.sh
         default: ""
         required: false
+      PUBLISH:
+        type: boolean
+        description: Publish badge?
+        default: false
+        required: false
     outputs:
       TEST_STATUS:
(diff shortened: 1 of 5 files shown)


terrykong:
Manual runs: 
* t5x: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6817018979
* pax: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6817021286

ashors1 (approved):
LGTM! Thanks!

yhtang (changes requested):
Is this PR going to be superceded by #401?

PR #370, opened 2023-11-09
Deletes the global git config file in rosetta images to avoid silently committing as the jax@nvidia.com user

terrykong:


yhtang:
@terrykong Do you think it would be useful to create some `cleanup.sh` script for running as the last step in every Dockerfile to remove such information?

Issue #338, opened 2023-10-25
merge x86 and ARM for pax


yhtang:

yhtang:
https://github.com/NVIDIA/JAX-Toolbox/pull/405/files#r1412660231

nouiz:
We don't test pax anymore.

Issue #229, opened 2023-09-13
Migrate to the sitrep system


yhtang:

yhtang:
/assistant rewrite the task lists above with more context so that each line is self-contained and is suitable as a standalone issue title

PR #371, opened 2023-11-10
Use pip-compile to help with consistent Python dependency resolution

### LINKED ISSUE #223: Reproducible builds
Starting this discussion with an initial proposal of how to support reproducible builds to freeze the ecosystem state leading up to a release.

# Problem
At the moment, we have not implemented a mechanism to pin version `X` of jax, `Y` of paxml, and `Z` of a transitive dependency of paxml in our containers; where X, Y, and Z can be versions on pypi, git-refs, or even a full distribution. Currently, we rely on nightly images to pin state, but these are not reproducible since running the same build script may install more recent libraries or transitive dependencies that break functionality.

This is an issue when we want to rebuild a container for a release, because we become cautious of changing build commands in previous layers (e.g., jax or pax) in fear that the build would pick up an unintentional change.

## Requirements
There are two things we need to freeze:
1. The version of a python package, e.g., pax, jax, transformer-engine
2. The distribution of a library, e.g., pax + (all of our patches)

## Proposal
Here is my initial proposal, which we can iterate on.

### Freezing python package versions
There are mature solutions to handle python package dependencies like pip-tools (pip-compile) and we can make use of those tools here. We can have:
* requirements-jax.in
* requirements-pax.in
* requirements-t5x.in

Where the jax images install `requirements-jax.in` and the t5x images installs `pip_compiled(requirements-jax.in, requirements-t5x.in)`.

### Freezing the state of the nightly
Previously, we froze the state of all upstream repos by building it in a docker image. In order to re-build that image, we will need to update our CI to do something like the following:

> (get git-sha of all upstream repos nightly) -> (write all SHA's to a manifest) -> (clone repos you need from the SHA-manifest) -> build

### Freezing the state of distributions
We already use patchlists, so we can continue to do this. One point of improvement though is perhaps we need a way to guarantee that the git-ref in the patchlist has not been force-updated. 

Since a git-ref (branch or PR) can be rebased, once an engineer updates the git-ref, you can't build the old build anymore. We need to store the patches somewhere. Perhaps we can freeze the diff of the git-ref in some location alongside the `requirements-*.in` and the "SHA manifest".

### Nightly CI
We can have the nightly CI freeze all of this metadata (it would also build from this metadata), and make write it somewhere in version control so that we can fork it for releases.

### Metadata location
There a couple of options we have for metadata location; here are two:
1. Within the JAX-Toolbox git repo (have nightly CI commit to `metadata` branch)
    * pros: simple; authoring CI is easier; will not clutter main branch
    * cons: can cause the repo to grow (especially if we store patch diffs here too)
2. In an auxiliary git repo
    * pros: keeps JAX-Toolbox history clean
    * cons: adds complexity to the process since atomicity is more difficult to guarantee; engineers need to be aware of multiple repos

I'm more in favor or (1) due to simplicity, but am open to other options. FWIW with (1), we can merge in (main + metadata) to form a release branch.

#### Installation/Building
It may be obvious, but since some of these repos have more complex build steps; often requiring `apt` packages, or python packages not listed in their `install_requires`, we'll also need one or more `./install-jax-req.sh` to setup or build any wheels we need.

This could potentially live alongside the "SHA manifest"

```[tasklist]
### Tasks
- [ ] Investigate pip-compile @terrykong 
```

### LINKED ISSUE #230: Add sitrep generation for the base container build workflow

### LINKED ISSUE #231: Add sitrep generation for the JAX core container build workflow

### LINKED ISSUE #232: Add sitrep generation for the upstream T5X container build workflow

### LINKED ISSUE #233: Add sitrep generation for the upstream PAX container build workflow

### LINKED ISSUE #328: Dockerfile.jax,pax and t5x try install flax.
The output of some grep:
```
Dockerfile.jax:RUN { pip install flax || true; } && rm -rf ~/.cache/*
Dockerfile.pax.amd64:ADD install-flax.sh /usr/local/bin
Dockerfile.pax.amd64:install-flax.sh --defer
Dockerfile.t5x:ADD install-flax.sh /usr/local/bin
Dockerfile.t5x:install-flax.sh
```

Dockerfile.jax try to install flax, but allows it to fail.
Then t5x and pax Dockerfile install it via install-flax.sh

We should be more consistent. Should we not try to install it in the JAX container?
Or we should use the install-flax.sh file?
This also seem to bypass the pinning of the flax version in the JAX container.

### LINKED ISSUE #337: Fix pip install version conflicts between upper/lower containers

yhtang:
# Summary

- All Python packages, except for a few build dependencies, are now installed using **pip-tools**.
- The JAX and upstream T5X/PAX containers are now built in a two-stage procedure:
  1. The **'meal kit'** stage: source packages are downloaded, wheels built if necessary (for TE, tensorflow-text, lingvo, etc.), but **no** package is installed. Instead, manifest files are created in the `/opt/pip-tools.d` folder to instruct which packages shall be installed by pip-tools. The stage is named due to its similarity in how ingredients in a meal kit are prepared while deferring the final cooking step.
  2. The **'final'** (cooking🔥) stage: this is when pip-tools collectively compile the manifests from the various container layers and then sync-install everything to exactly match the resolved versions.
- Note that downstream containers will **build on top of the meal kit image of its base container**, thus ensuring all packages and dependencies are installed exactly once to avoid conflicts and image bloating.
- The meal kit and final images are published as
  - mealkit: `ghcr.io/nvidia/image:mealkit` and `ghcr.io/nvidia/image:mealkit-YYYY-MM-DD`
  - final: `ghcr.io/nvidia/image:latest` and `ghcr.io/nvidia/image:nightly-YYYY-MM-DD`

# Additional changes to the workflows

- `/opt/jax-source` is renamed to `/opt/jax`. The `-source` suffix is only added to packages that needs compilation, e.g. XLA and TE.
- The CI workflow is now matricized against CPU arch.
- The reusable `_build_*.yaml` workflows are simplified to build only one image for a single architecture at a time. The logic for creating multi-arch images is relocated into the `_publish_container.yaml` workflows and involved during the nightly runs only.
- TE is now built as a wheel and shipped in the JAX core meal kit image.
- TE unit tests will be performed using the upstream-pax image due to the dependency on praxis.
- Build workflows now produce sitreps following the paradigm of #229.
- Removed the various one-off workflows for pinned CUDA/JAX versions.
- Refactored the PAX arm64 Dockerfile in preparation for #338

# What remains to be done

- [ ] Update the Rosetta container build + test process to use the upstream T5X/PAX mealkit (ghcr.io/nvidia/upstream-t5x:mealkit, ghcr.io/nvidia/upstream-pax:mealkit) containers

# Reviewing tips

This PR requires a multitude of reviewers due to its size and scope. I'd truly appreciate code owners to review any changes related to their previous contributions. An incomplete list of reviewer-scope is:
- @terrykong, @ashors1, @sharathts, @maanug-nv: Rosetta, TE, T5X and PAX MGMN tests
- @nouiz: JAX, TE and T5X build
- @joker-eph: PAX arm64 build
- @nluehr: Base image, NCCL, PAX
- @DwarKapex: base/JAX/XLA build, workflow logic

Closes #223
Closes #230 
Closes #231 
Closes #232 
Closes #233 
Closes #271
Fixes #328
Fixes #337 

Co-authored-by: Terry Kong <terryk@nvidia.com>


nouiz:
What is the replacement for `Removed the various one-off workflows for pinned CUDA/JAX versions.` in the description? Do we still build some multi-version CUDA containers?

yhtang:
> What is the replacement for `Removed the various one-off workflows for pinned CUDA/JAX versions.` in the description? Do we still build some multi-version CUDA containers?

There is no replacement yet. Do we still need these containers? I remember these were built for:
- Selene testing: Selene is getting CUDA 12.2 very soon.
- 23.11 release pinning: it is done, and future release will be handled by dl/dgx/jax

.github/container/Dockerfile.base
```diff
@@ -17,13 +18,28 @@ RUN apt-get update && \
...
         python3-pip \
+        liblzma-dev \
         wget \
     && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
-RUN pip install --upgrade --no-cache-dir pip
+RUN <<"EOF" bash -ex
+git config --global user.name "JAX Toolbox"
+git config --global user.email "jax@nvidia.com"
```
terrykong, line 35: There was some feedback earlier about how this could silently have commits become authored by the jax@nvidia.com user, which caused some issues during CLA presubmits. I reverted it for rosetta here: https://github.com/NVIDIA/JAX-Toolbox/pull/370. What do you think about reverting it here after it's no longer needed?

.github/container/Dockerfile.pax.arm64
```diff
@@ -1,65 +1,160 @@
...
++  bazel test $@ lingvo/core/... --  -//lingvo/tasks/mt:model_test -//lingvo/core:saver_test
+ fi
+
+ DST_DIR="/tmp/lingvo/dist"
+EOF
+
+pip install tensorflow_datasets==4.9.2 auditwheel tensorflow==2.13.0 /opt/tensorflow_text*.whl
+sed -i 's/tensorflow=/#tensorflow=/'  docker/dev.requirements.txt
+sed -i 's/tensorflow-text=/#tensorflow-text=/'  docker/dev.requirements.txt
+sed -i 's/dataclasses=/#dataclasses=/'  docker/dev.requirements.txt
```
terrykong, line 78: The previous stages seem to relate to TF, what is the significance of removing dataclasses as a dep?

.github/container/get-source.sh
```diff
@@ -0,0 +1,101 @@
...
+## -e /opt/flax
+
+## Parse command-line arguments
+
+usage() {
+    echo "Usage: $0 [OPTION]..."
+    echo "  -d, --dir PATH         [Required] Local path to check out the source code."
+    echo "  -f, --from URL         [Required] URL of the source repo."
+    echo "  -h, --help             Print usage."
+    echo "  -i, --install          Install the package immediately using pip install."
```
terrykong, line 15: Looks like --install and --manifest are mutually exclusive; was this intended? Under what scenario would we use --install?

.github/container/install-pax.sh
```diff
@@ -1,101 +1,3 @@
...
-      fi
-    done
-  else
-    pip install $@
-  fi
-}
-
 set -ex
 
 ## Install Praxis
```
terrykong, line 3: Should this file be deleted?

.github/container/Dockerfile.pax.arm64
yhtang: This is inherited from https://github.com/NVIDIA/JAX-Toolbox/blob/4e4e137afd884ebfa748cec636ea751dc8d7ff7a/.github/container/install_lingvo_aarch64.sh#L41. @joker-eph Do you still remember the reason for doing so?

.github/container/Dockerfile.base
yhtang: What is the exact CLA issue? Can we use some other username/email that do not annoy the CLA presubmits?

.github/container/install-pax.sh
yhtang: Yes indeed

.github/container/Dockerfile.pax.arm64
nouiz: It was needed at some point to install on GH. Maybe that changed, the only way to know is to test it.

nouiz:
This PR is gigantic. If we can't merge it in2-3 days, we should try to split it.
If we get there, we could do it per container.
Make only one PR for the base infra and split the JAX container into 2 steps, meal + final. The other stay based on final. Then one PR for each other containers that start to use the meal container instead of final.

yhtang:
> This PR is gigantic. If we can't merge it in2-3 days, we should try to split it. If we get there, we could do it per container. Make only one PR for the base infra and split the JAX container into 2 steps, meal + final. The other stay based on final. Then one PR for each other containers that start to use the meal container instead of final.

The challenge of splitting would be to rewire the workflows between passing the mealkit and final tags.
Will someone be able to do it, as I'm technically OOTO?

.github/container/Dockerfile.base
terrykong: TJ was using our containers to commit code and this global username/email got used to author the commits. When a PR was made, google's CLA bot flagged it as a developer who hadn't signed a CLA.

.github/workflows/_publish_container.yaml
```diff
@@ -85,26 +92,26 @@ jobs:
             docker buildx imagetools create --tag $tag ${{ steps.get-manifests.outputs.manifests }}
```
terrykong, line 92: Does this work instead?

`docker buildx imagetools create $(echo "${{ inputs.SOURCE_IMAGE }}" | tr '\n' ' ') --tag ${{ inputs.TARGET_IMAGE }}`

.github/workflows/nightly-te-test.yaml
```diff
@@ -12,7 +12,7 @@ on:
         type: string
         description: 'JAX-TE image build by NVIDIA/JAX-Toolbox'
         required: true
-        default: 'ghcr.io/nvidia/jax-te:latest'
+        default: 'ghcr.io/nvidia/jax:latest'
```
terrykong, line 15: Same Q about upstream-pax?
```suggestion
        default: 'ghcr.io/nvidia/upstream-pax:latest'
```

.github/workflows/nightly-te-test.yaml
```diff
@@ -25,7 +25,7 @@ permissions:
   packages: write # to upload container
 
 env:
-  DEFAULT_JAX_TE_IMAGE: 'ghcr.io/nvidia/jax-te:latest'
+  DEFAULT_JAX_TE_IMAGE: 'ghcr.io/nvidia/jax:latest'
```
terrykong, line 28: Same Q about upstream-pax
```suggestion
  DEFAULT_JAX_TE_IMAGE: 'ghcr.io/nvidia/upstream-pax:latest'
```

.github/workflows/weekly-base-build.yaml
```diff
@@ -35,21 +35,69 @@ jobs:
...
+
+  arm64:
+    needs: metadata
+    uses: ./.github/workflows/_build_base.yaml
+    with:
+      ARCHITECTURE: arm64
       BUILD_DATE: ${{ needs.metadata.outputs.BUILD_DATE }}
     secrets: inherit
 
   publish:
```
terrykong, line 54: Any reason this file doesn't use _publish_container.yaml?

.github/container/Dockerfile.base
terrykong: It's my opinion that we should remove the global username/email since this is kind of a silent failure since the end user doesn't get alerted that their commits are authored by jax@nvidia.com

.github/container/Dockerfile.pax.arm64
```diff
@@ -1,65 +1,152 @@
...
+# build lingvo
+RUN <<"EOT" bash -exu
+set -o pipefail
+
+pushd ${SRC_PATH_LINGVO}
+git fetch origin pull/329/head:pr329
+git cherry-pick --allow-empty pr329
+
+# Disable 2 flaky tests here
+patch -p1 <<"EOF"
```
terrykong, line 59: What do you think about using using a copy heredoc to add the patch to the container and then committing this so it doesn't look like we have unstaged files?

.github/workflows/_ci.yaml
```diff
@@ -0,0 +1,203 @@
...
+      SRC_PAXML:
+        type: string
+        required: true
+      SRC_PRAXIS:
+        type: string
+        required: true
+    outputs:
+      TAG_BASE:
+        description: "Tags of the base image built"
+        value: ${{ jobs.build-base.outputs.DOCKER_TAGS }}
```
terrykong, line 34: I think this should be:
```suggestion
        value: ${{ jobs.build-base.outputs.DOCKER_TAG }}
```

.github/container/pip-finalize.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -ex -o pipefail
+
+pushd /opt/pip-tools.d
+
+pip-compile -o requirements.txt $(ls manifest.*)
```
terrykong, line 7: This still yields unpinned VCS installs. I believe we still need a mechanism like this: https://github.com/NVIDIA/JAX-Toolbox/pull/271/files#diff-8386cfe5c29a3be212a911c9550a56c57420a17cae190b80f60e52c0ce265645R151-R175

.github/workflows/nightly-t5x-build.yaml
```diff
@@ -22,40 +22,86 @@ permissions:
...
     secrets: inherit
 
-  publish:
-    if: (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success') || (github.event_name == 'workflow_dispatch' && inputs.PUBLISH)
-    needs: [metadata, build]
+  arm64:
+    needs: metadata
+    runs-on: ubuntu-22.04
+    outputs:
+      DOCKER_TAG_FINAL: ''
```
terrykong, line 68: Should this be:
```suggestion
      DOCKER_TAG_MEALKIT: ''
      DOCKER_TAG_FINAL: ''
```
?

.github/workflows/nightly-te-test.yaml
```diff
@@ -3,7 +3,7 @@ run-name: Nightly Transformer Engine test (${{ github.event_name == 'workflow_ru
 
 on:
   workflow_run:
-    workflows: [Nightly Transformer Engine build]
+    workflows: [Nightly JAX build]
```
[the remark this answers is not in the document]
yhtang: Yes indeed. Good catch.

.github/workflows/ci.yaml
```diff
@@ -53,257 +53,35 @@ permissions:
...
     secrets: inherit
 
-  test-vit:
-    needs: build-rosetta-t5x
-    uses: ./.github/workflows/_test_vit.yaml
-    with:
-      ROSETTA_T5X_IMAGE: ${{ needs.build-rosetta-t5x.outputs.DOCKER_TAGS }}
-    secrets: inherit 
- 
   finalize:
```
terrykong, line 82: Are you planning on adding a "merge" step here to merge the arm64 and amd64 into a multiarch manifest? Or is the expectation for the concrete workflows `"Nightly ..."`  to provide a "ARCHITECTURE" workflow input?

I'm testing #354 and realizing that if a concrete workflow uses a multiarch image, it will fail using the image from the presubmit, since it's a single arch tag.

.github/workflows/nightly-t5x-build.yaml
```diff
@@ -22,40 +22,86 @@ permissions:
...
     steps:
+      - name: Check if the triggering workflow failed
+        id: if-upstream-failed
+        shell: bash -x -e {0}
+        run: |
+          echo "UPSTREAM_FAILED=${{ github.event.workflow_run.conclusion != 'success' }}" >> $GITHUB_OUTPUT
+
+      - name: Cancel workflow if upstream workflow did not success
+        if: ${{ steps.if-upstream-failed.outputs.UPSTREAM_FAILED == 'true' }}
+        uses: styfle/cancel-workflow-action@0.12.0
```
terrykong, line 38: I see there's new canceling logic. What does this do that the previous implementation didn't?

.github/workflows/_publish_container.yaml
yhtang: That won't work because:
1. `inputs.TARGET_IMAGE` is just the part before `:`)
2. `inputs.TARGET_IMAGE` may contain multiple lines of tags using the docker-metadata action format. For example, see https://github.com/NVIDIA/JAX-Toolbox/blob/5ae2e7913558ff134e17c592ea40f0aadc1f04d2/.github/workflows/nightly-jax-build.yaml#L66C12-L66C12

Also, the reason why I use the `manifests` step to pick up the SHAs is to get ride of an 'unknown' platform that gets added by buildx for every image. It is largely cosmetic but could still be confusing for end users.

.github/workflows/nightly-te-test.yaml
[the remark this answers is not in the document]
yhtang: Fixed in 6974a3ad.

.github/workflows/nightly-te-test.yaml
yhtang: Fixed in 6974a3ad

.github/workflows/weekly-base-build.yaml
yhtang: I think I might just forget to to so. Done in ce8ba2d.

.github/container/Dockerfile.pax.arm64
yhtang: This is for lingvo, which is a transitive dependency built in the `lingvo-builder` stage. The source code is not included in the later mealkit/final stages. Only the wheel is copied over.

.github/workflows/_ci.yaml
yhtang: Fixed in 6501ae905fad29b4869eba7c7a1f749b088cf616

.github/container/pip-finalize.sh
yhtang: Does this actually pin the versions or only document the SHA of the VCS installs at the time of pip-compile?

.github/workflows/ci.yaml
yhtang: Right now we are doing one job per multi-arch image (JAX/PAX/T5X) and it's too verbose. We could add a single job at the end of the presubmit to properly group all single-arch images. It can take advantage of the sitrep jsons to locate the single-arch images.

yhtang: Would it be OK to do it as a follow-up PR?

.github/workflows/nightly-t5x-build.yaml
yhtang: Potentially allowing for more complicated cancellation logic in the metadata job, plus nicer job graph visualization IMHO :)

yhtang: Previously the logic was limited to what a single-line GitHub Actions expression allows. Now we can do arbitrary bash scripting to determine whether a job needs to be cancelled.

.github/container/pip-finalize.sh
terrykong: IIRC this will pin it

.github/workflows/ci.yaml
terrykong: Sure. Let's not block this PR

.github/workflows/nightly-t5x-build.yaml
terrykong: Got it. That's helpful. IIRC, I think there were sometimes race conditions when `if-upstream-failed`'s outcome was "fail", since it wasn't listed in any job's `needs:` section, you could sometimes get farther along on jobs dependent on the `metadata` job before they fail.

.github/workflows/nightly-t5x-build.yaml
terrykong: This line was updated, but the change was FINAL -> MEALKIT. Should this not have both outputs?

.github/workflows/_ci.yaml
```diff
@@ -0,0 +1,205 @@
...
+    uses: ./.github/workflows/_test_pax.yaml
+    with:
+      PAX_IMAGE: ${{ needs.build-pax.outputs.DOCKER_TAG_FINAL }}
+    secrets: inherit
+
+  test-vit:
+    needs: build-rosetta-t5x
+    uses: ./.github/workflows/_test_vit.yaml
+    with:
+      ROSETTA_T5X_IMAGE: ${{ needs.build-rosetta-t5x.outputs.DOCKER_TAGS }}
```
ashors1, line 203: Should this be `needs.build-rosetta-t5x.outputs.DOCKER_TAG_FINAL`?

.github/workflows/nightly-rosetta-pax-build-test.yaml
```diff
@@ -31,69 +31,108 @@ permissions:
...
+        type=raw,value=latest,priority=1000
+        type=raw,value=nightly-${{ needs.metadata.outputs.BUILD_DATE }},priority=900  
 
   test-pax:
-    needs: build
+    needs: [metadata, amd64, arm64]
     uses: ./.github/workflows/_test_pax_rosetta.yaml
     if: (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success') || github.event_name == 'workflow_dispatch'
     with:
       PAX_IMAGE: ${{ needs.build.outputs.DOCKER_TAGS }}
```
ashors1, line 131: `needs.amd64.outputs.DOCKER_TAG_FINAL`?

.github/workflows/nightly-rosetta-t5x-build-test.yaml
```diff
@@ -31,79 +31,117 @@ permissions:
...
+        type=raw,value=latest,priority=1000
+        type=raw,value=nightly-${{ needs.metadata.outputs.BUILD_DATE }},priority=900  
 
   test-unit:
     if: (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success') || github.event_name == 'workflow_dispatch'
-    needs: build
+    needs: [metadata, amd64, arm64]
     uses: ./.github/workflows/_test_rosetta.yaml
     with:
       ROSETTA_IMAGE: ${{ needs.build.outputs.DOCKER_TAGS }}
```
ashors1, line 132: I think `ROSETTA_IMAGE` needs to be updated here

.github/workflows/nightly-rosetta-t5x-build-test.yaml
```diff
@@ -31,79 +31,117 @@ permissions:
...
     secrets: inherit
 
   test-t5x:
-    needs: build
+    needs: [metadata, amd64, arm64]
     uses: ./.github/workflows/_test_t5x_rosetta.yaml
     if: (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success') || github.event_name == 'workflow_dispatch'
     with:
       T5X_IMAGE: ${{ needs.build.outputs.DOCKER_TAGS }}
```
ashors1, line 140: Same as above

.github/workflows/_test_pax.yaml
```diff
@@ -20,137 +20,7 @@ on:
...
-              dump = {'state': "${{ steps.submit.outputs.SLURM_STATE }}", 'exitcode': "${{ steps.submit.outputs.SLURM_EXITCODE }}"}
-              json.dump(dump, f)
-          EOF
-        
-      - name: Upload training logs as artifacts
-        uses: actions/upload-artifact@v3
-        with:
-          name: ${{ steps.meta.outputs.JOB_NAME }}
-          path: output/*
-
```
ashors1, line 152: I think the singleprocess tests were erroneously removed here. Can we add them back if this is indeed a mistake?

.github/workflows/_test_pax_rosetta.yaml
```diff
@@ -20,138 +20,7 @@ on:
...
-              dump = {'state': "${{ steps.submit.outputs.SLURM_STATE }}", 'exitcode': "${{ steps.submit.outputs.SLURM_EXITCODE }}"}
-              json.dump(dump, f)
-          EOF
-        
-      - name: Upload training logs as artifacts
-        uses: actions/upload-artifact@v3
-        with:
-          name: ${{ steps.meta.outputs.JOB_NAME }}
-          path: output/*
-
```
ashors1, line 153: Same comment about the singleprocess tests

ashors1:
Note that the presubmit CI does not test the `nightly-*.yaml` files. Can we make sure to manually run `nightly-rosetta-{t5x,pax}-build-test.yaml` before merging as well to ensure there are no issues there?

yhtang:
Makes sense. I did test the nightly workflows manually before making the PR, but it's necessary to run them again as the PR has evolved.

.github/workflows/_ci.yaml
DwarKapex: Done

.github/workflows/nightly-rosetta-pax-build-test.yaml
DwarKapex: Agree. But question arises: do we want to run this tests on arm? If yes, it would be better to split this file into 2: build and test. I created a ticket for this: https://github.com/NVIDIA/JAX-Toolbox/issues/376

.github/workflows/nightly-rosetta-t5x-build-test.yaml
DwarKapex: Done

.github/workflows/_test_pax.yaml
DwarKapex: Reverted

.github/workflows/_test_pax_rosetta.yaml
DwarKapex: Done

terrykong:
FYI @yhtang  and @DwarKapex. Here is a PR rebased on this branch (before Vlad's changes) that added reproducible patches, but now incorporates the mealkit/final mechanic.

yhtang:
> FYI @yhtang and @DwarKapex. Here is a PR rebased on this branch (before Vlad's changes) that added reproducible patches, but now incorporates the mealkit/final mechanic.

Seems like the PR number is missing?

terrykong:
Apologies. Yes, @yhtang #354 is the one

.github/container/pip-finalize.sh
terrykong: After speaking with @DwarKapex , we will just address this in another PR since it's better to get this out ASAP. I will file an issue for this

terrykong:
Created issue to track the VCS installs: https://github.com/NVIDIA/JAX-Toolbox/issues/384 @DwarKapex  and @yhtang

.github/container/Dockerfile.base
yhtang: I see. So the container should not ship information that are user-specific. Makes sense.

.github/workflows/nightly-rosetta-pax-build.yaml
```diff
@@ -11,7 +11,7 @@ on:
       BASE_IMAGE:
         type: string
         description: 'PAX image built by NVIDIA/JAX-Toolbox'
-        default: 'ghcr.io/nvidia/upstream-pax:latest'
+        default: 'ghcr.io/nvidia/upstream-pax:mealkit'
```
terrykong, line 14: It looks like this default won't work anymore. Can the description and default be updated?

This image would result in:
```
BASE_IMAGE_AMD64=ghcr.io/nvidia/upstream-pax:mealkit-amd64-mealkit
BASE_IMAGE_ARM64=ghcr.io/nvidia/upstream-pax:mealkit-arm64-mealkit
```

My opinion is default = '',  and the description should be updated to specify it only accepts jax-toolbox internal images that get single arch tags

.github/workflows/nightly-rosetta-t5x-build-test.yaml
```diff
@@ -31,79 +31,120 @@ permissions:
 jobs:
```
terrykong, line 31: Same comment about inputs.BASE_IMAGE's default/description in rosetta pax's build-test workflow

rosetta/Dockerfile.pax
```diff
@@ -25,7 +30,7 @@ EOF
 COPY --from=rosetta-source / /opt/rosetta
 WORKDIR /opt/rosetta
 RUN --mount=target=/opt/pax-mirror,from=pax-mirror-source,readwrite \
-    --mount=target=/opt/praxis-mirror,from=praxis-mirror-source,readwrite <<"EOF" bash -e
```
terrykong, line 28: Any reason the `bash -e` was removed?

rosetta/Dockerfile.t5x
```diff
@@ -25,7 +25,7 @@ EOF
 COPY --from=rosetta-source / /opt/rosetta
 WORKDIR /opt/rosetta
 RUN --mount=target=/opt/t5x-mirror,from=t5x-mirror-source,readwrite \
-    --mount=target=/opt/flax-mirror,from=flax-mirror-source,readwrite <<"EOF" bash -e
```
terrykong, line 28: Same Q about the `bash -e` being removed

.github/workflows/nightly-rosetta-pax-build.yaml
DwarKapex: Yeap, agree. Updated the value

rosetta/Dockerfile.pax
DwarKapex: Reverted after PM

rosetta/Dockerfile.t5x
```diff
@@ -12,45 +15,47 @@ ADD --keep-git-dir=true https://github.com/google-research/t5x.git#main /
...
   -m https://github.com/nvjax-svc-0/t5x.git \
-  -d $(dirname $(python -c "import t5x; print(*t5x.__path__)")) \
+  -d /opt/t5x \
   -e /opt/t5x-mirror
 bash create-distribution.sh \
-  -p patchlist-flax.txt \
+  -p ${FLAX_PATCHLIST} \
   -m https://github.com/nvjax-svc-0/flax.git \
-  -d $(dirname $(python -c "import flax; print(*flax.__path__)")) \
+  -d /opt/t5x \
```
terrykong, line 45: This should be /opt/flax

rosetta/Dockerfile.pax
```diff
@@ -17,30 +20,43 @@ ADD --keep-git-dir=true https://github.com/google/paxml.git#main /
...
-EOF
+ARG PAXML_PATCHLIST
+ARG PRAXIS_PATCHLIST
 
 COPY --from=rosetta-source / /opt/rosetta
 WORKDIR /opt/rosetta
 RUN --mount=target=/opt/pax-mirror,from=pax-mirror-source,readwrite \
-    --mount=target=/opt/praxis-mirror,from=praxis-mirror-source,readwrite <<EOF
+    --mount=target=/opt/praxis-mirror,from=praxis-mirror-source,readwrite <<EOF 
+bash -e
```
terrykong, line 39: Was moving this from line 38 down intentional?

rosetta/Dockerfile.t5x
```diff
@@ -12,45 +15,47 @@ ADD --keep-git-dir=true https://github.com/google-research/t5x.git#main /
...
-EOF
+ARG T5X_PATCHLIST
+ARG FLAX_PATCHLIST
 
 COPY --from=rosetta-source / /opt/rosetta
 WORKDIR /opt/rosetta
 RUN --mount=target=/opt/t5x-mirror,from=t5x-mirror-source,readwrite \
-    --mount=target=/opt/flax-mirror,from=flax-mirror-source,readwrite <<EOF
+    --mount=target=/opt/flax-mirror,from=flax-mirror-source,readwrite <<EOF 
+bash -e
```
terrykong, line 34: Same Q about moving bash -e

DwarKapex (approved):

terrykong (approved):

PR #372, opened 2023-11-13
Update the native-fp8 guide with cudnn fmha

kaixih:
A recent commit in XLA has activated the cuDNN FMHA by default. Unfortunately, this change has led to a slowdown in fp8 performance. As a recommendation, we advise users to temporarily disable this flag until further optimizations are implemented.

@nouiz @ashors1 @terrykong

### DIFF
--- a/rosetta/docs/NATIVE_FP8.md
+++ b/rosetta/docs/NATIVE_FP8.md
@@ -113,6 +113,7 @@ In addition to the suggested XLA flags mentioned in [this section](https://githu
 export XLA_FLAGS=" \
     --xla_gpu_enable_reduction_epilogue_fusion=false \
     --xla_gpu_enable_triton_gemm=false \
+    --xla_gpu_enable_cudnn_fmha=false \
     --xla_gpu_enable_cudnn_layer_norm=true \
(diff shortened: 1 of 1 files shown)


terrykong:
Looks like it was rolled back: https://github.com/openxla/xla/commit/6591d85d78168fa93cbf50fb677f83b320b91333

Is this still needed?

kaixih:
It seems an internal test conducted by Google had triggered this rollback. I anticipate that re-enabling the flag in the future could lead to another slowdown unless our fp8 and their internal test share the same root cause of the slowdown. Therefore, to be conservative, I propose that we explicitly disable this flag until we have a clearer understanding of its impact. I am currently investigating the matter internally and will provide updates as they become available.

kaixih:
During the performance benchmark, I found the `SAVE_NOTHING` checkpoint policy is preferred in the fp8 training. So, updated the doc accordingly. @terrykong @nouiz

nouiz (approved):

terrykong (approved):

terrykong:
FFTM @kaixih

PR #373, opened 2023-11-13
Remove 3 jobs: upload_badge, summary, copy_gist.

DwarKapex:
The `_finalize.yaml` calls 3 reusabel workflows:

1. _upload_badge.yaml
2. _summary.yaml
3. _copy_gist.yaml

These 3 workflows are used only in `_finalyze.yaml`. One easy way to save a few workloads would be to move the contents of those 3 files directly into `_finalyze.yaml`.

Perhaps, they will be used in a future, so let's rename it by adding extra `_` at the beginning of the filename and add deprecation message.


yhtang (changes requested):
Let's hold on until #371 merges :)

ashors1 (approved):

terrykong (approved):

.github/workflows/_finalize.yaml
```diff
@@ -8,22 +8,121 @@ on:
         description: Update the landing page badges with run outcomes
         default: false
         required: false
+      # upload badge
```
yhtang, line 11: This comment is vague. Please extend it to make it more informative. Additionally, for the workflow inputs/outputs section, the `description` field is the perfect place for documentation.

.github/workflows/_finalize.yaml
```diff
@@ -8,22 +8,121 @@ on:
...
-  publish-badge:
-    needs: [upload-badge]
-    if: inputs.PUBLISH_BADGE == true
-    uses: ./.github/workflows/_copy_gist.yaml
-    with:
-      SRC_GIST_ID: ${{ needs.upload-badge.outputs.GIST_ID }}
-      DST_GIST_ID: ${{ vars.BADGE_ENDPOINT_GIST_ID }}
-      FILES: '.*badge.*.json'
-    secrets: inherit
+  action:
```
yhtang, line 24: Please use separate jobs for each of the reusable workflows that are being replaced.

.github/workflows/_finalize.yaml
```diff
@@ -8,22 +8,121 @@ on:
...
         default: false
         required: false
+      # upload badge
+      BADGE_FILES:
+        type: string
+        description: 'Name/bash regex for shields.io endpoint JSON files'
+        default: '*badge*.json'
+        required: false
+    outputs:
+      # upload badge
```
yhtang, line 18: Same comment as above.

yhtang (changes requested):

DwarKapex:
Close this PR and create a new one, bc it was partially addressed in `add-pip-compile` PR

Issue #376, opened 2023-11-16
Split nightly-rosetta-*-build-test.yaml into build and test files


DwarKapex:

terrykong:
Initially the workflows were separate, but they were merged in order to allow the publishing of the `:latest` tag to be contingent on the test results.

We can revisit when there's BW, but I think when splitting, the container publishing logic should be:
1. Run rosetta build workflow
    1. publish `:nightly-YYYY-MM-DD` tag if build successful
2. Run rosetta test workflow on `:nightly-YYYY-MM-DD` (if `workflow_run`), `:latest` (if `workflow_dispatch`)
    1. publish `:latest` if tests successful

As an aside, by re-splitting the workflows, we'll be at the limit for workflow run chaining ([3 levels max](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#:~:text=You%20can%27t%20use%20workflow_run%20to%20chain%20together%20more%20than%20three%20levels%20of%20workflows)). Not a blocker since I don't think we'll go beyond that in the near future.

PR #395, opened 2023-11-25
Adds a new synchronization workflow at the end of the nightly run

terrykong:
This PR introduces a new workflow that runs at the end of all nightly workflow leaves, that can be used to update a manifest file.

## The Idea
We want to have a single workflow run if all the test workflows pass, i.e.:
```
on:
  workflow_run:
    workflows:
      - Nightly JAX unit test
      - Nightly Transformer Engine test
      - Nightly Distribution test
      - Nightly Pax MGMN performance test
      - Nightly T5X MGMN performance test
      - Nightly Rosetta Paxml build and test
      - Nightly Rosetta T5x build and test
    types: [completed]
    branches: [main]
```

After all the tests finish, we will update the manifest. The PR to update the manifest will come after this change, so as of now this PR introduces `nightly-manifest-update.yaml`, but it doesn't update anything

## The problem
As of now, there is no good way to trigger a workflow if several dependent workflows succeed. There are several issues open about this matter and there does not seem to be a plan to support this anytime in the future. 

The issue with the above `on.workflow_run.workflows[]` syntax is a nightly-manifest-update.yaml run will be started for each entry in that list, regardless of if it fails or not. This is an issue because:

1. The workflow is started even if the parent workflow failed
2. There's no builtin way to inspect the test status of the other parent workflows, only the one that triggered the current workflow
3. With several of these workflows started, we can run into a race condition if we try to `git merge` from several independent workflows
4. There's not a good way to know which one of these workflows will be the "last" to finish, since that might be a function of the workload

## The solution 
The solution this PR implements is to let `N` `nightly-manifest-update.yaml` runs get created and cancel all but the last one that was started; `N` here is determined by `on.workflow_runs.workflows[] | length`. This works because each workflow run will be given a unique workflow run id, and we can just query all the workflow runs for the current nightly and kill all workflows that have ids less than the max id.

Roughly the algorithm in pseudocode is:
```python
# Each instance of nightly-manifest-update.yaml runs this
THIS_RUN_ID=12345
while True:
  ALL_MANIFEST_UPDATE_RUNS = (
    get_nightly_workflows() # all nightly workflows
      .filter(lambda w: w is 'nightly-manifest-update.yaml')  # just the nightly manifest ones
  )
  max_id = max(run.id for run in ALL_MANIFEST_UPDATE_RUNS)
  if len(ALL_MANIFEST_UPDATE_RUNS) == N and max_id == THIS_RUN_ID:
    # This is the workflow we want to do the update
    print_status()
    manifest_update()
  elif max_id != THIS_RUN_ID:
    cancel(THIS_RUN_ID)
  # If it's the max, but we haven't seen N runs, then sleep and loop
  sleep(sec=120)
```

## Example (`print_status()`):
The Nightly Manifest Update will also summarize the parent workflow's conclusions. Here's what it will look like
| run_id | parent_id | parent_conclusion | parent_workflow_name |
| --- | --- | --- | --- |
| 6997938809 | 6997937815 | failure | Nightly Jax Unit Test (nightly 2023-11-26T20:52:04Z) |
| 6997937819 | 6997936667 | success | Nightly T5x Test (nightly 2023-11-26T20:51:50Z) |
| 6997936668 | 6997935547 | success | Nightly Pax Test (workflow_dispatch) |

## Note:
* After the introduction of this workflow, we are now at the limit of the number of workflows that can be chained together by workflow_run ([3 workflow limit](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows))
    * Nightly Jax Build -> Nightly Pax Build -> Nightly Rosetta Pax build and test -> Nightly Manifest Update 
    * This means that a new mechanism may need to be introduced to tackle #376

### DIFF
--- a/.github/workflows/nightly-distribution-test.yaml
+++ b/.github/workflows/nightly-distribution-test.yaml
@@ -13,6 +13,15 @@ permissions:
   actions:  write # to cancel previous workflows
 
 jobs:
+  save_parent:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Save Parent workflow_id ${{ github.event.workflow_run.id }}
+        run: echo ${{ github.event.workflow_run.id }} | tee parent-run-id.txt
+      - uses: actions/upload-artifact@v3
+        with:
+          name: parent-run-id
+          path: parent-run-id.txt
 
   test-distribution:
     uses: ./.github/workflows/_test_distribution.yaml
--- a/.github/workflows/nightly-jax-build.yaml
+++ b/.github/workflows/nightly-jax-build.yaml
@@ -18,6 +18,15 @@ permissions:
   packages: write # to upload container
 
 jobs:
+  save_parent:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Save Parent workflow_id ${{ github.event.workflow_run.id }}
+        run: echo ${{ github.event.workflow_run.id }} | tee parent-run-id.txt
+      - uses: actions/upload-artifact@v3
+        with:
+          name: parent-run-id
+          path: parent-run-id.txt
 
   metadata:
     runs-on: ubuntu-22.04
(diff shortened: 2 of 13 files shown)


terrykong:
After discussion with @yhtang offline, we have decided to close this in favor of merging all of the workflows so that presubmit and nightly are run with the same monolithic workflow.

This change was only needed because all the workflows were kept separate.

PR #398, opened 2023-11-28
Increase TE test time

DwarKapex:
Because of limited number of runners, we encounter the error below:
```
slurmstepd: error: *** JOB 18886 ON compute-permanent-node-954 CANCELLED AT 2023-11-24T10:25:54 DUE TO TIME LIMIT ***
```

Try to increase a waiting time to reduce the number of test failures because of timeout

### DIFF
--- a/.github/workflows/_test_te.yaml
+++ b/.github/workflows/_test_te.yaml
@@ -115,7 +115,7 @@ jobs:
           #SBATCH --nodes=1
           #SBATCH --tasks=1
           #SBATCH --gpus-per-node=${{ matrix.N_GPU }}
-          #SBATCH --time=00:05:00
+          #SBATCH --time=00:15:00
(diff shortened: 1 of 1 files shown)


terrykong (approved):

terrykong:
Looks like TE tests passed 👍

yhtang:
Looks like this is the right prescription for the wrong diagnosis 😄 

The test failure is **NOT** a consequence of the limited number of available runners. The number of available runners determines how long a job may need to wait in the queue before it starts executing. If the SLURM job waits for too long, the corresponding GitHub Actions job may time out. In this particular case, what timed out was the SLURM job itself, and the error occurred because the job was only given 5 minutes to finish once it started yet needed more to run the tests.

@DwarKapex Could you please update the PR description so that it is more accurate from a documentation point of view?

PR #400, opened 2023-11-29
Remove 1 reusable workflow: copy_gist

DwarKapex:
The _finalize.yaml calls 1 reusabel workflows:
	1	_copy_gist.yaml

Theis 1 workflows are used only in _finalyze.yaml. One easy way to save a few workloads would be to move the contents of this file directly into `_finalyze.yaml`.
Perhaps, they will be used in a future, so let's rename it by adding extra _ at the beginning of the filename and add deprecation message.


ashors1 (approved):

yhtang (changes requested):
LGTM, however this is not tested by the presubmit CI. Could you please run a Sandbox workflow to verify if it works and can successfully publish a dummy badge?

PR #401, opened 2023-11-29
Migrate to sitrep mechanism for T5X and PAXML MGMN tests

### LINKED ISSUE #235: Add sitrep generation for the T5X MGMN tests workflow

### LINKED ISSUE #236: Add sitrep generation for the PAX MGMN tests workflow

### LINKED ISSUE #399: Improve badge reporting for upstream MGMN tests
Currently, the MGMN badges report "n/a" if any test fails:

image

This is because the `_test_{pax,t5x}.yaml` workflow has a "failure" status if any one of the tests fail. See https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/nightly-t5x-test-mgmn.yaml#L79 as an example

Instead of this, we should update the logic to summarize the number of tests that fail, regardless of `_test_*.yaml`'s result.

We are also sunsetting _publish_badge.yaml, so this is also a good opportunity to update these badges to the sitrep mechanism: https://github.com/NVIDIA/JAX-Toolbox/issues/229

hemildesai:
Addresses https://github.com/NVIDIA/JAX-Toolbox/issues/399, https://github.com/NVIDIA/JAX-Toolbox/issues/235 and https://github.com/NVIDIA/JAX-Toolbox/issues/236

Example badge: 

Completed run: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7039302468

## Changes
- Create a reusable job for publishing sitrep for MGMN workflows
- Integrate it in T5X MGMN
- Integrate it in PAX MGMN

## TODO
- [ ]  Change badge endpoint in README for T5X
- [ ]  Change badge endpoint in README for PAX

Closes #399 
Closes #235 
Closes #236


.github/workflows/_sitrep_mgmn.yaml
```diff
@@ -0,0 +1,116 @@
...
+
+      - name: Download all artifacts from the previous jobs
+        uses: actions/download-artifact@v3
+
+      - name: Write Test Status
+        shell: bash -x -e {0}
+        run: |
+          EXIT_STATUSES="${GITHUB_RUN_ID}-*/*-status.json"
+
+          cat <<EOF >>$GITHUB_STEP_SUMMARY
```
yhtang, line 38: Please use a custom variable to pass information between job steps. Our plan is to collectively create a workflow run summary using a 'finalization' job so that this summary is generated as a single piece of Markdown document. Otherwise, we cannot control the order in which the individual step summaries show up in the workflow summary and the result will be fragmented.

.github/workflows/_sitrep_mgmn.yaml
```diff
@@ -0,0 +1,116 @@
...
+          }
+          pytest_failed_tests=$(cnt_type failed)
+          pytest_passed_tests=$(cnt_type passed)
+          pytest_total_tests=$(all_outcomes | wc -l)
+
+          if ([[ $failed_tests -eq 0 ]] && [[ $total_tests -gt 0 ]] && \
+              [[ $pytest_failed_tests -eq 0 ]] && [[ $pytest_total_tests -gt 0 ]]); then
+            status=success
+            badge_color=brightgreen
+            badge_message="${passed_tests}/${total_tests} ran ${pytest_passed_tests}/${pytest_total_tests} pass loss+perf"
```
yhtang, line 77: It is a bit challenging and not quite self-explanatory as to what the differences are between the 'tests' and 'pytest tests', which I assume refer to the number of test files and tests cases, respectively. Is there a more intuitive name?

.github/workflows/_sitrep_mgmn.yaml
```diff
@@ -0,0 +1,116 @@
...
+            badge_color=red
+            badge_message="${passed_tests}/${total_tests} ran ${pytest_passed_tests}/${pytest_total_tests} pass loss+perf"
+          else
+            status=failure
+            badge_color=yellow
+            badge_message='error'
+          fi
+
+          badge_label='Upstream Tests'
+          summary="T5X MGMN Test: $badge_message"
```
yhtang, line 89: Should 'T5X' be parameterized by some input variable?

.github/workflows/_test_pax.yaml
```diff
@@ -13,10 +13,20 @@ on:
...
+      ARTIFACT_NAME:
+        type: string
+        description: 'Name of the artifact zip file'
+        required: false
+        default: 'artifact-pax-mgmn-test'
     outputs:
       TEST_STATUS:
         description: 'Summary pass/fail value indicating if results from tests are acceptable'
-        value: ${{ jobs.publish-test.outputs.STATUS }}
+        value: ${{ jobs.sitrep.outputs.STATUS }}
```
yhtang, line 29: This is similar to our current practice, which uses a final postprocessing job to determine whether the overall MGMN test suite succeeds.

It has the problem that error feedback is delayed and made it difficult to investigate which individual tests failed.

For the new sitrep reporting system, let's do the following:
- Individual jobs should fail immediately if the tests that it ran do not succeed. However, it should collect and generate test artifacts regardless of whether the tests pass or fail. [The `continue-on-error` option can be helpful here](https://github.com/NVIDIA/JAX-Toolbox/blob/1cec8a38ecc0ceb7f63b30ced4480910192fee7b/.github/workflows/_test_jax.yaml#L65).
- The sitrep/postprocessing step should then follow regardless of the success/fail status of individual test jobs.

This helps to locate the error status for easier debugging and tracking.

.github/workflows/_test_pax.yaml
```diff
@@ -518,12 +481,12 @@ jobs:
           ) | tee $GITHUB_STEP_SUMMARY
 
   outcome:
-    needs: publish-test
+    needs: sitrep
```
yhtang, line 484: This job would be unnecessary if we localize the error status to individual test jobs. However, there may still be situations where it is needed, i.e. when some checks are performed on the collective results of many/all test jobs.

.github/workflows/nightly-t5x-test-mgmn.yaml
```diff
@@ -66,45 +66,9 @@ jobs:
...
-          echo "MESSAGE='n/a'" >> $GITHUB_OUTPUT
-          echo "COLOR='red'" >> $GITHUB_OUTPUT
-        fi
-        echo "LABEL='Completion'" >> $GITHUB_OUTPUT
-        echo "STATUS='$STATUS'" >> $GITHUB_OUTPUT
-
   publish-verified:
-    if: needs.publish-completion.outputs.STATUS == 'success' && ((github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success') || (github.event_name == 'workflow_dispatch' && inputs.PUBLISH))
-    needs: [metadata, publish-completion]
+    if: needs.run-jobs.outputs.TEST_STATUS == 'success' && ((github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success') || (github.event_name == 'workflow_dispatch' && inputs.PUBLISH))
```
yhtang, line 70: Please move the logic to decide whether an image should be published to a separate `metadata` job. This would allow for better readability and the use of more complicated logics. For reference, see:
- https://github.com/NVIDIA/JAX-Toolbox/blob/1cec8a38ecc0ceb7f63b30ced4480910192fee7b/.github/workflows/nightly-jax-build.yaml#L59
- https://github.com/NVIDIA/JAX-Toolbox/blob/1cec8a38ecc0ceb7f63b30ced4480910192fee7b/.github/workflows/nightly-jax-build.yaml#L22

yhtang (changes requested):
Cool work! Please see some thoughts and suggestions below 😄

.github/workflows/_sitrep_mgmn.yaml
```diff
@@ -0,0 +1,116 @@
...
+    outputs:
+      STATUS:
+        description: 'Summary pass/fail value indicating if results from tests are acceptable'
+        value: ${{ jobs.sitrep.outputs.STATUS }}
+
+jobs:
+  sitrep:
+    runs-on: ubuntu-22.04
+    env:
+      BADGE_FILENAME_FULL: ${{ inputs.BADGE_FILENAME }}.json
```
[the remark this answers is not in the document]
hemildesai: No reason not to append it in the INPUT var directly, will shift to that and avoid the extra env var.

.github/workflows/_sitrep_mgmn.yaml
terrykong: How about:
```
"${passed_tests}/${total_tests} configs | ${pytest_passed_tests}/${pytest_total_tests} metrics"
```
?

.github/workflows/nightly-t5x-test-mgmn.yaml
terrykong: The logic to publish is dependent on the test result, so it cannot be moved to an ancestor of the test job

.github/workflows/_test_pax.yaml
hemildesai: For the MGMN usecase, the tests are submitted to slurm via SSH. To mark an individual job as failure I can inspect the exit code and mark it accordingly. We would still need the overall status though, if there's any other recommended way to get this output, I'm happy to incorporate that.

.github/workflows/_sitrep_mgmn.yaml
hemildesai: What do configs represent here? If it's just the jobs that ran to completion, can we use `jobs` instead of `configs` (might be easier to follow)?

.github/workflows/_sitrep_mgmn.yaml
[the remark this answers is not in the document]
hemildesai: Done.

.github/workflows/_sitrep_mgmn.yaml
hemildesai: Done.

.github/workflows/_sitrep_mgmn.yaml
hemildesai: Done.

.github/workflows/_sitrep_mgmn.yaml
yhtang: So is this the correct interpretation that a job may contain multiple tests?

.github/workflows/_test_pax.yaml
yhtang: Makes sense. How about we make individual jobs fail while also let the overall status be the Boolean AND of the individual job states?

.github/workflows/_test_pax.yaml
yhtang: Got it.

.github/workflows/_sitrep_mgmn.yaml
hemildesai: I think each job has multiple metrics to test, so yeah the tests should be a multiple of jobs I believe.

.github/workflows/_test_pax.yaml
hemildesai: Can do, but then we won't need `continue-on-error` I believe.

.github/workflows/_sitrep_mgmn.yaml
yhtang: Got it.

.github/workflows/_test_pax.yaml
yhtang: Also could you please add a natural language description of the logic regarding 
> The STATUS has to be derived not only from the job result, but also from its metrics
so that it serves as a form of documentation?

hemildesai: > At face value, the specific [error message](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7053172928/job/19199718748#step:8:18) from the job that you provided looks coincidental. I understand that it may be the result of the true underlying error, but the message itself does not clearly indicate that. Could we improve on it?

Not sure if there's a way to provide a more informative error based on the failure, might have to look at it outside the scope of this PR.

hemildesai: > Also could you please add a natural language description of the logic regarding
> 
> > The STATUS has to be derived not only from the job result, but also from its metrics
> 
> so that it serves as a form of documentation?

Yes will do it in the github actions yml.

.github/workflows/_sitrep_mgmn.yaml
hemildesai: Updated to 
` ${passed_tests}/${total_tests} jobs | ${pytest_passed_tests}/${pytest_total_tests} metrics`

.github/workflows/_test_pax.yaml
hemildesai: Done in https://github.com/NVIDIA/JAX-Toolbox/pull/401/commits/40aebf593c3e169b516f43ec3f429c327d2ec5e1

hemildesai:
@yhtang All feedback should be addressed now. Please take another look to see if any more changes are required.

.github/workflows/_sitrep_mgmn.yaml
```diff
@@ -0,0 +1,134 @@
...
+          EXIT_STATUS_SUMMARY_FILE="exit_status_summary.json"
+          echo -e "\n\n## ${{ inputs.FW_NAME }} MGMN+SPMD Test Status" >> $EXIT_STATUS_SUMMARY_FILE
+          cat <<EOF >>$EXIT_STATUS_SUMMARY_FILE
+          | Test Case | State | Exit Code |
+          | --- | --- | --- |
+          EOF
+
+          for i in $EXIT_STATUSES; do
+            # Files are named <FW_NAME>-<GHID>-<NAME>/<NAME>-status.json
+            echo "| $(echo $i | cut -d/ -f1 | cut -d- -f2) | $(jq -r .state $i) | $(jq -r .exitcode $i)"
```
ashors1, line 49: I think `cut -d- -f2` should be `cut -d- -f3` since we want to extract `<NAME>`.  Also, this will fail for `FW_NAME`s that include dashes, right? E.g. if we want to differentiate between the pax tests and rosetta pax tests, we cannot set `FW_NAME=rosetta-pax` for the rosetta tests. Perhaps it would be better to extract the last field rather than the third to handle such cases.

hemildesai: Makes sense, I will change to use last value

hemildesai: Fixed in https://github.com/NVIDIA/JAX-Toolbox/pull/401/commits/827a6ce885c536737a6881f3fddf480bb0ea4e27

hmonishN:
Listing the requirements for perf data dump here for tracking purpose @yhtang 

1. Job name to be dumped in artifacts (as part of sitrep.json file explaining the acronyms i.e TP, DP, FSDP, etc for pax and G, N for t5x, will be helpful if we add pax/t5x at the beginning of the job name)
2. _metric.json files to be dumped in the artifacts.

hemildesai:
@yhtang 

> I'm working on some light final touches on this PR.
> 
> A quick question regarding
> 
> ```
> echo "SHOULD_TRIAGE=${{ ((github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success') || github.event_name == 'workflow_dispatch') }} >> $GITHUB_OUTPUT"
> ```
> 
> If I use a different mechanism to cancel the workflow if its upstream worfklow did not succeed, do we still need the logic above at all? It appears that we always want to run triage as long as 1) upstream workflow succeeded or 2) the workflow is manually run via web GUI dispatch.
> 
> @hemildesai

@yhtang Yeah then it should be fine to remove this condition.

yhtang (dismissed):

yhtang:
Test nightly workflows:
https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7123351301/job/19395825884
https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7123352749/job/19395829302

hemildesai:
@yhtang Yes the changes look good. Thanks for the updates.

yhtang:
A latest job https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7127530028/job/19408352254 failed with the following error:
```
2023-12-07 12:39:40.366819: F external/xla/xla/stream_executor/cuda/cuda_dnn.cc:7393] Failed to generate cuDNN execution plan for opGraph Mul_Matmul_Add_Reduction_Sub_Exp_Reduction_Div_Matmul_. Status of final plan: CUDNN_STATUS_NOT_SUPPORTED
```
Although it is not caused by this PR, the fact that the job status is green seems to defeat [our expectation where individual job states should reflect the SLURM job exit states](https://github.com/NVIDIA/JAX-Toolbox/pull/401#discussion_r1412400540).

I'll investigate further.

hemildesai:
@yhtang This is because the status is retrieved after the job step is complete. Not sure if there's a way to retroactively set the job status. Either ways, I think the exit code of individual jobs is probably outside the scope of this PR. I'm happy to look into it and work on it in a separate PR (to prevent this PR from getting too verbose)

PR #402, opened 2023-11-30
Revert 4e4e137

ashors1:
This reverts the XLA WAR that was added to prevent NCCL OOM errors. This issue was addressed in XLA commit [0e751bf](https://github.com/openxla/xla/commit/0e751bf4ca01a081f6276d3353649aae441259b7), so the WAR is no longer needed.

PR #404, opened 2023-11-30
Consider only t5x tests when determining t5x test status

ashors1:
When running the presubmit CI, t5x's `publish-test` step incorrectly considered the outcomes of all tests rather than just the t5x ones. This PR filters out non-t5x tests to fix to fix this issue.


terrykong:
How about we do what you did for VIT and prefix the t5 tests with `T5` to allow us to easily filter for it?

yhtang (approved):
LGTM

yhtang:
Is this in a cyclic dependency situation with https://github.com/NVIDIA/JAX-Toolbox/pull/402 in terms passing the presubmit?

.github/workflows/_test_t5x.yaml
```diff
@@ -326,8 +326,8 @@ jobs:
         | --- | --- | --- |
         EOF
         for i in $EXIT_STATUSES; do
-          # Files are named <GHID>-<NAME>/<NAME>-status.json
-          echo "| $(echo $i | cut -d/ -f1 | cut -d- -f2) | $(jq -r .state $i) | $(jq -r .exitcode $i)"
+          # Files are named t5x-<GHID>-<NAME>/<NAME>-status.json
```
terrykong, line 329: Nit: Should this say `T5X`?

ashors1: Yep, fixed. Thanks!

terrykong (approved):

PR #405, opened 2023-11-30
Introduces manifest.yaml that is the "last working world state"

terrykong:
This introduces "the manifest" (`manifest.yaml`) that describes the complete state of the jax stack to allow for reproducible nightly builds and reproducible presubmit CI. The manifest (and patches) are staged in a "trial branch" each night, and if the build succeeds, we can merge the trial branch into main (TODO: GH Issue tracking automating merge). The presubmit CI runs on the PR's git-ref and so unless the author has committed a custom patch/manifest, they will always be running from the "last working state".

# Description of `manifest.yaml`
The manifest allows specifying/pinning libraries that serve different purposes:
* git repos: libraries that need their source checked out at a particular SHA
* pip VCS constraints: [VCS dependency specs](https://pip.pypa.io/en/stable/topics/vcs-support/) that need to be pinned like `fiddle @ git+https://github.com/google/fiddle`
* pip constraints: Python packages we need to constrain or pin to address one-off fixes, e.g., addressing a CVE

Here are example entries of each of these in the manifest.yaml

# Git repos
```yaml
jax:
  url: https://github.com/google/jax.git
  tracking_ref: main
  latest_verified_commit: b032a0271e3e2ea8d0df64d2f3f1a1e450a38dc9  # 2023-11-15
  mode: git-clone
```
* url: This is the url to lookup the latest git ref
* tracking_ref: This is the git-ref to lookup the latest commit SHA. Usually it’s main, but in some cases like lingvo, it’s master
* latest_verified_commit: SHA
* mode:
    * git-clone: Whether to clone the repo or not

## Git repos with patches
```yaml
flax:
  url: https://github.com/google/flax.git
  mirror_url: https://github.com/nvjax-svc-0/flax.git
  extra_dir: null
  tracking_ref: main
  latest_verified_commit: a572f6af2fef565c0f9ba2fc12b781e9e3385140
  mode: git-clone
  patches:
    pull/3340/head: file://patches/flax/PR-3340.patch # Add Sharding Annotations to Flax Modules
```
* mirror_url: a git url where we can pull patches that start with `^mirror/`
* extra_dir: an optional local dir to pull patches not found in upstream or the mirror. Useful for private repos
* patches: a dictionary where keys are git-refs and values are URIs for the local patch. The reason for the value is the git-ref can be rebased or be updated and the SHA will change; this makes patches a moving target when using git refs. So each `file://` URI will be a patch committed into Jax-Toolbox's VC to ensure reproducibility

# VCS Constraint
```yaml
clu:
  url: https://github.com/google/CommonLoopUtils.git 
  tracking_ref: main
  latest_verified_commit: 89c2face3474a7482358068d7a00d9bb6e4b31fe
  mode: pip-constraint
```
These aren't cloned (in fact `get-source.sh` will error if you try to clone). These are used in `pip-finalize.sh` to pin VCS dependencies like
```sh
clu @ git+https://github.com/google/CommonLoopUtils#egg=clu

# To

clu @ git+https://github.com/google/CommonLoopUtils@89c2face3474a7482358068d7a00d9bb6e4b31fe
```

# Pip constraint
```yaml
pydantic:
  mode: pip-constraint
  version: X.Y.Z
```
* mode: pip-constraint
    * Adds a pip constraint during pip-compile 
* version: If ${package}.version is present, then we can treat this as a pip constraint. This allows us to hotfix any python package in case we discover a bug or CVE.

# Changes to the CI
* `Nightly JAX Build` bumps the manifest.yaml and patches and commits them to a trial branch. If the tests pass, this trial branch should be merged to `main`
    * introduces `bump.sh` that bumps the world state given the manifest
* No more `REPO_*` and `REF_*` build args, since they are all specified in the `manifest.yaml`
* `get-source.sh` and `create-distribution.sh` now take the manifest
* Rename pip-tools "manifests" to "requirements-*.in` to avoid confusion and use the same terminology as pip-tools

# Not addressed in this PR
* [ ] Automating the merging of the trial branch at the end of a successful nightly run
* [ ] Running presubmit on latest world state (currently only will run on last-known working state)
* [ ] Move lingvo patches to the manifest (#409 )
* [ ] Move xla arm neon patch to the manifest
* [ ] Add a cleanup.sh script to remove things like `~/.gitconfig`
* [ ] Have not implemented `mode: pip-constraint` since there was no dep that needed a constraint


yhtang:
The history looks a bit weird. Maybe a rebase is needed? 😀

yhtang:
Some nitpicks mostly around naming:
- Would `ref` be better named `latest_verified_commit`  or something along the line?
- `mirror_url` -> `mirrors`: turn it into a list instead of a single string?
- `extra_dir` -> `patch_dir` to be more explicit?

.github/container/Dockerfile.base
```diff
@@ -17,13 +18,30 @@ RUN apt-get update && \
         git \
         lld \
         vim \
+        bat \
```
DwarKapex, line 21: Why do I see it? It is already in main, right?

.github/container/create-distribution.sh
```diff
@@ -78,18 +95,36 @@ if [[ $# -ge 1 ]]; then
...
-MIRROR_GIT_URL=${MIRROR_GIT_URL:-https://github.com/nvjax-svc-0/t5x.git}
-
-if [[ -z "${INSTALLED_DIR}" ]]; then
-  echo "[ERROR]: Need to specify -d/--dir"
+if [[ -z "$MANIFEST" || -z "$PACKAGE" ]]; then
+  echo "--manifest and --package must be provided"
   usage 1
 fi
 
+BASE_DIR=${BASE_DIR:-/opt}
```
DwarKapex, line 107: I don't see where you set BASE_DIR before this line. Looks like BASE_DIR is always `/opt`. Or `BASE_DIR` is env variable?

.github/container/Dockerfile.jax
```diff
@@ -47,48 +54,63 @@ RUN build-jax.sh \
...
 # The following environment variables tune performance
 ENV XLA_FLAGS="--xla_gpu_enable_latency_hiding_scheduler=true --xla_gpu_enable_async_all_gather=true --xla_gpu_enable_async_reduce_scatter=true --xla_gpu_enable_triton_gemm=false"
 ENV CUDA_DEVICE_MAX_CONNECTIONS=1
 ENV NCCL_IB_SL=1
 ENV CUDA_MODULE_LOADING=EAGER
+# Set the manifest env vars
+ENV MANIFEST_FILE=${DEST_MANIFEST_DIR}/${SRC_MANIFEST_FILE}
+
+ADD --chmod=777 create-distribution.sh ${DEST_MANIFEST_DIR}/
+COPY ${SRC_MANIFEST_FILE} ${DEST_MANIFEST_DIR}/${SRC_MANIFEST_FILE}
```
DwarKapex, line 80: Can you use env variable `MANIFEST_FILE` in dockerfile commands?

.github/container/Dockerfile.base
```diff
@@ -17,13 +18,40 @@ RUN apt-get update && \
...
+ADD --chmod=777 \
+    get-source.sh \
+    pip-finalize.sh \
+    /usr/local/bin/
+RUN wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_$(dpkg --print-architecture) -O /usr/local/bin/yq && \
+    chmod 777 /usr/local/bin/yq
+RUN git clone -b 23.3.1 https://github.com/pypa/pip.git /opt/pip
+# Patch is specific to 23.3.1
+# Generated via: "git diff > pip-vcs-equivalency.patch"
+ADD pip-vcs-equivalency.patch /opt/pip/
```
DwarKapex, line 47: Did you create an issue for JAX-Toolbox as a reminder to remove this hack when and if the change will be merged to pip?

.github/container/create-distribution.sh
terrykong: Good catch. It's not used anymore. Will remove

.github/container/create-distribution.sh
```diff
@@ -78,18 +95,36 @@ if [[ $# -ge 1 ]]; then
...
+UPSTREAM_URL=$(yq e ".${PACKAGE}.url" $MANIFEST)
+# The tracking_ref is interpreted as the default "main" branch and all patches are 
+# assumed to be rooted on a sha on the tracking_ref's history
+TRACKING_REF=$(yq e ".${PACKAGE}.tracking_ref" $MANIFEST)
+INSTALLED_DIR=${OVERRIDE_INSTALL_DIR:-/opt/${PACKAGE}}
+MIRROR_GIT_URL=$(yq e ".${PACKAGE}.mirror_url // \"\"" $MANIFEST)
+EXTRA_DIR=$(yq e ".${PACKAGE}.extra_dir // \"\"" $MANIFEST)
+
+SKIP_APPLY=${SKIP_APPLY:-0}
+GEN_PATCH_DIR=${GEN_PATCH_DIR:-$SCRIPT_DIR/patches/$PACKAGE}
```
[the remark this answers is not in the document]
terrykong: I originally left that in to give me a way to test this script without adding a cmdline arg since this script edits things in place. I no longer need that since I just copy the script to test it (see `rosetta/tests/*`), so I can remove the option to override it

.github/container/Dockerfile.jax
terrykong: I separated the `MANIFEST_FILE` into the filename itself and the dir, b/c I wanted to be able to use DEST_MANIFEST_DIR at build time for `ADD`/`COPY`, and re-use `SRC_MANIFEST_FILE` at build time to specify what "manifest file" to pull from the context, but also what the file should be named in `DEST_MANIFEST_DIR`

Do you see any other way to accomplish this? I implemented it this way in case down the road we have one manifest for arm and one for amd.

.github/container/Dockerfile.base
terrykong: I created an internal nvbug for this

.github/container/Dockerfile.jax
terrykong: Also, the `DEST_MANIFEST_DIR` is also used in the copy of the patches, and I can't `dirname` the `MANIFEST_FILE` since I need the dirname to be known at build time

.github/container/create-distribution.sh
```diff
@@ -5,35 +5,52 @@
...
+  -s, --skip-apply          [Optional] If provided, will only create patches, update manifest, and skip applying. When not provided, applies local patches.
 
-Relationship between --dir, --extra-dir, and --mirror-url repo args:
-  --dir: The upstream repo, locally cloned
-  --mirror-url: A mirror of the upstream repo
-  --extra-dir: A locally cloned mirror of the upstream repo. Helpful to incorporate changes from private repos.
+--------------
 
-Patches in the --patchlist will be applied from the repos above according to the following rules:
+This script has two modes of operation:
```
DwarKapex, line 17: Could you please add a general description of the purpose of thus script? Like, ""this script is for updating manifest with patches and apply patches to the closed repos".

.github/container/get-source.sh
```diff
@@ -0,0 +1,99 @@
...
+
+echo "Fetching $GIT_REPO#$GIT_REF to $INSTALL_DIR"
+
+set -ex -o pipefail
+
+git clone ${GIT_REPO} ${INSTALL_DIR}
+pushd ${INSTALL_DIR}
+git checkout ${GIT_REF}
+git submodule init
+git submodule update --recursive
```
DwarKapex, line 95: I think we can merge line 94 and 95 into one and use this git command:
```
git submodule update --init --recursive
```

.github/container/Dockerfile.pax.amd64
```diff
@@ -1,33 +1,46 @@
...
-ARG BASE_IMAGE=ghcr.io/nvidia/jax:latest
-FROM ${BASE_IMAGE}
+FROM ${BASE_IMAGE} as mealkit
+ARG SRC_PATH_PAXML
+ARG SRC_PATH_PRAXIS
 
-ADD install-pax.sh /usr/local/bin
-ADD install-flax.sh /usr/local/bin
-ADD install-te.sh /usr/local/bin
+# update TE manifest file to install the [test] extras
```
DwarKapex, line 15: Why here but not in Dockerfile.jax? TE is a part of JAX image now, so I would assume someone wants to test TE in JAX container.

.github/container/Dockerfile.pax.arm64
```diff
@@ -1,65 +1,143 @@
...
+RUN get-source.sh -l lingvo -m ${MANIFEST_FILE}
+
+# build lingvo
+RUN <<"EOT" bash -exu -o pipefail
+
+pushd ${SRC_PATH_LINGVO}
+git fetch origin pull/329/head:pr329
+git cherry-pick --allow-empty pr329
+
+# Disable 2 flaky tests here
```
DwarKapex, line 52: Why is it here but not a part of manifest file?

.github/container/bump.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
+
+## Parse command-line arguments
+
+usage() {
+cat <<EOF
+Usage: $0 [OPTION]...
+  -i, --input-manifest  PATH     If set, will clean the patch dir. Default is not to clean
+  -h, --help                     [Optional] Print usage.
```
DwarKapex, line 9: Don't think we need `[Optional]` here and below

.github/container/bump.sh
```diff
@@ -0,0 +1,97 @@
...
+  -o, --output-manifest PATH     [Optional] Use this if you don't want to update manifest in-place
+      --only-bump-patches        [Optional] If provided, update patch files and the patchlist in the manifest, but skip bumping refs
```
DwarKapex, line 11: '-b' maybe?

terrykong:
> Try to rebase/merge to main again (or create another clean PR), because it is pretty difficult to get an idea what changes are yours vs what changes came from `add-pip-compile` PR

SG. I've rebased (preview: #406 ), but it was a pretty gnarly rebase. Without reparenting, just squashing had conflicts. I need to test that I haven't broken anything, but once I've tested, I'll force push here

.github/container/bump.sh
```diff
@@ -0,0 +1,97 @@
...
+    echo "Un-recognized argument: $*"
+    echo
+    usage 1
+fi
+
+set -eou pipefail
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+MANIFEST_IN=${MANIFEST_IN:-}
```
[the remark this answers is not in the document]
terrykong: I don't have default values, but I `set -u` so the script would error if they weren't set. I folded it in to make it more clear

.github/container/bump.sh
terrykong: Done

.github/container/Dockerfile.pax.arm64
terrykong: I did not add the patch for lingvo, but I'm all for moving the patch into the manifest. Would it be okay to address this in a follow up PR? If so, I can file an issue

.github/container/Dockerfile.pax.amd64
terrykong: I don't mind. Any reason you kept it separate @yhtang ?

.github/container/get-source.sh
terrykong: done

yhtang:
Migrating comments from #406:
https://github.com/NVIDIA/JAX-Toolbox/pull/406#discussion_r1413408975

terrykong:
> given that the baking (pip-compile + pip-sync) step will be independent between the upstream and rosetta containers, it is technically possible to let the upstream JAX/T5X/PAX/etc packages use free-floating VCS installs and rosetta containers use pinned VCS installs. Whether it will bring in any advantage is another question, though.

Both rosetta and upstream containers will use free-floating VCS installs in the nightlies since `bump.sh` will always be run. It's just that if you try to rebuild the nightly from an existing mealkit, you'll get pinned unless you call `bump.sh`, or edit the manifest, yourself.

terrykong:
Re:
> extra_dir -> patch_dir to be more explicit?

@yhtang I've thought about this a little more and I don't like my suggestion of changing it to `mirror_dir` since currently I rely on `mirror_url` mapping to the `mirror` remote and `extra_dir` mapping to the `extra` remote, so if they both start with `mirror_`, it may be confusing. I documented the different sources of git-refs here: https://github.com/NVIDIA/JAX-Toolbox/pull/405/files#diff-bd111b2b880c006c97875516974cf4373b7fc486778ecb6f3b47e0b322911c53R52. Is this clear or if not, how would you suggest naming it?

.github/container/Dockerfile.pax.arm64
terrykong: TODO: Create issue

.github/container/Dockerfile.jax
terrykong: @DwarKapex and I synced offline on this, because of the limitation of dockerfile's not allowing bash commands at build time, the tradeoff here is between:

## Current
```dockerfile
ARG SRC_MANIFEST_FILE=manifest.yaml
ARG DEST_MANIFEST_DIR=/opt/manifest.d
ENV MANIFEST_FILE=${DEST_MANIFEST_DIR}/${SRC_MANIFEST_FILE}
ADD --chmod=777 create-distribution.sh ${DEST_MANIFEST_DIR}/
COPY ${SRC_MANIFEST_FILE} ${DEST_MANIFEST_DIR}/${SRC_MANIFEST_FILE}
COPY patches/ ${DEST_MANIFEST_DIR}/patches/
```
vs

## Proposed
```dockerfile
ARG MANIFEST_FILE=/opt/manifest.d/manifest.yaml
ARG DEST_MANIFEST_DIR=/opt/manifest.d
ENV MANIFEST_FILE=${MANIFEST_FILE}
ADD --chmod=777 create-distribution.sh ${DEST_MANIFEST_DIR}/
COPY ${SRC_MANIFEST_FILE} ${MANIFEST_FILE}
COPY patches/ ${DEST_MANIFEST_DIR}/patches/
```

Both have the issue that you must repeat something and there's possible for error. In the `Current` state, a developer changing `${DEST_MANIFEST_DIR}/${SRC_MANIFEST_FILE}` needs to do so in two places, which could lead to errors and in `Proposed` we need to make sure the `--build-args` both have the same absolute path prefix, otherwise things won't work. We have decided to stick with `Current` for now since a developer change is less frequent than a build arg change. We can revisit later if there's a more elegant way to accomplish this

### Aside
You can also use a separate build context and mount it into a `RUN` command, but we decided it was too much complexity and to just stick with either of these.

.github/container/Dockerfile.jax
```diff
@@ -1,30 +1,36 @@
...
 ARG SRC_PATH_JAX
 ARG SRC_PATH_XLA
 ARG BAZEL_CACHE
 
-RUN git clone "${REPO_JAX}" "${SRC_PATH_JAX}" && cd "${SRC_PATH_JAX}" && git checkout ${REF_JAX}
+# Set the manifest env vars
+ENV MANIFEST_FILE=${DEST_MANIFEST_DIR}/${SRC_MANIFEST_FILE}
+
+ADD --chmod=777 create-distribution.sh ${DEST_MANIFEST_DIR}/
+COPY ${SRC_MANIFEST_FILE} ${DEST_MANIFEST_DIR}/${SRC_MANIFEST_FILE}
```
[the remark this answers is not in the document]
terrykong: Resolved in another thread about the same build args

.github/container/Dockerfile.pax.arm64
terrykong: https://github.com/NVIDIA/JAX-Toolbox/issues/409

.github/container/manifest.yaml
```diff
@@ -0,0 +1,103 @@
...
+  mode: pip-vcs
+dllogger:
+  url: https://github.com/NVIDIA/dllogger.git
+  tracking_ref: master
+  latest_verified_commit: 0540a43971f4a8a16693a9de9de73c1072020769
+  mode: pip-vcs
+jestimator:
+  url: https://github.com/google-research/jestimator.git
+  tracking_ref: main
+  latest_verified_commit: "fa143d93e337ca8ab77c4510baf21ae52af24ab2"
```
DwarKapex, line 92: [tiny] No quotes are needed here =)

.github/container/Dockerfile.base
```diff
@@ -38,12 +38,24 @@ RUN <<"EOF" bash -ex
 git config --global user.name "${GIT_USER_NAME}"
 git config --global user.email "${GIT_USER_EMAIL}"
 EOF
-RUN pip install --upgrade --no-cache-dir pip pip-tools && rm -rf ~/.cache/*
 RUN mkdir -p /opt/pip-tools.d
 ADD --chmod=777 \
     get-source.sh \
     pip-finalize.sh \
     /usr/local/bin/
+RUN wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_$(dpkg --print-architecture) -O /usr/local/bin/yq && \
```
DwarKapex, line 46: [question] Will it be better to install it thru `apt install`:
```
$> add-apt-repository ppa:rmescandon/yq
$> apt update
$> apt install yq -y
```
?

.github/container/Dockerfile.pax.arm64
```diff
@@ -90,30 +84,27 @@ EOT
...
-echo "chex==0.1.7" >> /opt/pip-tools.d/manifest.pax
-echo "auditwheel" >> /opt/pip-tools.d/manifest.pax
-
-get-source.sh -f ${REPO_PAXML}  -r ${REF_PAXML}  -d ${SRC_PATH_PAXML}
-get-source.sh -f ${REPO_PRAXIS} -r ${REF_PRAXIS} -d ${SRC_PATH_PRAXIS}
-echo "-e file://${SRC_PATH_PAXML}[gpu]" >> /opt/pip-tools.d/manifest.pax
-echo "-e file://${SRC_PATH_PRAXIS}"     >> /opt/pip-tools.d/manifest.pax
+echo "tensorflow==2.13.0" >> /opt/pip-tools.d/requirements-paxml.in
+echo "tensorflow_datasets==4.9.2" >> /opt/pip-tools.d/requirements-paxml.in
+echo "chex==0.1.7" >> /opt/pip-tools.d/requirements-paxml.in
```
DwarKapex, line 101: Why do I see it here, but not on line 27?
I wonder, can we do something like this:

```
26: ENV EXTRA_PACKAGES="tensorflow==2.13.0 tensorflow_datasets==4.9.2 chex==0.1.7 auditwheel"
27: RUN <<"EOT" bash -exu -o pipefail
pip install ${EXTRA_PACKAGES}
get-source.sh -l tensorflow-text -m ${MANIFEST_FILE}
cd ${SRC_PATH_TFTEXT}
./oss_scripts/run_build.sh
EOT
<blah-blah-blah>
98:  RUN <<EOT bash -ex
99:  for spec_pkg in $(tr " " "\n"); do
100:    echo ${spec_pkg} >> /opt/pip-tools.d/requirements-paxml.in
101: done
```
?

.github/container/bump.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+
+## Parse command-line arguments
+
+usage() {
+cat <<EOF
+Usage: $0 [OPTION]...
```
DwarKapex, line 7: [minor] It would be nice to have a brief description what the purpose of the script is and what it does.

.github/container/bump.sh
```diff
@@ -0,0 +1,95 @@
...
+  -b, --only-bump-patches        If provided, update patch files and the patchlist in the manifest, but skip bumping refs
```
DwarKapex, line 8: Maybe, `-s, --skip-bump-refs`?

.github/container/bump.sh
```diff
@@ -0,0 +1,95 @@
...
+  -i, --input-manifest  PATH     If set, will clean the patch dir. Default is not to clean
```
DwarKapex, line 9: A bit lost. I think, the description is wrong. `--input-manifest` is a path to manifest file to be used to update the refs. What does it have in common with cleaning the patch dir?

.github/container/bump.sh
```diff
@@ -0,0 +1,95 @@
...
+  -h, --help                     Print usage.
+  -o, --output-manifest PATH     Use this if you don't want to update manifest in-place
```
DwarKapex, line 11: Update the description to something like: `Path to output manifest. Use this if you don't want to update manifest in-place`

.github/container/create-distribution.sh
```diff
@@ -0,0 +1,372 @@
...
+or fixes from the patches. This script does not build or install the library, but creates a local branch
+that includes all of the patches.
+
+Usage: $0 [OPTION]...
+  -c, --clean               [Optional] If set, will clean the patch dir. Default is not to clean
+  -h, --help                [Optional] Print usage.
+  -m, --manifest=PATH       Path to the manifest. Updates it in-place
+  -o, --override_dir=PATH   [Optional] Use this if there is a custom location of the upstream clone. If not specified, uses /opt/\${PACKAGE}
+  -p, --package=KEY         The package name in the manifest to use, e.g., t5x, paxml
+  -s, --skip-apply          [Optional] If provided, will only create patches, update manifest, and skip applying. When not provided, applies local patches.
```
DwarKapex, line 18: Remove [Optional] here an above

.github/container/get-source.sh
```diff
@@ -1,55 +1,54 @@
...
 ## Example:
-## get-source.sh -f https://github.com/google/flax.git -r main -d /opt/flax
+## get-source.sh -m manifest.yaml -l flax
 ## Output:
 ## -e /opt/flax
 
 ## Parse command-line arguments
 
 usage() {
     echo "Usage: $0 [OPTION]..."
```
DwarKapex, line 11: [minor] I think it is better to refactor it in the following manner:
```
usage() {
cat<<OEF
   <Description of the script: what it does, what is the purpose, another joke>
   Usage: $0 [OPTION]
...
```

.github/workflows/_ci.yaml
```diff
@@ -7,25 +7,7 @@ on:
...
-        type: string
-        required: true
-      SRC_T5X:
-        type: string
-        required: true
-      SRC_PAXML:
-        type: string
-        required: true
-      SRC_PRAXIS:
+      TRIAL_BRANCH:
```
DwarKapex, line 10: Add description?
```
      TRIAL_BRANCH:
        type: string
        description: 'Name of branch with bumped manifest and patches'
        required: true
        default: 'main'
```

.github/workflows/_sandbox.yaml
```diff
@@ -2,10 +2,27 @@ name: "~Sandbox"
 
 on:
   workflow_dispatch:
+  workflow_run:
```
DwarKapex, line 5: Don't forget to reverse it, when done with this PR

rosetta/tests/upstream-only-distribution.sh
```diff
@@ -6,20 +6,37 @@ cd $SCRIPT_DIR
...
+t5x:
+  url: https://github.com/google-research/t5x.git
+  mirror_url: https://github.com/nvjax-svc-0/t5x.git
+  tracking_ref: main
+  latest_verified_commit: $DISTRIBUTION_BASE_REF
+  mode: git-clone
+  patches:
+    pull/1372/head: null
+EOF
+bash ../../.github/container/get-source.sh --base-dir $tmp_base --manifest $manifest_tmp --library $LIBRARY
```
DwarKapex, line 28: I believe, relative paths are not reliable. You can use `GITHUB_WORKSPACE` here:
```
bash ${GITHUB_WORKSPACE}/.github/container/get-source.sh <...>
```

rosetta/tests/upstream-only-distribution.sh
```diff
@@ -6,20 +6,37 @@ cd $SCRIPT_DIR
...
+
+cp ../../.github/container/create-distribution.sh $workspace_tmp/
```
DwarKapex, line 30: Same here

rosetta/tests/mirror-only-distribution.sh
```diff
@@ -6,23 +6,38 @@ cd $SCRIPT_DIR
 set -eou pipefail
 
 # Version should work on Linux || Darwin
-#repo_tmp=$(mktemp -d 2>/dev/null || mktemp -d -t 'mytmpdir')
-repo_tmp=/tmp/t5x
-patchlist_tmp=$(mktemp /tmp/patchlist.txt.XXXXXX)
+tmp_base=$(mktemp -d 2>/dev/null || mktemp -d -t 'upstream')
```
DwarKapex, line 9: I do see code duplication: `mirror-only-discribution.sh` is identical to `upstream-only-distribution.sh` (except maybe couple lines of code). Do you plan to merge them?

rosetta/Dockerfile.t5x
```diff
@@ -2,52 +2,76 @@
...
+cd -
+############DELETE
+#
+MANIFEST_DIR=$(dirname ${MANIFEST_FILE})
+if [[ "${UPDATE_PATCHES}" != "true" && "${UPDATE_PATCHES}" != "false" ]]; then
+  echo "UPDATE_PATCHES can only be true or false"
+  exit 1
+fi
+if [[ "${UPDATE_PATCHES}" == "true" ]]; then
+  cp -r /mnt/jax-toolbox/.github/container/patches ${MANIFEST_DIR}/patches
```
DwarKapex, line 48: 1. Why do you copy all the stuff to another place?
2. If UPDATE_PATCHES == false, this docker build fails on line 60.

.github/container/manifest.yaml
terrykong: It's weird. `yq` will actually keep the quotes when you set it to "empty" as a "starting value". I'll remove

rosetta/Dockerfile.pax
```diff
@@ -2,55 +2,73 @@
...
+  # TODO: remove
+  cp /mnt/jax-toolbox/.github/container/pip-finalize.sh /usr/local/bin/
+fi
+cp -r /mnt/jax-toolbox/rosetta /opt/rosetta
+
+# Setting the username/email is required to author commits from patches
+git config --global user.email "${GIT_USER_EMAIL}"
+git config --global user.name "${GIT_USER_NAME}"
+
+bash ${MANIFEST_DIR}/create-distribution.sh \
```
DwarKapex, line 60: It fails, if UPDATE_PATHCES == false, with error `create-disctribution.sh is not found`

.github/container/Dockerfile.base
terrykong: I guess it would be better to have it be managed by `apt`, then the package can be inspected by `dpkg`. FWIW, I had to install `apt-get install software-properties-common` to get `add-apt-respository` which is 68.4MB according to `apt`. Is that okay with you?

.github/container/Dockerfile.pax.arm64
terrykong: I didn't author this dockerfile so I defer to @joker-eph regarding `chex`

W.r.t `EOT` vs `EOF`, either works, but one section of the dockerfile has two nested here-docs, so you need to choose a different char sequence, otherwise bash will end the heredoc early. There's one above it using EOT, but not nested; I can change that to EOF, which is more common

terrykong: I can rename the nested one to `EOF` and `EOFINNER` to be more clear

.github/container/bump.sh
terrykong: Done

.github/container/bump.sh
terrykong: Good catch. I described "--only-bump-patches" twice. I updated the description

.github/container/bump.sh
terrykong: Yea, sounds better. Done.

.github/container/bump.sh
terrykong: Done

.github/container/manifest.yaml
terrykong: Done

.github/container/create-distribution.sh
terrykong: Done

.github/container/get-source.sh
terrykong: Done

.github/workflows/_ci.yaml
terrykong: done

.github/workflows/_sandbox.yaml
terrykong: Done

rosetta/tests/upstream-only-distribution.sh
terrykong: If I make that change, this test script will only be runnable within the CI. At the top I set `SCRIPT_DIR` and `cd` into it, so this path will always be correct, unless it moves; and if it does, this script will error

rosetta/tests/upstream-only-distribution.sh
terrykong: So `create-distribution.sh` will create the patches alongside it, which will clutter your local clone with untracked files. I don't want the git source to change (since I'd need to `trap` to clean up), so I move it into a workspace dir.

rosetta/Dockerfile.pax
terrykong: Could you share how you ran this? I can look into this

.github/container/Dockerfile.pax.amd64
terrykong: So I took a look at it, and it appears we could merge the two without any issue since there's just a `sed` in [Dockerfile.pax.amd64](https://github.com/NVIDIA/JAX-Toolbox/pull/405/files#diff-249510ffdb78ebd74d306e940904393c34b2d8d580d2492a5cd60c58867cb5f4R16) that replaces the line, so I think it'd be fine to do that in requirements-jax.in

Any objections? @DwarKapex @yhtang 

Actually, another issue I found looking at this is that Dockerfile.pax.arm64 doesn't include this `sed`. Is that intentional @yhtang ?

terrykong:
Summarizing the topic/concerns @yhtang and I were discussing offline regarding the manifest file and `--build-arg`s

# Current state

The current implementation has a `--build-arg` to accept the manifest in the first build (Dockerfile.jax), and takes away the other `--build-arg`s that added the `(REPO_REF)_.*` args. 
* Why: This was done b/c it was simpler and b/c there was no way to pass the trial branch between `workflow_run`s without adding logic to propagate parent artifacts, or adding logic to trace the root workflow that triggered this tree of workflows

# Concern
The concern is that introducing the manifest makes it difficult to understand what's happening in each dockerfile, since the manifest is presumed to be there and its contents are not visible unless inspecting the base container (from Dockerfile.jax)

# Proposals

## (1) Keep `--build-arg MANIFEST_FILE` for this PR, but later re-introduce `--build-arg (REPO|REF)_` once CI is consolidated (#416) to __override__ the manifest

## (2) Keep `--build-arg MANIFEST_FILE` for this PR, but later re-introduce `--build-arg (REPO|REF)_` and derive them from the manifest once CI is consolidated (#416)

I am okay with either (1) or (2), but the point is I think it's best to keep `--build-arg MANIFEST_FILE` for now since the current state of the CI is not ready to handle REPO|REF --build-args. Anything I write here will be throw-away and undone with @DwarKapex tackles #416

@yhtang are you okay with keeping the PR as is, and @DwarKapex  tackling your `--build-arg` concern in the follow up PR? I presume what you had in mind was more like (2) than (1).

.github/container/Dockerfile.pax.arm64
joker-eph: If I wrote it, that was likely because it was required at the time to make the container work on ARM64.
We have lot of issues with incompleteness of the ARM64 pip packages, and the order in which we install things, their incompatible requirements, etc.

Ultimately line 27 and here are trying to get to a different state (earlier is just the requirement for building PAXML I believe?).
It is also very possible that the situation evolved since (that was over the summer if I remember correctly) and remove this from here would work.

terrykong:
* Weekly base build:  https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7147399495 
    *
* Nightly Jax Build: 
    * Image:
* Nightly JAX Unit Test: 
* Nightly Pax Build:
    * image:
* Nightly Pax MGMN Test: 
* Nightly T5x Build:
    * image:
* Nightly T5x MGMN Test:
* Nightly Rosetta t5x build and test:
* Nightly Rosetta pax build and test:

.github/workflows/nightly-rosetta-pax-build.yaml
```diff
@@ -105,6 +100,28 @@ jobs:
       BASE_IMAGE: ${{ needs.metadata.outputs.BASE_IMAGE_ARM64 }}
     secrets: inherit
 
+  publish-build-badge:
+    needs: [metadata, amd64, arm64]
+    uses: ./.github/workflows/_publish_badge.yaml
```
yhtang, line 105: Note that `_publish_badge.yaml` is deprecated. In the future, we will generate Badge JSONs as part of the job sitrep artifact, and then collectively publish them using `_fianlize.yaml`.

.github/workflows/nightly-rosetta-pax-build.yaml
```diff
@@ -105,6 +100,28 @@ jobs:
...
+    if: always()
```
yhtang, line 106: Instead of `always()`, use `!cancelled()` to prevent the step from running even if the workflow is cancelled.

.github/workflows/nightly-rosetta-pax-build.yaml
```diff
@@ -149,38 +167,42 @@ jobs:
...
     secrets: inherit
 
-  publish-test:
-    needs: [metadata, amd64, arm64, test-pax]
+  # TODO: ARM
+  publish-test-badge:
+    needs: [metadata, amd64, test-amd64]
     uses: ./.github/workflows/_publish_badge.yaml
-    if: ( always() )
+    if: always()
```
yhtang, line 174: `"!cancelled()"`?

.github/workflows/nightly-rosetta-t5x-build-test.yaml
```diff
@@ -95,20 +90,46 @@ jobs:
...
       - name: Generate placeholder warning
         shell: bash -x -e {0}
         run: |
           echo "WARNING: arm64 build is not yet supported"
-    
+
+  # TODO: ARM
+  publish-build-badge:
+    needs: [metadata, amd64, arm64]
+    uses: ./.github/workflows/_publish_badge.yaml
```
yhtang, line 109: Same comment for `_publish_badge.yaml`.

.github/workflows/nightly-rosetta-t5x-build-test.yaml
```diff
@@ -117,65 +138,65 @@ jobs:
...
       EXPERIMENT_SUBDIR: ROSETTA_T5X
       ARTIFACT_NAME: "rosetta-T5X-"
     secrets: inherit
 
-  publish-test:
-    needs: [metadata, amd64, arm64, test-unit, test-t5x]
+  # TODO: ARM
+  publish-test-badge:
+    needs: [metadata, amd64, test-unit-amd64, test-t5x-amd64]
     uses: ./.github/workflows/_publish_badge.yaml
```
yhtang, line 182: Same comment for `_publish_badge.yaml`.

yhtang (changes requested):

yhtang:
I feel that the way that we pass the version bump targets around could be better streamlined. The current trial branch approach seems a little risky. Admittedly, this could be done much more easily if the presubmits and the nightlies can be converged first.

.github/workflows/nightly-pax-build.yaml
```diff
@@ -58,11 +68,26 @@ jobs:
...
 
+      - name: Set base image
+        id: base-image
+        shell: bash -x -e {0}
+        run: |
+          if [[ -z "${{ inputs.BASE_IMAGE }}" ]]; then
+            BASE_IMAGE_AMD64=${{ env.DOCKER_REGISTRY }}/jax:mealkit
+            BASE_IMAGE_ARM64=${{ env.DOCKER_REGISTRY }}/jax:mealkit
+          else
+            BASE_IMAGE_AMD64=${{ inputs.BASE_IMAGE }}-amd64-mealkit
```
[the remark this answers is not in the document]
terrykong: I am following the convention @DwarKapex introduced. I understand the reason he introduced this was because we don't provide a multiarch tag within `jax-toolbox-internal` if publish=false. We could provide such a tag in `jax-toolbox-internal` even if publish=false and that should simplify this. How does that sound?

.github/workflows/nightly-jax-build.yaml
```diff
@@ -38,20 +43,64 @@ jobs:
         run: |
           echo "PUBLISH=${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && inputs.PUBLISH) }}" >> $GITHUB_OUTPUT
 
-  amd64:
+  bump-world-state:
```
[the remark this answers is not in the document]
terrykong: FYI @DwarKapex ^

.github/workflows/nightly-rosetta-pax-build.yaml
terrykong: In this case, I think we want `always()`. If upstream workflows fail, then this workflow is canceled, but if the badge creation only happens if `!cancelled()`, then the badge does not get updated, and we see yesterday's badge.

.github/workflows/nightly-rosetta-pax-build.yaml
terrykong: responded on other thread

.github/workflows/nightly-rosetta-t5x-build-test.yaml
terrykong: responded on other thread

.github/workflows/nightly-rosetta-t5x-build-test.yaml
terrykong: responded on other thread

terrykong:
> I feel that the way that we pass the version bump targets around could be better streamlined. The current trial branch approach seems a little risky. Admittedly, this could be done much more easily if the presubmits and the nightlies can be converged first.

@yhtang I don't think there's any risk b/c the ability to merge the trial branch isn't included in this change-set. Nightlies will continue to be built via `bump.sh`. If nightlies and presubmits converged, I agree this would be more easily done, but I think we can address that after this is in. If we wait for the converge PR, that'll add +N weeks and then this change needs to be rebased after N weeks of changes and that seems like avoidable dev-work.

I am in favor of merging this in as it works and the converge PR can make use of the mechanism I am introducing here.

.github/workflows/nightly-rosetta-pax-build.yaml
terrykong: I agree we should move to sitrep, but given how large this PR is, I'd like to tackle that in follow-up PR. This shouldn't change the status-quo; in fact it should fix things b/c this badge is referenced on the front-page readme and it was somehow removed in a PR so it hasn't been getting updated

.github/workflows/_ci.yaml
```diff
@@ -106,12 +63,7 @@ jobs:
       ARCHITECTURE: ${{ inputs.ARCHITECTURE }}
       BUILD_DATE: ${{ needs.metadata.outputs.BUILD_DATE }}
       BASE_IMAGE: ${{ needs.build-base.outputs.DOCKER_TAG }}
-      REPO_JAX: ${{ needs.metadata.outputs.REPO_JAX }}
-      REF_JAX: ${{ needs.metadata.outputs.REF_JAX }}
-      REPO_XLA: ${{ needs.metadata.outputs.REPO_XLA }}
-      REF_XLA: ${{ needs.metadata.outputs.REF_XLA }}
-      REPO_TE: ${{ needs.metadata.outputs.REPO_TE }}
-      REF_TE: ${{ needs.metadata.outputs.REF_TE }}
+      TRIAL_BRANCH: ${{ inputs.TRIAL_BRANCH }}
```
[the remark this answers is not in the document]
terrykong: It's not provided to the dockerfile. You're looking at `_build_jax.yaml`'s `workflow_call.inputs`. 

The reason it's passed is because within `_build_jax.yaml` is the `actions/checkout@v3` and I want to checkout the state of the manifest+patches on the trial branch (see [this](https://github.com/NVIDIA/JAX-Toolbox/pull/405/files/0828455c4d9478eba642ab24a8d17140b84c51d9#diff-4e3cbf29dd2e5a1bbae7236b97ab898d70d297ee50fba4765d4a71b9b8696694R92)).

This solves the problem that occurs when building the Nightly builds off of the `main` branch, and we need to build from "staged" changes, which is accomplished by checking out the "trial branch".

.github/workflows/_ci.yaml
```diff
@@ -7,27 +7,11 @@ on:
...
-        required: true
-      SRC_PAXML:
-        type: string
-        required: true
-      SRC_PRAXIS:
+      TRIAL_BRANCH:
         type: string
+        description: 'Name of (branch|tag|sha) with bumped manifest and patches'
         required: true
+        default: 'main'
```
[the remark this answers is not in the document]
terrykong: Since it's required, I can omit the default, but I left it in as a form of documentation. I can remove if that's your preference

.github/workflows/_build_jax.yaml
```diff
@@ -67,6 +37,11 @@ on:
         description: 'User email in GIT to perform git pull/push'
         required: false
         default: 'jax@nvidia.com'
+      TRIAL_BRANCH:
+        type: string
+        description: 'Name of branch with bumped manifest and patches'
+        required: true
+        default: 'main'
```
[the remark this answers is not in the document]
terrykong: That's true. Since the arg is required, I can just removed the default. Is that okay with you?

.github/container/manifest.yaml
```diff
@@ -0,0 +1,103 @@
...
+  latest_verified_commit: 0274fa20b4ff194c1c118b94b5f778caa5d9a84a
+  mode: git-clone
+tensorflow-text:
+  # Used only in ARM pax builds
+  url: https://github.com/tensorflow/text.git
+  tracking_ref: v2.13.0
+  latest_verified_commit: 917a681d7220ebf9b62a08b6f9ce7b7db886ddef
+  mode: git-clone
+pydantic:
+  version: X.Y.Z
```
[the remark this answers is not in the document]
terrykong: Actually I planned on including this pin for the ARM build for pax, but I see that you removed that in the pip-compile change: https://github.com/NVIDIA/JAX-Toolbox/pull/371/files#diff-2f69dd1b68217a122ece65bac9df3f6763e53ca122f367ce90c6610ca21584e9L41

Did you happen to test that this is no longer needed? 

I didn't actually implement the pip-constraint mechanism in this PR since I didn't have an example, so if you've tested that we don't need this anymore, I can remove it for now. If you haven't, then I can implement it. I think given how pinning appears to be platform specific, I was thinking of something like this:
```yaml
pydantic:
  versions:
     arm64: ==1.10.13
  mode: pip-constraint
```

.github/container/get-source.sh
```diff
@@ -83,9 +95,8 @@ set -ex -o pipefail
 git clone ${GIT_REPO} ${INSTALL_DIR}
 pushd ${INSTALL_DIR}
 git checkout ${GIT_REF}
-git submodule init
-git submodule update --recursive
+git submodule update --init --recursive
```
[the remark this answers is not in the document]
terrykong: From my experience, most modern git versions will support this. I took a look at git's history and you have to go reeeeally far back (>11 years) to find a version that doesn't support these two flags:

* had it: 1.8.1.1 (11 years ago): https://github.com/git/git/blob/v1.8.1.1/Documentation/git-submodule.txt
* didn't have it: 1.6.1 (15 years ago): https://github.com/git/git/blob/v1.6.1/Documentation/git-submodule.txt

.github/container/Dockerfile.pax.arm64
```diff
@@ -24,39 +22,35 @@ RUN wget https://github.com/bazelbuild/bazelisk/releases/download/v1.17.0/bazeli
...
-COPY --from=tftext-builder /opt/tensorflow-text/tensorflow_text*.whl /opt/
+COPY --from=tftext-builder ${SRC_PATH_TFTEXT}/tensorflow_text*.whl /opt/
 
-RUN get-source.sh -f ${REPO_LINGVO} -r ${REF_LINGVO} -d ${SRC_PATH_LINGVO}
+RUN get-source.sh -l lingvo -m ${MANIFEST_FILE}
 
 # build lingvo
-RUN <<"EOT" bash -exu
-set -o pipefail
+RUN <<"EOF" bash -exu -o pipefail
```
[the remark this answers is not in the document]
terrykong: I mean we don't have to rename, but I feel that it's more clear since a question already came up about its significance. This way the identifier is also self-documenting

.github/workflows/_build_rosetta.yaml
```diff
@@ -21,9 +21,14 @@ on:
...
         default: 'NOT SPECIFIED'
+      ARTIFACT_NAME:
+        type: string
+        description: 'Name of the artifact zip file'
+        required: false
+        default: 'artifact-rosetta-build'
       BADGE_FILENAME:
         type: string
-        description: 'Name of the endpoint JSON file for shields.io badge'
+        description: 'Name of the endpoint JSON file for shields.io badge (w/o .json || arch || library)'
```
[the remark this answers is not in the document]
terrykong: Done

.github/workflows/_ci.yaml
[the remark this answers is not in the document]
terrykong: I removed the default

.github/workflows/_build_jax.yaml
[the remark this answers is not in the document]
terrykong: I removed the default

.github/workflows/nightly-rosetta-pax-build.yaml
yhtang: Sounds good. I just wanted to raise the issue so that you are aware of the coming change.

rosetta/rosetta/projects/t5x/README.md
```diff
@@ -197,7 +197,6 @@ t5x/contrib/gpu/scripts_gpu/singlenode_ft_frompile.sh \
 
 # Known Issues
 * There is a known sporadic NCCL crash that happens when using the T5x container at node counts greater than or equal to 32 nodes. We will fix this in the next release. The issue is tracked [here](https://github.com/NVIDIA/JAX-Toolbox/issues/194).
-* The T5x nightlies disable `NCCL_NVLS_ENABLE=0` ([doc](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html#nccl-nvls-enable)). Future releases will re-enable this feature.
```
terrykong, line 200: Note to self: this was mistakenly deleted during rebase (revert)

.github/container/manifest.yaml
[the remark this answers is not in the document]
yhtang: IIRC pydantic was blocking the installation of T5X because, back then, JAX requested a version that was too new for T5X and pip somehow did not want to downgrade. The issue doesn't seem to exist anymore once we moved to pip-compile.

yhtang:
I'll move to merge this PR once the new tests are done after the rebasing. Follow-up work could include:

1. Converge the pre-submit and nightly workflows, which will automatically solve the following issues
  - Existing issue: the logic to decide whether to publish a nightly badge depends on upstream status (hidden dependency)
  - New issue: a workflow can use a manifest file from a branch other than the one that it is triggered on
2. Avoid creating trial branches for pre-submits. Currently, a trial branch seems necessary if we want to test the edge/latest versions. This will be problematic if we matricize the CI, considering the number of trial branches that will be created by each matrix.
  - Related issue: the name of a trial branch is passed between various nightly workflows. A better approach is to always use the manifest file from the same branch that triggers the workflow. This should be fixed automatically if we complete (1).

yhtang:
Merging as the remaining tests are all matrix MGMN tests (with partially confirmed success) and there are ~200 backlog jobs in the queue at this moment.

PR #406, opened 2023-12-01
The manifest rebase

terrykong:


.github/container/Dockerfile.jax
```diff
@@ -1,16 +1,10 @@
 ARG BASE_IMAGE=ghcr.io/nvidia/jax-toolbox:base
-ARG REPO_JAX="https://github.com/google/jax.git"
-ARG REPO_XLA="https://github.com/openxla/xla.git"
-ARG REPO_FLAX="https://github.com/google/flax.git"
-ARG REPO_TE="https://github.com/NVIDIA/TransformerEngine.git"
-ARG REF_JAX=main
-ARG REF_XLA=main
-ARG REF_FLAX=main
-ARG REF_TE=main
+ARG SRC_MANIFEST_FILE=manifest.yaml
```
yhtang, line 2: I would prefer we place the manifest file parsing mechanism in the CI code, while maintaining simple Dockerfiles. Otherwise the Dockerfile will no longer be self-contained and less reusable by external users.

.github/container/Dockerfile.jax
```diff
@@ -22,22 +16,28 @@ ARG BUILD_DATE
...
+
+ADD --chmod=777 create-distribution.sh ${DEST_MANIFEST_DIR}/
+COPY ${SRC_MANIFEST_FILE} ${DEST_MANIFEST_DIR}/${SRC_MANIFEST_FILE}
+COPY patches/ ${DEST_MANIFEST_DIR}/patches/
+
+RUN get-source.sh -l jax -m ${MANIFEST_FILE}
 RUN --mount=type=ssh \
     --mount=type=secret,id=SSH_KNOWN_HOSTS,target=/root/.ssh/known_hosts \
-    git clone "${REPO_XLA}" "${SRC_PATH_XLA}" && cd "${SRC_PATH_XLA}" && git checkout ${REF_XLA}
+    get-source.sh -l xla -m ${MANIFEST_FILE}
```
yhtang, line 37: Similarly, `get-source` should be preferrably self-contained.

.github/container/bump.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
```
yhtang, line 1: `bump.sh` -> `.github/workflows/scripts/bump-manifest.sh` so that the logic remains in CI?

.github/container/manifest.yaml
```diff
@@ -0,0 +1,103 @@
+# Updated in: 2023-12-01
```
yhtang, line 1: Move to either `.github/workflows` or just root?

.github/container/Dockerfile.base
```diff
@@ -38,12 +38,24 @@ RUN <<"EOF" bash -ex
...
+# Patch is specific to 23.3.1
+# Generated via: "git diff > pip-vcs-equivalency.patch"
+ADD pip-vcs-equivalency.patch /opt/pip/
+RUN <<EOF bash -e -x
+cd /opt/pip
+git apply </opt/pip/pip-vcs-equivalency.patch
+git add -u
+git commit -m 'Adds JAX_TOOLBOX_VCS_EQUIVALENCY as a trigger to treat all github VCS installs for a package as equivalent. The spec of the last encountered version will be used'
+EOF
+RUN pip install --upgrade --no-cache-dir -e /opt/pip pip-tools && rm -rf ~/.cache/*
```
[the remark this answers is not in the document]
yhtang: Ah I see. That should be sufficient. BTW, another point to note is that giving the baking (pip-compile + pip-sync) step will be independent between the upstream and rosetta containers, it is technically possible to let the upstream JAX/T5X/PAX/etc packages use free-floating VCS installs and rosetta containers use pinned VCS installs. Whether it will bring in any advantage is another question, though.

.github/container/Dockerfile.jax
terrykong: Don't we want to be prescriptive here since the jax ecosystem is difficult to pin down and we don't want external users to run into the same issues as us? In this case the prescription is: an external user has to use a manifest file to build the jax-toolbox containers.

I agree we want the dockerfiles to be simple and self-contained. I guess "simple" here is subjective, but I think the change to only require a two build args + `COPY` to get the manifest in, takes away the many `--build-args` one would need to build any of the images. Could you elaborate on what I can make simpler?

terrykong:
Sorry @yhtang . I just realized we've been commenting on the wrong PR. I only meant to have this open to confirm github was better after the rebase. Will close this in favor of #405

PR #418, opened 2023-12-08
Fix syntax error in _build_rosetta

hemildesai:


terrykong (approved):
Assuming CI passes. FFTM

PR #419, opened 2023-12-11
Cancel orphaned slurm jobs

### LINKED ISSUE #415: Need to cancel orphaned SLURM jobs
Currently, if a multi-node CI job submits a SLURM job and then gets canceled while waiting for the SLURM job to run, the SLURM job will become orphaned: it will be left in the queue to run, thus wasting computing resources, even though its result will not be collected. To fix this, we need to:

1. Keep track of the submitted SLURM job id, e.g. using a step output variable
2. Define a cancellation step that runs if the job is canceled using the `if: canceled ()` condition to properly remove the SLURM job.

This will help to reduce the congestion in our CI cluster, especially when repeated updates to a PR cause the SLURM queue to be flooded with orphaned CI jobs.

yhtang:
Fixes #415 

Currently, if a multi-node CI job submits a SLURM job and then gets canceled while waiting for the SLURM job to run, the SLURM job will become orphaned: it will be left in the queue to run, thus wasting computing resources, even though its result will not be collected. This is fixed by:
1. Keeping track of the submitted SLURM job id using the `SLURM_JOB_ID` step output variable
2. Defining a cancellation step that runs if the job is canceled cancel the SLURM job.

The effect of cancellation is verified by manually cancelling the following sandbox workflow:
https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7173955368

### DIFF
--- a/.github/workflows/_runner_ondemand_slurm.yaml
+++ b/.github/workflows/_runner_ondemand_slurm.yaml
@@ -54,42 +54,57 @@ jobs:
         id: submit
         shell: bash -x -e {0}
         run: |
-          ssh ${{ secrets.CLUSTER_LOGIN_USER }}@${{ vars.HOSTNAME_SLURM_LOGIN }} \
-            sbatch --parsable \
-          <<"EOF"
-          #!/bin/bash
-          #SBATCH --job-name=${{steps.meta.outputs.JOB_NAME }}
-          #SBATCH --exclusive
-          #SBATCH --nodes=1
-          #SBATCH --tasks=1
-          #SBATCH --gpus-per-node=8
-          #SBATCH --time="${{ inputs.TIME }}"
-          #SBATCH --output=${{ steps.meta.outputs.LOG_FILE }}
+          SLURM_JOB_ID=$(
+            ssh ${{ secrets.CLUSTER_LOGIN_USER }}@${{ vars.HOSTNAME_SLURM_LOGIN }} \
+              sbatch --parsable \
+            <<"EOF"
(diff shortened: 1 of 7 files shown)


.github/workflows/_runner_ondemand_slurm.yaml
```diff
@@ -93,3 +94,17 @@ jobs:
             -e DOCKERD_IN_RUNNER=true \
             ghcr.io/yhtang/actions-runner-dind:ubuntu-22.04  # use personal repo for the time being to avoid auth/cost issues
           EOF
+
+          echo "SLURM_JOB_ID=$(cat ${SLURM_JOB_ID_FILE})" >> $GITHUB_OUTPUT
+
+      - name: Wait for SLURM job to complete
+        shell: bash -x -e {0}
+        run: |
+          while ssh ${{ secrets.CLUSTER_LOGIN_USER }}@${{ vars.HOSTNAME_SLURM_LOGIN }} squeue -j${{ steps.submit.outputs.SLURM_JOB_ID }} > /dev/null 2>&1; do echo "wait"; sleep 15; done
```
terrykong, line 103: Would it be more robust to exit only if the squeue command returns that the job status is completed? This looks like it'll fail if `ssh` fails to connect (for w/e networking reason)

yhtang: Makes sense. In fact we could harden all places of SSH usage for such possibilities. Maybe worth another PR for it.

terrykong (approved):

PR #421, opened 2023-12-12
[Draft] Make the t5x container work on Aarch64

nouiz:
Replace https://github.com/NVIDIA/JAX-Toolbox/pull/252 with updates for the new pip-compile changes.

grain-nightly is installed as a dependency. But the wheel is bugged. It is marked as arch independent, but it is arch dependent.
When I try to build locally, it fails.

### DIFF
--- a/.github/container/Dockerfile.t5x
+++ b/.github/container/Dockerfile.t5x
@@ -15,6 +15,46 @@ ARG REPO_T5X
 ARG REF_T5X
 ARG SRC_PATH_T5X
 
+# force a recent version to have latest protobuf dep
+# Install now as some build need them.
+RUN pip install "tensorflow_datasets==4.9.2"
+RUN pip install "tensorflow==2.13.0"
+RUN pip install "chex==0.1.7"
+RUN <<"EOF" bash -ex
+cd /opt
+git clone http://github.com/tensorflow/text.git
+cd text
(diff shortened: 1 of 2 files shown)


yhtang (commented):
To ensure Python dependencies are installed atomically by pip-tools in a single step, we need to refactor this Dockerfile to:
1. use multi-stage build, so that packages that are only needed for building certain dependencies are installed to some 'builder' stage but not carries over to the mealkit/final image.
2. instead of using `pip install` to directly install packages, add them to a `requirements-XYZ.in` file in `/opt/pip-tools.d`.

An example of this paradigm is in the [pax arm64 dockerfile](https://github.com/NVIDIA/JAX-Toolbox/blob/7eaa5d2f0fc192dcd902928fa958127273c89e0a/.github/container/Dockerfile.pax.arm64), where the `tftext-builder` and `lingvo-builder` stages are used to build the respectively wheels, while the actual packages to be installed in the final image is [specified via `/opt/pip-tools.d`](https://github.com/NVIDIA/JAX-Toolbox/blob/7eaa5d2f0fc192dcd902928fa958127273c89e0a/.github/container/Dockerfile.pax.arm64#L108).

nouiz:
Continued in https://github.com/NVIDIA/JAX-Toolbox/pull/484

PR #433, opened 2023-12-19
Use aarch distribution of protobuf instead of building from src for lingvo

hemildesai:
To fix ARM builds, instead of building protos from source, changed https://github.com/tensorflow/lingvo/blob/master/lingvo/repo.bzl#L235 to use the aarch_64 release. There was a proto related error when trying to build them from source which is hard to debug without upstream support. Using the `aarch64` release for protos directly avoids compilation of protos and fixes the issue. This skips tests for lingvo since there are some flaky tests (occurring on both aarch64 and x86). We can revisit the tests later (or sync with upstream team)


nouiz (approved):
LGTM

PR #436, opened 2023-12-20
Fix default base image with schedule workflow trigger

ashors1:
Github workflows do not have access to the `workflow-trigger` defaults when workflows are automatically triggered. Because of this, the jax `BASE_IMAGE` was empty when running the scheduled build, causing the build to fail.


nouiz (approved):
LGTM.

PR #439, opened 2023-12-21
Fix extract-patches to look for correct patch dir in image

hemildesai:
`extract-patches.sh` looks for incorrect dir in image which results in https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7279876719/job/19837106057#step:13:24

This PR fixes the script to look in the correct path and make that path configurable.

Example workflow - https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7282838670

### DIFF
--- a/rosetta/scripts/extract-patches.sh
+++ b/rosetta/scripts/extract-patches.sh
@@ -6,13 +6,14 @@ cd $SCRIPT_DIR
 if [[ $# -lt 1 || $# -gt 2 ]]; then
   echo "Copies the patches from within an image to the GIT_ROOT/rosetta/patches dir"
   echo
-  echo "Usage: $0 <image> <rosetta dir: default GIT_ROOT/rosetta/>"
+  echo "Usage: $0 <image> <patch dir in image: default /opt/manifest.d/patches> <rosetta dir: default GIT_ROOT/rosetta/>"
   exit 1
 fi
 
 IMAGE=$1
-ROSETTA_DIR=${2:-$(readlink -f ../)}
+IMAGE_PATCH_DIR=${2:-"/opt/manifest.d/patches"}
+ROSETTA_DIR=${3:-$(readlink -f ../)}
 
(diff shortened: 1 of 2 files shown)


rosetta/scripts/extract-patches.sh
```diff
@@ -6,13 +6,14 @@ cd $SCRIPT_DIR
...
   echo
-  echo "Usage: $0 <image> <rosetta dir: default GIT_ROOT/rosetta/>"
+  echo "Usage: $0 <image> <patch dir in image: default /opt/manifest.d/patches> <rosetta dir: default GIT_ROOT/rosetta/>"
   exit 1
 fi
 
 IMAGE=$1
-ROSETTA_DIR=${2:-$(readlink -f ../)}
+IMAGE_PATCH_DIR=${2:-"/opt/manifest.d/patches"}
+ROSETTA_DIR=${3:-$(readlink -f ../)}
```
terrykong, line 15: I think this should be the directory under `.github/container/patches`, so

```suggestion
ROSETTA_DIR=${3:-$(readlink -f ../../.github/container/)}
```

The patches used to be under `./rosetta/patches` but that was before the manifest was introduced.

terrykong: Actually, now that I think about it, this is no longer needed b/c the patches are stored in the trial branch. We can keep this script as a helper, but I think it would be better to just remove these lines:

https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/_build_rosetta.yaml#L144-L151

hemildesai: Removed in https://github.com/NVIDIA/JAX-Toolbox/pull/439/commits/f0f3c880a105bbf1c01f3a4e5bb82845543b09d2

terrykong (approved):
LGTM. Fftm assuming CI passes

PR #440, opened 2023-12-26
Synchronization of Trigger JAX Toolbox action jobs from XLA CI. 

### LINKED ISSUE #234: Add sitrep generation for the Rosetta containers build workflow

### LINKED ISSUE #237: Add sitrep generation for the TE tests workflow

### LINKED ISSUE #238: Add sitrep generation for the Rosetta tests workflow

hmonishN:
Added changes to synchronize the job execution result of ci workflow jobs that get's triggered from Trigger JAX Toolbox action repo as part of XLA CI. Following are the details of the changes added:

1. Added changes to generate sitrep file to dump the job result summary for rosetta t5x and pax build and test jobs, test_te job.
2. Added changes in finalize.yaml file to consolidate the sitrep files generated for each job into one single sitrep file (final_sitrep.json) and saved it into artifacts.

The final_sitrep.json file is parsed as part of the post processing step in Trigger JAX Toolbox action repo and the results is displayed in a tabular format.

Workflow run for above changes: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7332912108

Closes #238 
Closes #237 
Closes #234


.github/workflows/_finalize.yaml
```diff
@@ -8,7 +8,11 @@ on:
         description: Update the landing page badges with run outcomes
         default: false
         required: false
-
+      ARTIFACT_NAME:
+        type: string
+        description: 'Name of the artifact zip file'
+        required: false
+        default: 'artifact-final-job-status'
```
yhtang, line 15: Can we make the default name more intuitive so that it standup out among a bunch of artifacts that "this is THE final report"?

yhtang (changes requested):
LGTM with some minor nitpicks 😄 
Before merging, I would also like to manually inspect the collective sitreps that the workflow generates to make sure the contents match expectations.

.github/workflows/_build_rosetta.yaml
```diff
@@ -178,7 +178,7 @@ jobs:
       - name: Upload sitrep and badge
         uses: actions/upload-artifact@v3
         with:
-          name: ${{ env.ARTIFACT_NAME_FULL }}
+          name: ${{ inputs.ARTIFACT_NAME }}-${{ inputs.BASE_LIBRARY }}-${{ inputs.ARCHITECTURE }}
```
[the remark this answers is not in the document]
terrykong: Doesn't this change effectively undo what's on line 57?

.github/workflows/_build_rosetta.yaml
```diff
@@ -140,27 +140,27 @@ jobs:
...
-
+            
       - name: Generate sitrep
         if: success() || failure()
         shell: bash -x -e {0}
         run: |
           # bring in utility functions
           source .github/workflows/scripts/to_json.sh
 
-          badge_label='${{ inputs.BASE_LIBRARY }} ${{ inputs.ARCHITECTURE }} build'
```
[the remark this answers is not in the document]
terrykong: I think we shouldn't introduce `Rosetta` in the `badge_label=` since we made the product decision a while back to just offer/advertise "t5x" where "upstream-t5x" is the internal package in case we need that. So for consistency, my opinion is to remove it from the summary as well.

Also our badges are pretty minimal right now:
image
Assuming we introduce two badges (one for arm one for amd), perhaps we should have the badge label just be `${{ inputs.ARCHITECTURE }}` and show both the arm/amd build badges on the frontpage readme. I'm thinking this full name might actually be redundant.

I feel that each stages badge_label shouldn't contain the "base library" or "framework" since that will be part of the badge json endpoint filename. Does that align with how you thought of sitreps & badges @yhtang ?

.github/workflows/_test_te.yaml
```diff
@@ -51,20 +69,61 @@ jobs:
...
+            errors total_tests passed_tests failed_tests \
+            badge_label badge_color badge_message \
+          > /log/sitrep.json
+
+          schemaVersion=1 \
+          label="${badge_label}" \
+          message="${badge_message}" \
+          color="${badge_color}" \
+          to_json schemaVersion label message color \
+          > /log/$te-unit-test.json
```
terrykong, line 109: Is that dollar sign reference intentional?

.github/workflows/_test_te.yaml
```diff
@@ -106,6 +168,7 @@ jobs:
         id: submit
         shell: bash -O expand_aliases -x -e {0}
         run: |
+          cd $GITHUB_WORKSPACE
```
terrykong, line 171: Why do you `cd` to the workspace here, and in several other places?

.github/workflows/_build_rosetta.yaml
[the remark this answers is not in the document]
yhtang: In other words, it is OK for the JSON endpoint file to contain `"label": "T5X nightly build"`, and while we place the badge in the T5X row and the build column, we will replace the label with `https://img.shields.io/endpoint?label=nightly...`

terrykong:
Note to add all the sitrep issues this closes

.github/workflows/_build_rosetta.yaml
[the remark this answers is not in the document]
terrykong: Oh cool, I didn't know you could override, that voids my suggestion.

And just to address the initial question, `BASE_LIBRARY` doesn't encode the information of rosetta or upstream. Putting "Rosetta T5x" and "Upstream T5x" in the badge_label/summary sounds okay to me since we wouldn't be surfacing that in the front-page readme

.github/workflows/_build_rosetta.yaml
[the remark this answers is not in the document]
hmonishN: changed it back to: 
name: ${{ env.ARTIFACT_NAME_FULL }}

.github/workflows/_test_te.yaml
hmonishN: fixed the typo.

.github/workflows/_test_te.yaml
hmonishN: in _ci.yaml file since we run multiple build and test jobs for jax, t5x, pax I did notice that sometimes the current working dir gets changed because of previous job and is not the GITHUB_WORKSPACE dir due to which it either throws error for repo files not found (e.g. to_json.sh) or the files get dumped at different locations and when we are saving the artifact it either complains about the folder/file doesn't exist or the dumped file is missing. for eg. in this run: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/6741297375/job/18326079678#step:9:45

.github/workflows/_finalize.yaml
hmonishN: something like: artifact-the-final-report?

.github/workflows/_test_te.yaml
terrykong: Interesting. I never noticed this before, have you seen this before @yhtang ?

I'm surprised since it's not obvious previous jobs change the working directory. Out of curiosity, does setting the [default for the entire job work](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-setting-default-run-step-options-for-a-job)?

yhtang: Each GitHub Actions "job" runs in a clean VM so it's unlikely that a previous job can chang the working path of a later job. However, an earlier job *step* can change that for a later step.

For the particular job that failed, it looks like the reason is that the repo is not checked out at all. Note that GitHub Actions does not by default check out the triggering repo. It needs to be done using `actions/checkout`.

.github/workflows/_finalize.yaml
```diff
@@ -112,16 +116,22 @@ jobs:
...
-      #       echo "$(dirname $f): $(cat $f)," >>
-      #     done
+      - name: Concatenate all sitreps
+        shell: bash -x -e {0}
+        run: |
+          # combine all sitreps files into a single file, where each sitrep json sits
+          # in a field named by the folder that contained it
 
+          echo "{" >> final_sitrep.json
+          find -name "sitrep.json" | while read -s f; do
```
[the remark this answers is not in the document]
hmonishN: Used jq to concatenate all json files.

.github/workflows/_finalize.yaml
```diff
@@ -112,16 +116,22 @@ jobs:
...
+            echo "\"$(dirname $f)\"": "$(cat $f)," >> final_sitrep.json
+          done
+          echo "\"END\": 1 }" >> final_sitrep.json
+      - name: Upload training logs as artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ inputs.ARTIFACT_NAME }}
+          path: ./final_sitrep.json
```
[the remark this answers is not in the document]
hmonishN: changed it to sitrep.json

.github/workflows/_test_te.yaml
hmonishN: Yeah I did add checkout step too but sometimes the current working dir gets changed maybe because of earlier job step. I do cd just to make sure I am dumping the data in the right dir to save it into artifact, it's just a safe thing to do without causing any issue.

yhtang: That's perfectly fine.

yhtang:
> Note to add all the sitrep issues this closes

I just added these to the PR description.

yhtang (approved):

PR #327, opened 2023-10-21
Hmonish test t5x fmha

hmonishN:
Adding the JAX T5x FMHA E2E system test to check for fmha lowering support. Following are the steps implemented in the test:

1. Enabling the fmha lowering flag along with dumping the hlo to track fmha instructions.
2. Jitted function in training loop.
3. After the test is finished check for the fmha instruction in the dumped hlo files.

PR #442, opened 2023-12-29
Add FMHA T5x test

hmonishN:
Adding the JAX T5x FMHA E2E system test to check for fmha lowering support. Following are the steps implemented in the test:

FMHA lowering flag is enabled by default now, enabled the dumping of hlo to track fmha forward and backward instructions.
Added the test as part of _ci.yaml file and also added a nightly workflow file for it. We will add this test as part of performance benchmarking later and add hlo to baseline.
Also added changes for correction of seq length of decoder (should be a multiple of 64)

Currently the test fails with following error related to CUDNN_STATUS_BAD_PARAM. The PR [PR](https://github.com/openxla/xla/pull/6872) the fix is open in upstream and the test will pass after it is merged. [Bug](https://nvbugspro.nvidia.com/bug/4409713) for this error.


hmonishN:
Closed the old PR #327 as this one is the latest one.

.github/container/test-t5x-fmha.sh
```diff
@@ -0,0 +1,244 @@
+#!/bin/bash
```
terrykong, line 1: Can the hlo dump feature be merged with test-t5x.sh since saving hlo dumps sounds generally useful?

Here's the current diff:
```diff
$ diff test-t5x*
23d22
<     echo "  --save-hlo {0, 1}         1 to save the dumped hlo, 0 to remove the hlo dumped folder"
47d45
< SAVE_HLO=${SAVE_HLO:-1}
87,90d84
<         --save-hlo)
<             SAVE_HLO="$2"
<             shift 2
<             ;;
114,119d107
< # Set hlo dump folder after output folder is set.
< HLO_DIR=${OUTPUT}/hlo
<
< ## Setting the env variables for FMHA
< export XLA_FLAGS="--xla_dump_hlo_as_text --xla_dump_to=${HLO_DIR} --xla_gpu_graph_level=0 --xla_gpu_fused_attention_use_cudnn_rng=true"
<
134,135d121
< print_var HLO_DIR
< print_var SAVE_HLO
223,244d208
<
< ## Check if fmha instructions are present in the HLO dumped file or not.
< fmha_regex="fmha[-bmm]?[-scale]?[-bias]?[-mask]?[-softmax]?[-dropout]?[-bmm]?[-backward]?*"
<
< result=$(grep -irlnE "$fmha_regex" "${HLO_DIR}/"*.txt)
<
< if [[ $SAVE_HLO -eq 0 ]]; then
<       rm -rf $HLO_DIR
<       echo "Removed dumped HLO directory!"
< fi
<
< if [ -z "$result" ]; then
<         echo "E: No FMHA instructions were found in the hlo files!"
<       exit 1
< else
<       echo -e "Found FMHA instructions in the following HLO files: \n $result" | while read -r line; do
<               echo "$line"
<       done
<         exit 0
<
< fi
<
```
The hlo dump gets put into the OUTPUT_DIR. How large is the hlo dump for this model size (since it will be saved as a github artifact)? 

At the moment, I feel that this whole script should be merged with test-t5x.sh and the "check fmha in hlo dump" be behind a flag so that it's not tested if not enabled. What do you think?

.github/container/test-t5x-fmha.sh
```diff
@@ -0,0 +1,244 @@
...
+if [[ $BATCH_SIZE == 0 ]]; then
+    echo "ERROR: Batch size must be specified."
+    usage 1
+fi
+
+# Set hlo dump folder after output folder is set.
+HLO_DIR=${OUTPUT}/hlo
+
+## Setting the env variables for FMHA
+export XLA_FLAGS="--xla_dump_hlo_as_text --xla_dump_to=${HLO_DIR} --xla_gpu_graph_level=0 --xla_gpu_fused_attention_use_cudnn_rng=true"
```
terrykong, line 118: So these flags will override anything in the container, which could possibly be needed. I prefer this pattern for setting these env vars:
```sh
export BASE_XLA_FLAGS="${BASE_XLA_FLAGS:---xla_dump_hlo_as_text --xla_dump_to=${HLO_DIR} --xla_gpu_graph_level=0 --xla_gpu_fused_attention_use_cudnn_rng=true}"
export XLA_FLAGS="${BASE_XLA_FLAGS} ${XLA_FLAGS:-}"
```

That way you pick up the environment, but you can still "zero out" the env var in this script without a code base by just doing `BASE_XLA_FLAGS=""`

terrykong (changes requested):

terrykong:
Moved to draft since there still seems to be active development on this PR.

BTW, it may be better to add this to _test_t5x.yaml since we were already running pretty close to the maximum number of reusable workflows allowed in github.

.github/container/test-t5x-fmha.sh
hmonishN: Merged and enabled dumping of hlo folder in t5x as well. The XLA dev team also wants to dump the hlo for paxml tests, will add changes for that as part of FMHA for paxml. The hlo folder size is 49-55 MB per testcase job.

.github/container/test-t5x.sh
```diff
@@ -178,7 +178,7 @@ $(
 import dummy_wikipedia
 
 MIXTURE_OR_TASK_NAME = "dummy_wikipedia"
-TASK_FEATURE_LENGTHS = {"inputs": 512, "targets": 114}
+TASK_FEATURE_LENGTHS = {"inputs": 512, "targets": 128}
```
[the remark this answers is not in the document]
hmonishN: The right sequence length for decoder should be a multiple of 64 in general also suggested by XLA dev team. It was passing without FMHA but will fail with FMHA feature enabled hence it's better to change it to 128.

.github/container/test-t5x-fmha.sh
hmonishN: Incorporated review comments.

.github/container/test-t5x.sh
```diff
@@ -206,3 +234,29 @@ ENABLE_TE=$ENABLE_TE python -m t5x.train \
     $ADDITIONAL_ARGS \
     $([[ $MULTIPROCESS != 0 ]] && echo --multiprocess_gpu)
 echo "Output at ${OUTPUT}"
+
+if [[ "$ENABLE_FMHA" -eq "1" ]]; then
```
terrykong, line 238: Nit: the indenting in this if block is kind of weird. Do you mind normalizing it?

.github/container/test-t5x.sh
```diff
@@ -206,3 +234,29 @@ ENABLE_TE=$ENABLE_TE python -m t5x.train \
...
+    fi
+
+    if [ -z "$result" ]; then
+        echo "E: No FMHA instructions were found in the hlo files!"
+	    exit 1
+    else
+	    echo -e "Found FMHA instructions in the following HLO files: \n $result" | while read -r line; do
+		echo "$line"
+	done
+    	exit 0
```
terrykong, line 255: This shouldn't be needed

.github/container/test-t5x.sh
```diff
@@ -206,3 +234,29 @@ ENABLE_TE=$ENABLE_TE python -m t5x.train \
...
+    if [[ $SAVE_HLO -eq 0 ]]; then
+	    rm -rf $HLO_DIR
+ 	    echo "Removed dumped HLO directory!"
+    fi
+
+    if [ -z "$result" ]; then
+        echo "E: No FMHA instructions were found in the hlo files!"
+	    exit 1
+    else
+	    echo -e "Found FMHA instructions in the following HLO files: \n $result" | while read -r line; do
```
terrykong, line 252: Maybe I'm missing something, but why the `while`? `echo -e "$multiline_string_var" should preserve newlines

.github/workflows/_test_upstream_t5x.yaml
```diff
@@ -28,6 +28,11 @@ on:
         description: 'Name of the framework being used'
         required: false
         default: 'upstream-t5x'
+      ENABLE_FMHA:
```
terrykong, line 31: Is it okay that fmha is tested on demand since the default = 0?

If you'd rather it be part of the test suite, then here's a reference for how rosetta t5x tests look: 
https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/_test_t5x_rosetta.yaml#L38-L53

Basically we matrixify the args passed and the JOB_NAME is defined as part of the matrix.

terrykong (changes requested):

.github/container/test-t5x.sh
hmonishN: Incorporated review comment.

.github/container/test-t5x.sh
hmonishN: Incorporated review comments

.github/workflows/_test_upstream_t5x.yaml
hmonishN: Since we dispatch ci workflow and fmha is relevant to xla I have disabled it by default and only enable it here: https://github.com/NVIDIA/JAX-Toolbox/blob/add_t5x_fmha_test/.github/workflows/_ci.yaml#L249 That way separate job with fmha in the name is triggered with the job matrix. Why should I matrixify the args in this case?

.github/workflows/_test_upstream_t5x.yaml
```diff
@@ -174,8 +192,47 @@ jobs:
   t5x-multi-node:
     strategy:
       matrix:
-        N_GPU: [1, 2, 4, 8]
-        N_NODE: [1, 2]
+        include:
+          - TEST_NAME: "1N1G"
```
terrykong, line 196: There are some assumptions that the test names should match [their older names ](https://github.com/NVIDIA/JAX-Toolbox/tree/main/.github/workflows/baselines/T5X_MGMN/upstream)so that baselines are found. Do you mind permuting the nodes and gpus? So `2N8G` -> `8G2N`?

terrykong (approved):
Tests that were added passed; metrics fail due to issue already resolved in `main`

PR #446, opened 2024-01-03
Fixes issue where dated rosetta nightlies were not tagged if tests failed

terrykong:
The issue was that if tests failed, the `publish-final` step wasn't run b/c a failure in any of the `needs:` jobs would implicitly cause `publish-final` to be skipped. So the ternary operator to conditionally add `latest` wasn't run and `nightly-YYYY-MM-DD` was skipped.

Simple remedy is to introduce two publish steps. The code is repeated, but it is much clearer than adding complicated `if:` logic

### DIFF
--- a/.github/workflows/nightly-rosetta-pax-build.yaml
+++ b/.github/workflows/nightly-rosetta-pax-build.yaml
@@ -183,19 +183,9 @@ jobs:
       TARGET_TAGS: |
         type=raw,value=latest,priority=1000
 
-  test-pax:
-    # needs: [metadata, amd64, arm64]
-    needs: [metadata, amd64]
-    uses: ./.github/workflows/_test_pax_rosetta.yaml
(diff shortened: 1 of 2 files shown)


hemildesai (dismissed):

terrykong:
Change will not be tested in presubmit codepath, so need to run manual runs but with "dummy" failed test to validate dated tags pushed:
* t5x: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7402931506
* pax: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7402930936
Will merge if output is as expected

terrykong:
Also noticed that pax is runnings the entire test suite twice, so removing that in this PR as well

terrykong:
Apologies, can you re-review @hemildesai

hemildesai (approved):

PR #447, opened 2024-01-03
Copy patches to correct location in Rosetta dockerfiles with UPDATE_PATCHES=true

hemildesai:
`cp -r /mnt/jax-toolbox/.github/container/patches ${MANIFEST_DIR}/patches` copies to a subdirectory inside `patches` and the structure looks like
```
${MANIFEST_DIR}/patches/patches/paxml
```

This PR removes the extra `patches` subdir


terrykong (approved):
LGTM

PR #451, opened 2024-01-05
Update XLA+arm64 patch

olupton:
https://github.com/openxla/xla/commit/49f45a12bcca8edc7f215887dbb112b2639d8a97 introduced some ARM intrinsic code that does not compile with GCC: https://godbolt.org/z/aYfds9Y15.
That commit also stopped the patch we apply to XLA from applying cleanly.
This updates the patch to disable vectorisation of that function when compiling for ARM with GCC.
It's not clear to me precisely which failure the old patch introduced in #367 was targeting (cc: @DwarKapex).

### DIFF
--- a/.github/container/manifest.yaml
+++ b/.github/container/manifest.yaml
@@ -2,31 +2,31 @@
 jax:
   url: https://github.com/google/jax.git
   tracking_ref: main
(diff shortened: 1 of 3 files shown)


nouiz (approved):

olupton:
Opened openxla/xla#8247 for the underlying issue.

PR #453, opened 2024-01-06
Publish ARM tags for upstream PAX image

hemildesai:
These ARM tags are required for downstream Rosetta PAX build jobs


terrykong (approved):
Lgtm

PR #456, opened 2024-01-09
Move manifest file to base container

DwarKapex:
This is the first step towards unifying Nightly and Presubmit builds.

The idea of this PR is to move manifest file to base container from jax container:
1. to avoid code repeating
2. as a preparation for the next step

The next step (the follow-up PR) involves modification of manifest file for Nightly Builds only (Presubmit builds will work with upstream manifest file, because it assumes changing in pipeline only).


yhtang (approved):
LGTM

PR #457, opened 2024-01-09
Remove duplication in JAX Dockerfile

olupton:
These lines are duplicated just above, we don't need them twice.


yhtang (approved):
Good catch 😄

PR #423, opened 2023-12-13
Build a container with Pallas enabled

olupton:

PR #459, opened 2024-01-09
Pallas container: publish containers under JAX, add nightly test and README entries

olupton:
Builds on #423. Publishes Pallas-enabled images as `jax:latest-pallas` etc. instead of `pallas:latest`.


yhtang (changes requested):
Please update the release location of the Pallas containers per our offline discussion 😄

yhtang (approved):
LGTM

PR #462, opened 2024-01-10
Adding a bidirectional CKPT converter of TE <-> Pax/T5X

mingxu1067:
Replaces #425.

### DIFF
--- a/rosetta/utils/te_pax_t5x_ckpt_converter/README.md
+++ b/rosetta/utils/te_pax_t5x_ckpt_converter/README.md
@@ -1,0 +1,142 @@
+# Bidirectional CKPT Converter for TE <-> T5X/Pax
+
+### Arguments
+```bash
+-h, --help            show this help message and exit
+--input-path INPUT_PATH
+                        the path to load a source checkponint for this conversion. (Required)
+--output-path OUTPUT_PATH
+                    the path to store the converted checkponint. (Required)
+--fw {pax,t5x}        the framework that stored the given source checkpoint. (Required)
+--direction {fw2te,te2fw}
+                    the direction of this conversion. (Required)
+--num-of-layer NUM_OF_LAYER
+                    the number of Transformer layer of the given source checkpoint. (Required)
+--num-of-head NUM_OF_HEAD
+                    the number of head of multi-head attention of the given source checkpoint. (Required)
+--head-dim HEAD_DIM   the head dimension of multi-head attention of the given source checkpoint. (Required)
+--mlp-intermediate-dim MLP_INTERMEDIATE_DIM
+                    the intermediate dimension of MLP block (FFN) of the given source checkpoint. (Required)
+--embed-dim EMBED_DIM
+                    the embeded dimension of the given source checkpoint, must give if --fw=t5x. (default: None)
+--kernel-chunk-size KERNEL_CHUNK_SIZE
+                    the size to chucnk kernel (weighs) then store, only support with --fw=pax. Setting None means no chunking. (default: None)
+--pax-repeat          indicate if the source Pax checkpoint enables Repeat. (default: False)
+--t5x-fuse-qkv        indicate if the source T5X checkpoint enables fused_qkv_params of TE. (default: False)
+```
+
+### Usage Examples
+#### Pax
+
+1. TE -> Pax (Repeat):
+```bash
+python  converter/main.py \
+    --input-path=/your_path_to_src_ckpt \
+    --output-path=/your_path_to_output_ckpt \
+    --fw=pax \
+    --direction=te2fw \
+    --pax-repeat \
+    --num-of-layer=8 \
+    --num-of-head=6 \
+    --head-dim=64 \
+    --mlp-intermediate-dim=1024
+```
+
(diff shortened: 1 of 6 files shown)


yhtang (approved):

terrykong:
@hemildesai @ashors to validate converter

ashors1:
It might be useful to provide some documentation on how to convert BF16 checkpoints to TE FP8 checkpoints, as IIRC the converter only works OOTB for TE BF16. WDYT @mingxu1067?

mingxu1067:
> It might be useful to provide some documentation on how to convert BF16 checkpoints to TE FP8 checkpoints, as IIRC the converter only works OOTB for TE BF16. WDYT @mingxu1067?

Yes, it currently only works OOTB for TE/BF16 only and requires some hacky trick to make the converted CKPT working for TE/FP8. Agree that adding a documentation would be helpful. For longtem support, I am thinking if we could auto-generate FP8 meta for those CKPTs trained on BF16 to further TE/FP8, then things could be much convenient to users.

yhtang:
@mingxu1067 Do you plan to add the documentation in this PR or in a subsequent one?

mingxu1067:
It would be a subsequent one.

ashors1:
> It would be a subsequent one.

In that case, can we add a relnote to the documentation outlining this limitation?

ashors1:
Another suggestion: Pax expects checkpoints to have the following directory structure: `${checkpoint_dir}/checkpoints/checkpoint_${step}`. If the checkpoints do not have this structure, restoring from the checkpoint will error. The Pax converter writes the flat checkpoint to `$output_path`. Perhaps it would be good to instead write the checkpoint to `${output_path}/checkpoints/checkpoint_${step}` so users can pass `$output_path` directly into pax without manually restructuring the directory.

mingxu1067:
> In that case, can we add a relnote to the documentation outlining this limitation?

Sure, where would the proper place to put the note?

> Another suggestion: Pax expects checkpoints to have the following directory structure: ${checkpoint_dir}/checkpoints/checkpoint_${step}. If the checkpoints do not have this structure, restoring from the checkpoint will error. The Pax converter writes the flat checkpoint to $output_path. Perhaps it would be good to instead write the checkpoint to ${output_path}/checkpoints/checkpoint_${step} so users can pass $output_path directly into pax without manually restructuring the directory.

Will sync with team to schedule this.

mingxu1067:
@hemildesai,
We added an argument `--skip-ln` to support T5 v1.1, please give it a try.

@ashors1,
> In that case, can we add a relnote to the documentation outlining this limitation?
We added a description in README.md to guide the steps to run the converted CKPT with TE + FP8. Please help to review.

> Another suggestion: Pax expects checkpoints to have the following directory structure: ${checkpoint_dir}/checkpoints/checkpoint_${step}. If the checkpoints do not have this structure, restoring from the checkpoint will error. The Pax converter writes the flat checkpoint to $output_path. Perhaps it would be good to instead write the checkpoint to ${output_path}/checkpoints/checkpoint_${step} so users can pass $output_path directly into pax without manually restructuring the directory.
After thinking, could we set output path with the same structure of Pax as a solution, like `--output-path=/your_path_to_output/checkpoints/checkpoint_${step}`?

mingxu1067:
> Yeah, that would work. It might be good to document this as well so users do not run into issues trying to pass the flat checkpoint into Pax.

Sure, I think it might be worth to document for both Pax and T5X. I know the CKPT structure for pax is like `/your_path_to_output/checkpoints/checkpoint_${step}/state`, and what is for T5X?

terrykong:
For T5x it looks like:
`/your_path_to_output/checkpoint_${step}/...`

Here's an example:
```
./checkpoint_144000
./checkpoint_144000/flax_mutables.fp8_meta_collection.encoder.layers_10.attention.qkv.fp8_max
./checkpoint_144000/flax_mutables.fp8_meta_collection.decoder.layers_5.self_attention.qkv.fp8_meta_scale
./checkpoint_144000/flax_mutables.fp8_meta_collection.encoder.layers_21.attention.out.fp8_meta_amax
./checkpoint_144000/flax_mutables.fp8_meta_collection.decoder.layers_13.encoder_decoder_attention.out.fp8_max
./checkpoint_144000/flax_mutables.fp8_meta_collection.decoder.layers_2.encoder_decoder_attention.query.fp8_meta_amax
./checkpoint_144000/flax_mutables.fp8_meta_collection.decoder.layers_13.encoder_decoder_attention.kv.fp8_meta_amax
./checkpoint_144000/flax_mutables.fp8_meta_collection.decoder.layers_18.mlp.fp8_meta_amax
```

mingxu1067:
Done, kindly review README.md and correct me if there are any santence not proper to publish.

yhtang (approved):

ashors1:
> I know the CKPT structure for pax is like /your_path_to_output/checkpoints/checkpoint_${step}/state

I think it should be `/your_path_to_output/checkpoints/checkpoint_${step}`. If using `/your_path_to_output/checkpoints/checkpoint_${step}/state`, pax also expects metadata which is not present with this converted checkpoint.

PR #463, opened 2024-01-10
Unification CI Presubmit and Nightly: Step 2

DwarKapex:
**This is the second step towards unification of the nightly and presubmit workflows.**

What is done:
1. Bump the manifest file in-place with the base container build, when the workflow is scheduled as a nightly run.
2. Remove the trial branch because it is no longer needed: the bumped manifest will be stored inside the base container.
3. In `_ci.yaml`, save the tags for all generated Docker images (both meal-kit and final) in a single JSON object (to get rid of long and hard-to-read variables in the `outputs` section) to be able to publish them in the `ci.yaml` workflow (`publish` job).
4. Move single-use workflows (`_test_distribution.yaml`) into the `ci.yaml` workflow to prevent hitting the 20-workflow limit for the CI.

**The next step is to add another job to update the manifest in the main branch**:
A. If the pipeline is all green, then create a PR with a bumped manifest, approve it, and merge it automatically.
B. If the pipeline is not all green, then create a PR and ping the maintainer of the repo (@yhtang, @terrykong) to perform a manual check of the pipeline and merge the PR manually if it’s acceptable.


.github/workflows/ci.yaml
```diff
@@ -22,21 +24,25 @@ permissions:
...
-            echo "TRIAL_BRANCH=${TRIAL_BRANCH}"
-          fi | tee $GITHUB_OUTPUT
+          BUILD_DATE=$(TZ='US/Los_Angeles' date '+%Y-%m-%d')
+          echo "BUILD_DATE=${BUILD_DATE}" >> $GITHUB_OUTPUT
+
+      - name: Determine whether results will be 'published'
+        id: if-publish
+        shell: bash -x -e {0}
+        run: |
+          echo "PUBLISH=${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && inputs.PUBLISH) }}" >> $GITHUB_OUTPUT
```
terrykong, line 44: You are missing `on.workflow_dispatch.inputs.PUBLISH`

.github/workflows/ci.yaml
```diff
@@ -54,10 +60,66 @@ jobs:
       TRIAL_BRANCH: ${{ needs.metadata.outputs.TRIAL_BRANCH }}
     secrets: inherit
 
+  publish-mealkit:
+    needs: [metadata, amd64, arm64]
+    if: "!canceled()" && github.event_name == 'schedule' && needs.metadata.outputs.PUBLISH == 'true'
```
terrykong, line 65: `github.event_name == 'schedule' ` looks like it's already captured in `needs.metadata.outputs.PUBLISH`

.github/workflows/ci.yaml
```diff
@@ -54,10 +60,66 @@ jobs:
...
+    strategy:
+      fail-fast: true
```
terrykong, line 67: I can't think of a case where one of these fails and the others don't (maybe an API failure?). But what do you think about letting it be `false` so that we at least get some images published

.github/workflows/ci.yaml
```diff
@@ -54,10 +60,66 @@ jobs:
...
+        ${{ needs.amd64.outputs.TAG_[format('{0}_MEALKIT', matrix.TAG)] }}
+        ${{ needs.arm64.outputs.[format('{0}_MEALKIT', matrix.TAG)] }}
+      TARGET_IMAGE: ${{ matrix.TARGET_IMAGE }}
+      TARGET_TAGS: |
+        type=raw,value=mealkit,priority=500
+        type=raw,value=mealkit-${{ needs.metadata.outputs.BUILD_DATE }},priority=500
+
+  publish-final:
+    needs: [metadata, amd64, arm64]
+    if: "!cancelled()" && github.event_name == 'schedule' && needs.metadata.outputs.PUBLISH == 'true'
```
terrykong, line 93: Same comment about `github.event_name == 'schedule'` being included in `needs.metadata.outputs.PUBLISH`

.github/workflows/ci.yaml
```diff
@@ -54,10 +60,66 @@ jobs:
...
+        type=raw,value=nightly-${{ needs.metadata.outputs.BUILD_DATE }},priority=500
+
   finalize:
-    needs: [amd64, arm64]
+    needs: [metadata, amd64, arm64]
     if: "!cancelled()"
     uses: ./.github/workflows/_finalize.yaml
     with:
-      PUBLISH_BADGE: false
+      PUBLISH_BADGE: ${{ metadata.outputs.PUBLISH }}
```
terrykong, line 124: `needs.metadata.outputs.PUBLISH`?

.github/workflows/ci.yaml
```diff
@@ -54,10 +60,66 @@ jobs:
...
+        - TAG: PALLAS
+          TARGET_IMAGE: pallas
+        - TAG: PAX
+          TARGET_IMAGE: upstream_pax
+        - TAG: T5X
+          TARGET_IMAGE: upstream_t5x
+        - TAG: ROSSETA_PAX
+          TARGET_IMAGE: pax
+        - TAG: ROSSETA_T5X
+          TARGET_IMAGE: T5x
```
terrykong, line 80: I'm not sure casing matters, but should this be `t5x`

terrykong (changes requested):

.github/workflows/ci.yaml
yhtang: Tags are indeed case-sensitive.

.github/workflows/_build_base.yaml
```diff
@@ -92,6 +92,15 @@ jobs:
           labels:
             org.opencontainers.image.created=${{ inputs.BUILD_DATE }}
 
+      - name: Update manifest and patches in-place - show diff
+        working-directory: .github/container
+        shell: bash -x -e {0}
+        run: |
+          build_event="${{ github.event_name }}"
```
yhtang, line 99: For better readability, I recommend the pattern where a dedicated predicate step would run the logic to determine whether version bump should happen, and then store the outcome as its step output. The actual bump step will then be conditional on the output of the predicate step. Take for example how the `PUBLISH` output is set in the `metadata` step of many nightly workflows.

yhtang: This way, it will also be easier to visually examine if the bump step did run in the GitHub Web GUI.

.github/workflows/_build_jax.yaml
```diff
@@ -37,10 +37,6 @@ on:
         description: 'User email in GIT to perform git pull/push'
         required: false
         default: 'jax@nvidia.com'
-      TRIAL_BRANCH:
```
yhtang, line 40: Is this duplicating part of #465?

.github/workflows/_ci.yaml
```diff
@@ -15,18 +15,43 @@ on:
       TAG_BASE:
         description: "Tags of the base image built"
         value: ${{ jobs.build-base.outputs.DOCKER_TAG }}
-      TAG_JAX:
-        description: "Tags of the JAX image built"
+      TAG_JAX_MEALKIT:
```
yhtang, line 18: Maybe we should just use a single JSON object to store all the mealkit/final tags, respectively. This is just a thought and please feel free to throw in yours :)

.github/workflows/ci.yaml
```diff
@@ -11,6 +11,8 @@ on:
         required: false
         default: ''
         description: 'Branch that contains manifest and patches. Default is empty which uses GITHUB_SHA'
+  schedule:
```
yhtang, line 14: Does the `workflow_dispatch` section still need the `TRIAL_BRANCH` input?

.github/workflows/ci.yaml
yhtang: Is this already resolved?

.github/workflows/ci.yaml
yhtang: +1 for `fail-fast: false`

.github/workflows/ci.yaml
yhtang: This is the `image` part of the complete tag, which is required to be in lower case:
```
Error parsing reference: "Fedora:37" is not a valid repository/tag: invalid reference format: repository name must be lowercase
```

yhtang (changes requested):

.github/workflows/_ci.yaml
```diff
@@ -15,18 +15,43 @@ on:
       TAG_BASE:
```
terrykong, line 15: can't comment on the line, but looks like you don't use TRIAL_BRANCH on line 10

.github/workflows/ci.yaml
terrykong: If TRIAL_BRANCH isn't used, then lines 52,60 aren't needed

.github/workflows/ci.yaml
terrykong: It doesn't look like it to me. The only workflow_dispatch input is TRIAL_BRANCH, but I don't see PUBLISH

.github/workflows/_build_base.yaml
DwarKapex: Done

.github/workflows/_build_jax.yaml
DwarKapex: Yes, because I started this PR before #465

.github/workflows/_ci.yaml
DwarKapex: Removed

.github/workflows/_ci.yaml
DwarKapex: Never used it before, but let me give a shot

DwarKapex: Not sure it looks better now, but more flexible I think

.github/workflows/ci.yaml
DwarKapex: Removed

.github/workflows/ci.yaml
DwarKapex: Added PUBLISH to inputs in `workflow_dispatch`

DwarKapex:
Addressed comments above but one.

.github/workflows/_ci.yaml
```diff
@@ -52,81 +32,143 @@ jobs:
...
+    steps:
+      - name: Build base container
+        id: container
+        uses: ./.github/workflows/_build_base.yaml
+        with:
+          ARCHITECTURE: ${{ inputs.ARCHITECTURE }}
+          BUILD_DATE: ${{ needs.metadata.outputs.BUILD_DATE }}
+      - name: Save base container tag for publishing step
+        run: |
+          echo "TAG_BASE=${{ steps.container.outputs.DOCKER_TAG }}" | tee -a $GITHUB_OUTPUT
```
terrykong, line 46: Is `TAG_BASE` used somewhere?

yhtang:
Converting the PR to draft to prevent accidental merge, as it appears to be in active development.

.github/workflows/_ci.yaml
DwarKapex: Removed

yhtang:
These workflows will be broken if we remove `_finalize.yaml`:
```
$ grep _finalize *.yaml
ci.yaml:    uses: ./.github/workflows/_finalize.yaml
nightly-base-build.yaml:    uses: ./.github/workflows/_finalize.yaml
nightly-jax-build.yaml:    uses: ./.github/workflows/_finalize.yaml
nightly-jax-test-unit.yaml:    uses: ./.github/workflows/_finalize.yaml
nightly-pallas-build.yaml:    uses: ./.github/workflows/_finalize.yaml
nightly-pallas-test-unit.yaml:    uses: ./.github/workflows/_finalize.yaml
nightly-pax-build.yaml:    uses: ./.github/workflows/_finalize.yaml
nightly-pax-test-mgmn.yaml:    uses: ./.github/workflows/_finalize.yaml
nightly-rosetta-pax-build.yaml:    uses: ./.github/workflows/_finalize.yaml
nightly-rosetta-t5x-build-test.yaml:    uses: ./.github/workflows/_finalize.yaml
nightly-t5x-build.yaml:    uses: ./.github/workflows/_finalize.yaml
nightly-t5x-test-mgmn.yaml:    uses: ./.github/workflows/_finalize.yaml
```

In the longer run, all these nightly workflows will no longer exist once we unify the presubmit and the nightlies. However, I don't think we can afford to have them broken during the transition period.

.github/workflows/_build.yaml
```diff
@@ -134,6 +134,12 @@ jobs:
             BASE_IMAGE=${{ inputs.BASE_IMAGE }}
             BUILD_DATE=${{ inputs.BUILD_DATE }}
 
+      - name: Save container tags (mealkit and final) for publishing step
+        run: |
+          CONTAINER_TAG_NAME=$(echo ${{ inputs.CONTAINER_NAME }} | tr '[:lower:]' '[:upper:]' | tr '-' '_')
+          echo "TAG_${CONTAINER_TAG_NAME}_MEALKIT=${{ steps.mealkit-metadata.outputs.tags }}" | tee -a $GITHUB_OUTPUT
```
yhtang, line 140: Given that this is a generic and reusable workflow, its output should also have generic names.
The outputs from different calls to this workflow can be distinguished by their job name.
i.e. we want
```
${{ needs.build_A.outputs.TAG_MEALKIT }}
${{ needs.build_B.outputs.TAG_MEALKIT }}
```
instead of
```
${{ needs.build_A.outputs.TAG_A_MEALKIT }}
${{ needs.build_B.outputs.TAG_B_MEALKIT }}
```

.github/workflows/_build_base.yaml
```diff
@@ -57,10 +52,22 @@ permissions:
 
 jobs:
 
+  metadata:
```
yhtang, line 55: This logic should be placed directly in `ci.yaml` since it is a global condition. `_build_base.yaml` can take the output of the logic as an input.

.github/workflows/_build_rosetta.yaml
```diff
@@ -141,6 +141,12 @@ jobs:
           build-args: |
             BASE_IMAGE=${{ steps.defaults.outputs.BASE_IMAGE }}
 
+      - name: Save container tags (mealkit and final) for publishing step
+        run: |
+          CONTAINER_TAG_NAME=$(echo ${{ inputs.BASE_LIBRARY }} | tr '[:lower:]' '[:upper:]' | tr '-' '_')
+          echo "TAG_ROSETTA_${CONTAINER_TAG_NAME}_MEALKIT=${{ steps.mealkit-metadata.outputs.tags }}" | tee -a $GITHUB_OUTPUT
```
yhtang, line 147: Similar comment like those for `_build.yaml`.

.github/workflows/_ci.yaml
```diff
@@ -7,26 +7,11 @@ on:
...
-      TAG_T5X:
-        description: "Tags of the T5X image built"
-        value: ${{ jobs.build-t5x.outputs.DOCKER_TAG_FINAL }}
-      TAG_PALLAS:
-        description: "Tags of the Pallas image built"
-        value: ${{ jobs.build-pallas.outputs.DOCKER_TAG_FINAL }}
-      TAG_PAX:
-        description: "Tags of the PAX image built"
-        value: ${{ jobs.build-pax.outputs.DOCKER_TAG_FINAL }}
+      ALL_TAGS:
```
yhtang, line 11: Since this is the only output related to tags, `ALL_` won't be necessary. How about just `CONTAINER_TAGS`?

.github/workflows/_ci.yaml
```diff
@@ -7,26 +7,11 @@ on:
...
+        description: Single JSON object to store all tags
```
yhtang, line 12: Maybe clarify these are docker tags rather than git tags?

.github/workflows/_ci.yaml
```diff
@@ -129,10 +113,55 @@ jobs:
...
+  save-tags:
+    runs-on: ubuntu-22.04
+    needs: [build-jax, build-t5x, build-pallas, build-pax, build-rosetta-t5x, build-rosetta-pax]
+    outputs:
+      ALL_TAGS: ${{ steps.save-docker-tags.outputs.ALL_TAGS }}
+    steps:
+      - name: Save docker tags in a single JSON object
+        id: save-docker-tags
+        run: |
+          ALL_TAG_JSON=$(cat << EOF
```
yhtang, line 125: Need to update the output names once `_build.yaml`/`_build_rosetta.yaml` is updated.

.github/workflows/_ci.yaml
```diff
@@ -129,10 +113,55 @@ jobs:
...
+            { \
+            "TAG_JAX_MEALKIT": "${{ needs.build-jax.outputs.DOCKER_TAG_MEALKIT }}", \
+            "TAG_JAX_FINAL": "${{ needs.build-jax.outputs.DOCKER_TAG_FINAL }}", \
+            "TAG_UPSTREAM_T5X_MEALKIT": "${{ needs.build-t5x.outputs.DOCKER_TAG_MEALKIT }}", \
+            "TAG_UPSTREAM_T5X_FINAL": "${{ needs.build-t5x.outputs.DOCKER_TAG_FINAL }}", \
+            "TAG_PALLAS_MEALKIT": "${{ needs.build-pallas.outputs.DOCKER_TAG_MEALKIT }}", \
```
yhtang, line 131: Is this consistent with `_build.yaml`:
```
echo "TAG_${CONTAINER_TAG_NAME}_MEALKIT=${{ steps.mealkit-metadata.outputs.tags }}" | tee -a $GITHUB_OUTPUT
```
should it be
```
"TAG_PALLAS_MEALKIT": "${{ needs.build-pallas.outputs.TAG_PALLAS_MEALKIT}}"
```
?

.github/workflows/ci.yaml
```diff
@@ -22,42 +25,233 @@ permissions:
...
     secrets: inherit
 
+  publish-mealkit:
+    needs: [metadata, amd64, arm64]
+    if: (!cancelled()) && needs.metadata.outputs.PUBLISH
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - TAG: JAX
```
yhtang, line 68: Looks like the correspondence between TAG and TARGET_NAME can be programmed:
- lower case -> upper case
- middle hyphen -> underscore

.github/workflows/ci.yaml
```diff
@@ -22,42 +25,233 @@ permissions:
...
+          - TAG: PAX
+            TARGET_IMAGE: pax
+          - TAG: T5X
+            TARGET_IMAGE: t5x
+    uses: ./.github/workflows/_publish_container.yaml
+    with:
+      SOURCE_IMAGE: |
+        ${{ fromJson(needs.amd64.outputs.ALL_TAGS)[format('TAG_{0}_MEALKIT', matrix.TAG)] }}
+        ${{ fromJson(needs.arm64.outputs.ALL_TAGS)[format('TAG_{0}_MEALKIT', matrix.TAG)] }}
+      TARGET_IMAGE: ${{ matrix.TARGET_IMAGE }}
```
yhtang, line 85: We probably don't want to publish N images in `ghcr.io/nvidia`, but this is worth discussing with everyone and the product team. @sbhavani

.github/workflows/ci.yaml
```diff
@@ -22,42 +25,233 @@ permissions:
...
+      TARGET_TAGS: |
+        type=raw,value=mealkit,priority=500
+        type=raw,value=mealkit-${{ needs.metadata.outputs.BUILD_DATE }},priority=500
+
+  publish-final:
```
yhtang, line 90: This publication workflow should be turned into a reusable one now that we are repeating it twice.

.github/workflows/ci.yaml
```diff
@@ -22,42 +25,233 @@ permissions:
...
+        ${{ fromJson(needs.amd64.outputs.ALL_TAGS)[format('TAG_{0}_FINAL', matrix.TAG)] }}
+        ${{ fromJson(needs.arm64.outputs.ALL_TAGS)[format('TAG_{0}_FINAL', matrix.TAG)] }}
+      TARGET_IMAGE: ${{ matrix.TARGET_IMAGE }}
+      TARGET_TAGS: |
+        type=raw,value=latest,priority=500
+        type=raw,value=nightly-${{ needs.metadata.outputs.BUILD_DATE }},priority=500
+
   finalize:
-    needs: [amd64, arm64]
+    runs-on: ubuntu-22.04
```
yhtang, line 120: Instead of breaking encapsulation and inlining the workflow here, how about we try to merge a few `_test_*.yaml` workflows?

yhtang (changes requested):

.github/workflows/ci.yaml
terrykong, line 68: Can this be simplified and derived from ALL_TAGS?

terrykong:
Also, should these lines be removed? https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/nightly-base-build.yaml#L5-L6

It looks like both the old nightlies and this new nightly both run at the same time.

.github/workflows/ci.yaml
terrykong: Just clarifying, is your suggestion to publish fewer, or push them elsewhere?

.github/workflows/_build.yaml
DwarKapex: This is leftover from my experiments. This is step is not required any more. All tags will be saved inside `_ci.yaml` workflow under `save-tags` job

.github/workflows/_build_base.yaml
DwarKapex: Moved to `ci.yaml`, as well as `BUILD_DATE`

.github/workflows/_build_rosetta.yaml
DwarKapex: Removed completely

.github/workflows/_ci.yaml
DwarKapex: Done

.github/workflows/_ci.yaml
DwarKapex: Done

.github/workflows/_ci.yaml
DwarKapex: Yes. See comments above.

.github/workflows/ci.yaml
DwarKapex: Yeah, I wanted to do that in place, but looks like I need to do it in 2 steps. Or is there anternative?

.github/workflows/ci.yaml
DwarKapex: OK, that is something new. As of now, when we run nightly we publish both mealkit and final. Why do we want to change the publishing politics now?

.github/workflows/ci.yaml
DwarKapex: Convert it to matrix to avoid introducing another workflow

.github/workflows/ci.yaml
DwarKapex: Reverted. Will do some investigation what can be merged to reduce the number of workflows.

.github/workflows/ci.yaml
DwarKapex: I tried to prototype it and almost found the solution, but the problem is that I have 2 different `ALL_TAGS` (one from amd64 build, the other one is from arm64 one). I need to combine these 2 `ALL_TAGS` into 1 JSON object somehow  and feed `matrix` with it.
But let's create an issue for further pipeline improvements after I merge this PR.
Agree? @terrykong

.github/workflows/ci.yaml
yhtang: Great that's a better idea.

.github/workflows/ci.yaml
yhtang: If we do it as a job step, then I guess it's not a problem if we use 2 or even more lines to compute it.

.github/workflows/ci.yaml
```diff
@@ -26,38 +29,85 @@ jobs:
...
       ARCHITECTURE: arm64
-      TRIAL_BRANCH: ${{ needs.metadata.outputs.TRIAL_BRANCH }}
+      BUILD_DATE: ${{ needs.metadata.outputs.BUILD_DATE}}
+      BUMP_MANIFEST: ${{ needs.metadata.outputs.BUMP_MANIFEST}}      
     secrets: inherit
 
+  publish:
+    runs-on: ubuntu-22.04
+    needs: [metadata, amd64, arm64]
+    if: ${{ !cancelled() }} && needs.metadata.outputs.PUBLISH
```
yhtang, line 76: `${{...}}` is not required but OK in a `if:` key (this is a GitHub Actions peculiarity IMHO)
Either way, it looks a bit strange since we have half of the condition inside of `${{...}}` and the other half out of it.

.github/workflows/ci.yaml
```diff
@@ -26,38 +29,85 @@ jobs:
...
+        run: |
+          CONTAINER_TAG=$(echo ${{ matrix.TARGET_IMAGE }} | tr '[[:lower:]]' '[[:upper:]]' | tr '-' '_')
+          TARGET_TAG_1=${{ matrix.FLAVOR }}
+          TARGET_TAG_2=${{ matrix.FLAVOR }}
+          if [[ "${{ matrix.FLAVOR }}" == "final" ]]; then
+            TARGET_TAG_1="latest"
+            TARGET_TAG_2="nightly"
+          fi
+          echo "CONTAINER_TAG=${CONTAINER_TAG}" >> $GITHUB_OUTPUT
+          echo "TARGET_TAG_1=${TARGET_TAG_1}" >> $GITHUB_OUTPUT
```
yhtang, line 94: Could we use a more descriptive suffix rather than `..._1` and `..._2`?

yhtang (changes requested):

.github/workflows/ci.yaml
yhtang: Seems like this is already resolved in the new 'publish' job.

.github/workflows/ci.yaml
```diff
@@ -26,38 +29,85 @@ jobs:
...
+            ${{ fromJson(needs.amd64.outputs.ALL_TAGS)[format('TAG_{0}_{1}', steps.transform.outputs.CONTAINER_TAG, matrix.FLAVOR)] }}
+            ${{ fromJson(needs.arm64.outputs.ALL_TAGS)[format('TAG_{0}_{1}', steps.transform.outputs.CONTAINER_TAG, matrix.FLAVOR)] }}
+          TARGET_IMAGE: ${{ matrix.TARGET_IMAGE }}
+          TARGET_TAGS: |
+            type=raw,value=${{ steps.transform.outputs.TARGET_TAG_1 }},priority=500
+            type=raw,value=${{ steps.transform.outputs.TARGET_TAG_2 }}-${{ needs.metadata.outputs.BUILD_DATE }},priority=500
+
   finalize:
     needs: [amd64, arm64]
-    if: "!cancelled()"
```
yhtang, line 59: This change is not necessary. Do you have a particular reason?

.github/workflows/ci.yaml
DwarKapex: Ah, got it. Thanks for clarification

.github/workflows/ci.yaml
DwarKapex: Removed ${{}} around canceled()

.github/workflows/ci.yaml
DwarKapex: Reverted

terrykong (dismissed):
LGTM, but would wait for @yhtang 's approval

terrykong:
Just a note on this comment in the description:
> B. If pipeline is not all green, then create PR and ping developers (Yu-Hang, Terry, me, someone else) to perform a manual check of pipeline and merge the PR manually.

One thing I'm worried about is the PR can become outdated, so it should either be closed or labelled as stale since a later nightly may update the manifest and then this PR would be considered a regression if accidentally merged. Not something to be addressed in this PR, but just a note.

DwarKapex:
> Just a note on this comment in the description:
> 
> > B. If pipeline is not all green, then create PR and ping developers (Yu-Hang, Terry, me, someone else) to perform a manual check of pipeline and merge the PR manually.
> 
> One thing I'm worried about is the PR can become outdated, so it should either be closed or labelled as stale since a later nightly may update the manifest and then this PR would be considered a regression if accidentally merged. Not something to be addressed in this PR, but just a note.

Agree, we need to set sort of timeout for automatically generated PRs.

.github/workflows/ci.yaml
```diff
@@ -26,34 +28,93 @@ jobs:
...
+        id: if-publish
+        shell: bash -x -e {0}
+        run: |
+          echo "PUBLISH=${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && inputs.PUBLISH) }}" >> $GITHUB_OUTPUT
+
+      - name: Determine whether need to bump manifest 
+        id: if-bump-manifest
+        shell: bash -x -e {0}
+        run: |
+          echo "BUMP_MANIFEST=${{ github.event_name == 'schedule' }}" >> $GITHUB_OUTPUT
```
yhtang, line 52: Please add a workflow input to allow  manual override of `BUMP_MANIFEST`, in case if we need to manually run the nightly workflow.

yhtang (changes requested):
LGTM but please add the `BUMP_MANIFEST` input to `ci.yaml` as commented below 😄 
Also, please revise the PR description for readability and typo, as it will become part of the commit message.

.github/workflows/ci.yaml
DwarKapex: Done

yhtang:
I'm seeing the error when running manually: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7626810302/job/20774071918#step:8:23

DwarKapex:
> https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7626810302/job/20774071918#step:8:23

Fixed: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7632038858/job/20791396138#step:7:56
@yhtang

.github/workflows/ci.yaml
```diff
@@ -26,34 +33,93 @@ jobs:
...
+                      CONTAINER_TAG=latest
+                      TAG_DATED=nightly
+                  fi
+                  JSON=$(echo ${JSON}\"${target}-${flavor}-container-tag\":\"${CONTAINER_TAG}\",)
+                  JSON=$(echo ${JSON}\"${target}-${flavor}-tag-dated\":\"${TAG_DATED}\",)
+              done
+          done
+          JSON="${JSON::-1} }"
+          echo "TARGET_TAGS=${JSON}" 
+          echo "TARGET_TAGS=${JSON}" >> $GITHUB_OUTPUT
```
terrykong, line 103: nit:
```suggestion
          echo "TARGET_TAGS=${JSON}" | tee -a $GITHUB_OUTPUT    
```

DwarKapex: Btw, why? What's the difference?

terrykong: In the orig, you print to stdout and right below you redirect (And append) to GITHUB_OUTPUT.

The suggestion was just to use `tee -a`, which appends, but also passes thru to stdout, so it is effectively the two lines you originally had.

yhtang (dismissed):

yhtang:
Note that after this is merged, we will have two sets of nightly runs: the old ones and the new one. We will continue using the old ones while we test the new one and make sure it is fully functional, before we flip the switch.

terrykong (changes requested):
>Note that after this is merged, we will have two sets of nightly runs: the old ones and the new one. We will continue using the old ones while we test the new one and make sure it is fully functional, before we flip the switch.

Will there be any issues if we let both of them run? One concern I have is both start at the same time, so if someone is looking at the `ghcr.io/nvidia/jax:{nightly,mealkit}...` tags, they won't know which pipeline it is from without digging into the digests

yhtang:
> > Note that after this is merged, we will have two sets of nightly runs: the old ones and the new one. We will continue using the old ones while we test the new one and make sure it is fully functional, before we flip the switch.
> 
> Will there be any issues if we let both of them run? One concern I have is both start at the same time, so if someone is looking at the `ghcr.io/nvidia/jax:{nightly,mealkit}...` tags, they won't know which pipeline it is from without digging into the digests

I've add `if: false` to the publish section of the new `ci.yaml`, so that we won't be publishing any containers using the unified nightly for the time being.

terrykong (approved):

PR #464, opened 2024-01-10
Maxtext docker

kocchop:
Adding support for MaxText to the JAX-toolbox


abhinavgoel95:
@terrykong could you review this PR? It adds MaxText support to Jax-Toolbox.

.github/workflows/_sandbox.yaml
```diff
@@ -1,41 +1,127 @@
-name: "~Sandbox"
+name: ~sandbox
```
terrykong, line 1: just leaving a reminder to revert before merging

.github/container/Dockerfile.maxtext
```diff
@@ -0,0 +1,37 @@
...
+## Download source and add auxiliary scripts
+###############################################################################
+
+FROM ${BASE_IMAGE} as mealkit
+
+ARG REPO_MAXTEXT
+ARG REF_MAXTEXT
+ARG SRC_PATH_MAXTEXT
+
+RUN <<"EOF" bash -ex
```
terrykong, line 18: Is there a particular reason to keep this out of the central manifest.yaml? Putting it here means that your `latest_verified_commit` won't be bumped by our CI so JAX will move ahead and maxtext will stay where it is.

.github/container/Dockerfile.maxtext
```diff
@@ -0,0 +1,37 @@
+# syntax=docker/dockerfile:1-labs
+
+ARG BASE_IMAGE=ghcr.io/nvidia/jax:mealkit
+ARG REPO_MAXTEXT=https://github.com/google/maxtext.git
+ARG REF_MAXTEXT=main
```
terrykong, line 5: lines 4-5 & 14-15 aren't needed

terrykong (changes requested):

terrykong:
If you'd like one extra guard before enabling building these images in the CI, you could contribute a nightly workflow like this one: https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/nightly-pax-build.yaml, except make it run a no-op. Then in a follow up PR you can fill its contents. This is to work around a limitation of Github Actions where you cannot run a workflow that does not exist on the `default` branch, which in our case is `main`

yhtang (changes requested):
Please add an associated build workflow as @terrykong suggested.

kocchop:
> If you'd like one extra guard before enabling building these images in the CI, you could contribute a nightly workflow like this one: https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/nightly-pax-build.yaml, except make it run a no-op. Then in a follow up PR you can fill its contents and test it. This is to work around a limitation of Github Actions where you cannot run a workflow that does not exist on the `default` branch, which in our case is `main`

kocchop:
@terrykong @yhtang I have made the necessary changes. I have added a placeholder file for nightly maxtext build. Could you please review and let me know?

.github/workflows/nightly-maxtext-build.yaml
```diff
@@ -0,0 +1,8 @@
+name: Nightly MaxText build
+'on': workflow_dispatch
```
terrykong, line 2: I'm not sure how strict GH actions is, but technically on.workflow_dispatch should be a dict. It doesn't need any args, see the sandbox for an example: https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/_sandbox.yaml#L4

terrykong (changes requested):

terrykong (approved):

terrykong:
LGTM

@yhtang any other comments?

abhinavgoel95 (dismissed):

yhtang (changes requested):
Shall we also build it in the presubmit?

Also, I noticed that the Dockerfile is only for amd64. What's the current status for this on arm?

kocchop:
Hi @yhtang currently there is a tensorflow version mismatch for arm. We may need to create a separate docker file with corresponding wheels. Right now focusing on x86 version to be working. Will work towards the arm build soon

kocchop:
Hi @yhtang , I have added maxtext to the presubmit CI. I have also tested the dockerfile with sandbox and it is working for amd64. Please let me know if you have any other suggestions.

yhtang (approved):
LGTM. Note that we need to add tests (smoke, functional, perf, etc.) via subsequent PRs.

PR #473, opened 2024-01-16
Merge jobs in _finalize.yml to reduce number of workflows (limit is 20)

DwarKapex:
Unfortunately, GitHub Actions has limits on number of workflows to run (now it is 20). To reduce number of workflows, I merged all jobs in `_finalize.yaml` into one job `finalize`.

### DIFF
--- a/.github/workflows/_finalize.yaml
+++ b/.github/workflows/_finalize.yaml
@@ -34,7 +34,9 @@ jobs:
   #         # | ROSETTA(t5x) | ${{ needs.build-rosetta-t5x.outputs.DOCKER_TAGS }} |
   #         # | ROSETTA(pax) | ${{ needs.build-rosetta-pax.outputs.DOCKER_TAGS }} |
 
-  upload-badge:
+
+
+  finalize:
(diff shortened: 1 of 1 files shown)


terrykong:
IIRC, the limit was on the [number of reusable workflows ](https://docs.github.com/en/actions/using-workflows/reusing-workflows#:~:text=You%20can%20call%20a%20maximum%20of%2020%20reusable%20workflows%20from%20a%20single%20workflow%20file.%20This%20limit%20includes%20any%20trees%20of%20nested%20reusable%20workflows%20that%20may%20be%20called%20starting%20from%20your%20top%2Dlevel%20caller%20workflow%20file.) and not to the number of jobs, so I don't think this would change the current reusable workflow count

yhtang (changes requested):
The limit is on the number of *workflows*, not the number of jobs in the workflow.
Please use separate jobs for each task instead of concatenating them 😄

DwarKapex:
Ah, OK, I see. Now I got it. Thanks for the clarifications.

PR #478, opened 2024-01-17
[NVIDIA] Update XLA-fp8 docs for ninja user guide

kaixih:
This PR adds two subsections on how to use the FP8 gemm and FP8 params directly.

cc @mingxu1067

### DIFF
--- a/rosetta/docs/NATIVE_FP8.md
+++ b/rosetta/docs/NATIVE_FP8.md
@@ -144,7 +144,7 @@ XLA will pattern match the above and rewrite it to FP8 matmul when:
 3. `(x)`: an arbitrary number of allowed ops (e.g., all-gather, copy, bitcast...), the full list of which can be found [here](https://github.com/openxla/xla/blob/e4fc3298eefa91702f86068af45340cad78d0335/xla/service/gpu/gemm_rewriter.cc#L259-L265).
 
 ### Gradient accumulation of FP8 params
-FP8 params, also known as OWG params (or FP8 meta), may be shared across different iterations of a loop in the context of pipeline parallelism. Consequently, the autograd system accumulates their gradients through addition operations. This is undesirable as addition is meaningless for FP8 params.
+FP8 params, also known as OWG params (or FP8 meta), may be shared across different iterations of a loop in the context of pipeline parallelism. During backpropagation, the autograd system accumulates their gradients from each iteration through the default addition operation. This is undesirable as addition is meaningless for FP8 params.
 
 To address this, we introduce a custom dtype wrapper `fm32` (means fp8 meta with the 32-bit physical size). It tells the autograd system to perform the max operation for gradient accumulation. This aligns with our expectations for FP8 params. The basic usage is demonstrated below:
 
@@ -154,15 +154,14 @@ def outer_fn(scale_f32, ...):
   scale_fm32 = jax.lax.convert_element_type(scale_f32, fp8_ops.fm32)
 
   def body_fn(carry, _):
-    # use scale_fm32; can temperarily convert it back to f32 for general
-    # math operations
+    # Can temperarily convert scale_fm32 back to f32 for general math operations
     return carry, None
 
   jax.lax.scan(body_fun, ..., length=3)
   return ...
 ```
 
-The main point is that we need to convert the FP8 params (e.g. the scale) from the original `f32` to `fm32` before launching the scan loop so that the autograd can apply the correct grad accumulation between loop iterations. Inside each iteration (i.e. `body_fn`), we can convert them from `fm32` to `f32` for general math operations (e.g. `mul`, `div`, etc.) and convert back to `fm32` at exit.
+In the example, we need to convert the FP8 params (e.g. the scale) from the original `f32` to `fm32` before launching the scan loop so that the autograd can apply the correct grad accumulation between loop iterations. Inside each iteration (i.e. `body_fn`), we can convert them from `fm32` to `f32` for general math operations (e.g. `mul`, `div`, etc.) and convert back to `fm32` at exit.
 
 ## Transformer Engine vs Native FP8 Support
 Native XLA-FP8 specifically targets matrix multiplication operations. In contrast, the Transformer Engine focuses on enhancing the overall performance of the entire transformer layer. This encompasses not only the FP8 matrix multiplication but also attention mechanisms, layer normalizations, and other components.
--- a/rosetta/docs/NATIVE_FP8.md
+++ b/rosetta/docs/NATIVE_FP8.md
@@ -128,7 +128,53 @@ python -m paxml.main \
 
 Please ensure you include the first two flags, `--xla_gpu_enable_reduction_epilogue_fusion=false` and `--xla_gpu_enable_triton_gemm=false`, as they are essential for enabling the FP8 functionality. The additional flags primarily focus on performance enhancement and should also prove beneficial for non-FP8 executions.
 
-### Transformer Engine vs Native FP8 Support
+## Guide for Ninja Users
+
+### How to trigger XLA to use FP8 MatMul
+The specific graph pattern that XLA supports for FP8 matmul is illustrated below:
+
+```
+convert -> multiply -> (x) -> dot
+  bcast ->
+```
+
+XLA will pattern match the above and rewrite it to FP8 matmul when:
+1. `convert`: input has to be FP8.
+2. `bcast`: input has to be a scalar.
+3. `(x)`: an arbitrary number of allowed ops (e.g., all-gather, copy, bitcast...), the full list of which can be found [here](https://github.com/openxla/xla/blob/e4fc3298eefa91702f86068af45340cad78d0335/xla/service/gpu/gemm_rewriter.cc#L259-L265).
+
+### How to ensure FP8 params accumulated correctly
+FP8 params, also known as OWG params, may undergo fan-outs (reuse) across different iterations within the context of pipeline parallelism. Consequently, the autograd system accumulates their gradients (hacked to store new values) through addition operations. This is undesirable as addition is meaningless for FP8 params. To address this, we introduce a custom dtype wrapper `fp8_ops.fm32` that performs the max operation for gradient accumulation. This aligns with our expectations for FP8 params. The basic usage is demonstrated below:
+
+```python
+def fwd(sf_fm32, ...):
+  # Convert fm32->f32 for mathematical operations                                         
+  sf = lax.convert_element_type(sf_fm32, jnp.float32)
+  
+  # Apply mathematical operations as before over sf
+
+  # Convert f32->fm32 back                                         
+  sf_fm32 = lax.convert_element_type(sf_fm32, fp8_ops.fm32)
+  return out, sf_fm32
+
+def bwd(res, ...):
+  return res, ...
+
+func_reusing_sf.defvjp(fwd, bwd)
+
+def launcher_func(sf_f32, ...):
+  # Convert f32->fm32 within the scope of jitting function
+  sf_fm32 = jax.lax.convert_element_type(sf_f32, fp8_ops.fm32)
+  func_reusing_sf(sf_fm32, ...)
+  return ...
+
+sf = jnp.array([1.], jnp.float32)
+jax.jit(launcher_func)(sf, ...)
+```
+
+The crucial point is that `fp8_ops.fm32` must be used within the scope of the jitted function. While we maintain the signature of the jit function, accepting fp32 inputs, internally, we convert them to `fp8_ops.fm32`. Then, the custom VJP functions operate with `fp8_ops.fm32` inputs and outputs, while performing conversions back and forth to fp32 to ensure that mathematical operations remain in the fp32 domain.
+
+## Transformer Engine vs Native FP8 Support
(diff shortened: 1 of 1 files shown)


rosetta/docs/NATIVE_FP8.md
```diff
@@ -129,7 +129,53 @@ python -m paxml.main \
...
+### How to trigger XLA to use FP8 MatMul
+The specific graph pattern that XLA supports for FP8 matmul is illustrated below:
+
+```
+convert -> multiply -> (x) -> dot
+  bcast ->
+```
+
+XLA will pattern match the above and rewrite it to FP8 matmul when:
+1. `convert`: input has to be FP8.
```
nouiz, line 143: what about output? bf16, f16 and f32 are supported? Maybe list the outputs supported.

rosetta/docs/NATIVE_FP8.md
```diff
@@ -129,7 +129,53 @@ python -m paxml.main \
...
+2. `bcast`: input has to be a scalar.
```
nouiz, line 144: Any dtype restriction?

rosetta/docs/NATIVE_FP8.md
```diff
@@ -129,7 +129,53 @@ python -m paxml.main \
...
+3. `(x)`: an arbitrary number of allowed ops (e.g., all-gather, copy, bitcast...), the full list of which can be found [here](https://github.com/openxla/xla/blob/e4fc3298eefa91702f86068af45340cad78d0335/xla/service/gpu/gemm_rewriter.cc#L259-L265).
+
+### How to ensure FP8 params accumulated correctly
+FP8 params, also known as OWG params, may undergo fan-outs (reuse) across different iterations within the context of pipeline parallelism. Consequently, the autograd system accumulates their gradients (hacked to store new values) through addition operations. This is undesirable as addition is meaningless for FP8 params. To address this, we introduce a custom dtype wrapper `fp8_ops.fm32` that performs the max operation for gradient accumulation. This aligns with our expectations for FP8 params. The basic usage is demonstrated below:
+
+```python
+def fwd(sf_fm32, ...):
```
nouiz, line 151: what fm32 mean? what sf mean?

rosetta/docs/NATIVE_FP8.md
```diff
@@ -129,7 +129,53 @@ python -m paxml.main \
...
+  # Convert f32->fm32 within the scope of jitting function
+  sf_fm32 = jax.lax.convert_element_type(sf_f32, fp8_ops.fm32)
+  func_reusing_sf(sf_fm32, ...)
+  return ...
+
+sf = jnp.array([1.], jnp.float32)
+jax.jit(launcher_func)(sf, ...)
+```
+
+The crucial point is that `fp8_ops.fm32` must be used within the scope of the jitted function. While we maintain the signature of the jit function, accepting fp32 inputs, internally, we convert them to `fp8_ops.fm32`. Then, the custom VJP functions operate with `fp8_ops.fm32` inputs and outputs, while performing conversions back and forth to fp32 to ensure that mathematical operations remain in the fp32 domain.
```
nouiz, line 176: All this seem complex.
Could we provide an fp8_gemm() function that wrap all that?
That way, customer could just call impot/copy and call it.

yhtang:
@nouiz Could you please submit your review as `request for change` so as to ensure the merge is blocked until all comments are resolved?

kaixih:
Sorry for the late response. Just made a couple of changes and PTAL and then we can forward to more people who are interested in the pipeline parallelism + fp8. @yhtang @nouiz

rosetta/docs/NATIVE_FP8.md
```diff
@@ -128,7 +128,42 @@ python -m paxml.main \
...
+    # Can temperarily convert scale_fm32 back to f32 for general math operations
+    return carry, None
+
+  jax.lax.scan(body_fun, ..., length=3)
+  return ...
+```
+
+In the example, we need to convert the FP8 params (e.g. the scale) from the original `f32` to `fm32` before launching the scan loop so that the autograd can apply the correct grad accumulation between loop iterations. Inside each iteration (i.e. `body_fn`), we can convert them from `fm32` to `f32` for general math operations (e.g. `mul`, `div`, etc.) and convert back to `fm32` at exit.
+
+## Transformer Engine vs Native FP8 Support
```
nouiz, line 166: Maybe keep that section before the new one?
It will help people know if they need to read the new section.

rosetta/docs/NATIVE_FP8.md
```diff
@@ -128,7 +128,42 @@ python -m paxml.main \
...
+
+```
+convert -> multiply -> (x) -> dot
+  bcast ->
+```
+
+XLA will pattern match the above and rewrite it to FP8 matmul when:
+1. `convert`: supports `f8` to [`bf16`|`f16`|`f32`].
+2. `bcast`: broadcasts a [`bf16`|`f16`|`f32`] scalar.
+3. `(x)`: an arbitrary number of allowed ops (e.g., all-gather, copy, bitcast...), the full list of which can be found [here](https://github.com/openxla/xla/blob/e4fc3298eefa91702f86068af45340cad78d0335/xla/service/gpu/gemm_rewriter.cc#L259-L265).
```
nouiz, line 144: Maybe add the full list? Only 6 elements. Much more easier then to go look at the URL. You can keep the url too, incase it is updated. But that is rare, so you can also remove it.

rosetta/docs/NATIVE_FP8.md
```diff
@@ -128,7 +128,42 @@ python -m paxml.main \
...
+The specific graph pattern that XLA supports for FP8 matmul is illustrated below:
+
+```
+convert -> multiply -> (x) -> dot
+  bcast ->
+```
+
+XLA will pattern match the above and rewrite it to FP8 matmul when:
+1. `convert`: supports `f8` to [`bf16`|`f16`|`f32`].
+2. `bcast`: broadcasts a [`bf16`|`f16`|`f32`] scalar.
```
nouiz, line 143: Does the dtype of 1 and 2 must be the same? Document in all cases if they are or not.

rosetta/docs/NATIVE_FP8.md
```diff
@@ -128,7 +128,42 @@ python -m paxml.main \
...
 
-### Transformer Engine vs Native FP8 Support
+## Guide for Ninja Users
+
+### Exact pattern that XLA can match for FP8 MatMul
+The specific graph pattern that XLA supports for FP8 matmul is illustrated below:
+
+```
+convert -> multiply -> (x) -> dot
+  bcast ->
```
nouiz, line 138: bcast -> broadcast

Is that optional? If so, clarify it.

rosetta/docs/NATIVE_FP8.md
```diff
@@ -128,7 +128,42 @@ python -m paxml.main \
...
+```
+
+XLA will pattern match the above and rewrite it to FP8 matmul when:
+1. `convert`: supports `f8` to [`bf16`|`f16`|`f32`].
+2. `bcast`: broadcasts a [`bf16`|`f16`|`f32`] scalar.
+3. `(x)`: an arbitrary number of allowed ops (e.g., all-gather, copy, bitcast...), the full list of which can be found [here](https://github.com/openxla/xla/blob/e4fc3298eefa91702f86068af45340cad78d0335/xla/service/gpu/gemm_rewriter.cc#L259-L265).
+
+### Gradient accumulation of FP8 params
+FP8 params, also known as OWG params (or FP8 meta), may be shared across different iterations of a loop in the context of pipeline parallelism. During backpropagation, the autograd system accumulates their gradients from each iteration through the default addition operation. This is undesirable as addition is meaningless for FP8 params.
```
nouiz, line 147: Add in parathesys what OWG mean?

rosetta/docs/NATIVE_FP8.md
```diff
@@ -128,7 +128,42 @@ python -m paxml.main \
...
+
+To address this, we introduce a custom dtype wrapper `fm32` (means fp8 meta with the 32-bit physical size). It tells the autograd system to perform the max operation for gradient accumulation. This aligns with our expectations for FP8 params. The basic usage is demonstrated below:
```
nouiz, line 149: fm32 doesn't seem a good choice. I can't relate it to any meaning even after hearing the explanation twice already.
I think fp8fp32 would be much clearer. Or fp8meta.

Or even better, fp32_max_grad?
That would make it clear what it is (an fp32 tensor) and what make it different to the normal fp32.

rosetta/docs/NATIVE_FP8.md
```diff
@@ -128,7 +128,42 @@ python -m paxml.main \
...
+def outer_fn(scale_f32, ...):
+  # Convert fp8 meta f32->fm32 before the scan_loop
+  scale_fm32 = jax.lax.convert_element_type(scale_f32, fp8_ops.fm32)
+
+  def body_fn(carry, _):
+    # Can temperarily convert scale_fm32 back to f32 for general math operations
+    return carry, None
+
+  jax.lax.scan(body_fun, ..., length=3)
+  return ...
```
nouiz, line 161: Can you make the example runable by copy-paster?

rosetta/docs/NATIVE_FP8.md
```diff
@@ -128,7 +128,61 @@ python -m paxml.main \
...
+### Exact pattern that XLA can match for FP8 MatMul
+The specific graph pattern that XLA supports for FP8 matmul is illustrated below:
+
+```
+  convert -> multiply -> (x) -> dot
+broadcast ->
+
+# or
+
+  convert -> (x) -> dot
```
nouiz, line 142: Can you confirm we don't fuse anything else, like the bias addition?

rosetta/docs/NATIVE_FP8.md
```diff
@@ -128,7 +128,61 @@ python -m paxml.main \
...
+FP8 params, also known as `OverwriteWithGrad` params (or FP8 meta), may be shared across different iterations of a loop in the context of pipeline parallelism. During backpropagation, the autograd system accumulates their gradients from each iteration through the default addition operation. This is undesirable as addition is meaningless for FP8 params.
+
+To address this, we introduce a custom dtype wrapper `fp32_max_grad`. It tells the autograd system to perform the max operation for gradient accumulation. This aligns with our expectations for FP8 params. The basic usage is demonstrated below:
+
+```python
+from flax.linen import fp8_ops
+# ah and sf are short for amax history and scaling factor used as the FP8 params.
+f32 = jnp.float32
+fm32 = fp8_ops.fp32_max_grad
+def outer(x, ah_f32, sf_f32):
```
nouiz, line 165: Longer name than ah and sf? Or add a comment that tell what they mean?

rosetta/docs/NATIVE_FP8.md
```diff
@@ -128,7 +128,61 @@ python -m paxml.main \
...
+### Gradient accumulation of FP8 params
+FP8 params, also known as `OverwriteWithGrad` params (or FP8 meta), may be shared across different iterations of a loop in the context of pipeline parallelism. During backpropagation, the autograd system accumulates their gradients from each iteration through the default addition operation. This is undesirable as addition is meaningless for FP8 params.
+
+To address this, we introduce a custom dtype wrapper `fp32_max_grad`. It tells the autograd system to perform the max operation for gradient accumulation. This aligns with our expectations for FP8 params. The basic usage is demonstrated below:
+
+```python
+from flax.linen import fp8_ops
+# ah and sf are short for amax history and scaling factor used as the FP8 params.
+f32 = jnp.float32
+fm32 = fp8_ops.fp32_max_grad
```
nouiz, line 164: Did you test it?
To my knowledge, we ned to update flax, it probably use fm32.
At least, in the example, you shouldn't use fm32. Maybe fmax32 if you want something shorter?

rosetta/docs/NATIVE_FP8.md
kaixih: I think bias and other epilog ops work same with the dot of the higher precision (e.g. bf16, fp32). We do support epilog quantize but it is complex and deserve a special section.

rosetta/docs/NATIVE_FP8.md
kaixih: Yes, it is already in [here](https://github.com/google/flax/blob/144486b5fa7b3dfb1f3ecb58f708e5e9f2f66175/tests/linen/linen_test.py#L1393).

And the PR will be like https://github.com/google/flax/compare/main...kaixih:flax:rename_fp8_custom_dtype?expand=1.

rosetta/docs/NATIVE_FP8.md
kaixih: And on a second thought, I don't mention it here is the bias (or other epilog) won't stop us using the fp8. I mean XLA will do a good job on whether to fuse the epilog to dot, like if they have consumers/ spmd insertions. Users shouldn't worry too much on it. But for the input quantize, users need to be careful to follow the above pattern, otherwise they will not use fp8 gemm.

nouiz: It can be in another PR. But having an example that show the bias fusion to have at least one example with epilog fusion would be great.

kaixih: Done.

rosetta/docs/NATIVE_FP8.md
kaixih: Done.

nouiz (approved):

PR #483, opened 2024-01-17
Add documentation and scripts for Pax benchmarking using synthetic dataset

ashors1:
Note that this PR bumps the manifest. This is needed because `test-pax.sh` has been modified to use the synthetic config from upstream. When the manifest is not bumped, `test-pax.sh` will fail because the synthetic config is not present in upstream Pax.


ashors1:
This PR depends on the following paxml PR: https://github.com/google/paxml/pull/61

terrykong (approved):
LGTM

terrykong:
Could you update the description of the PR to add that you are bumping the manifest because it's required and why it is required?

terrykong (dismissed):

yhtang (changes requested):
Could you please check if `manifest.yaml` content is indeed the latest? #487 seems to also updated it.

ashors1:
> Could you please check if `manifest.yaml` content is indeed the latest? #487 seems to also updated it.

Yep, verified that `manifest.yaml` is indeed updated. #487 needed to update the manifest to include an XLA version with the necessary flag. This further updates the manifest to avoid a temporary TE performance degradation we observed.

ashors1:
The presubmit CI is showing that the TE tests are still crashing an error that looks related to the LLVM issue we have been seeing, but I don't think we should let that crash block this PR. I'll look into that issue independently, but I think we should merge this PR in the meantime to unblock our nightly tests. Thoughts @yhtang @terrykong ?

ashors1:
Also, there appears to be a regression with the single GPU TE test. This I will look into as well, but as long as this PR is not merged, our nightly tests will continue to fail.

terrykong:
I'm seeing that there are regressions in multiple rosetta-pax tests. I see in one test:
```
jaxlib.xla_extension.XlaRuntimeError: UNKNOWN: <unknown>:0: error: loc("reduce.3890"): 'mhlo.reduce' op inferred type(s) 'tensor<3072xbf16>' are incompatible with return type(s) of operation 'tensor<3072xf32>'
<unknown>:0: error: loc("reduce.3890"): 'mhlo.reduce' op failed to infer returned types
<unknown>:0: note: loc("reduce.3890"): see current operation: 
%5154 = "mhlo.reduce"(%5151, %5153) ({
^bb0(%arg641: tensor<bf16>, %arg642: tensor<bf16>):
  %5156 = "mhlo.add"(%arg641, %arg642) : (tensor<bf16>, tensor<bf16>) -> tensor<bf16>
  "mhlo.return"(%5156) : (tensor<bf16>) -> ()
}) {dimensions = dense<1> : tensor<1xi64>} : (tensor<3072x8192xf32>, tensor<f32>) -> tensor<3072xf32>
```
And this looks new, can you see if selectively updating the pax manifest will fix this? In particular, revert all the SHAs except praxis/paxml?

I think we should merge this since it fixes tests, but I'm trying to understand how we could improve our process next time around. @ashors1 If in #487, the manifest was selectively updated so that only JAX/XLA was bumped, but everything else wasn't; assuming there wasn't a JAX API change affecting Pax, could that have avoided us introducing a broken state?

ashors1:
> I'm seeing that there are regressions in multiple rosetta-pax tests. I see in one test

Yes, that is the new failure we are seeing as a result of the proposed WAR `--xla_gpu_enable_llvm_module_compilation_parallelism=true`. I will try reverting all the SHAs besides paxml and praxis.

> If in https://github.com/NVIDIA/JAX-Toolbox/pull/487, the manifest was selectively updated so that only JAX/XLA was bumped, but everything else wasn't; assuming there wasn't a JAX API change affecting Pax, could that have avoided us introducing a broken state?

 I can't say for sure, but my current understanding is that the flag that was supposed to workaround the LLVM issue did not work as expected and introduced this error. If that is true, updating only JAX/XLA would not have avoided the failures.

yhtang (approved):

PR #484, opened 2024-01-18
[Draft] Make the t5x container work on Aarch64 (v3)

gspschmid:
This expands upon @nouiz's [earlier PR](https://github.com/NVIDIA/JAX-Toolbox/pull/421) aimed at getting an arm64 build for T5X.

I took a stab at resolving the remaining issues by building both `grain` and its dependency, `array_record`, from source.

Building the new dockerfile with
```bash
$ docker buildx build -f Dockerfile.t5x.arm64 --tag t5x --build-arg BASE_IMAGE=ghcr.io/nvidia/jax:mealkit-2023-12-12 .
```
gives me an image in which the import succeeds:
```bash
$ docker run --rm -it --shm-size=1g t5x python3 -c 'import t5x; print(t5x)'
(...)
<module 't5x' from '/opt/paxml/t5x/__init__.py'>
```
🥳 
---

The underlying issue was that both `array_record` and `grain` were at some point erroneously published as universal (arch-independent) wheels. For `grain` this is still the case, for `array_record` this was only the case until 0.4.1, whereas from 0.5.0 only an `x86_64` wheel was published, meaning pip ends up pulling the older (arm64-incompatible) wheel anyways.

The `array_record` build is essentially unchanged and replicates that repo's Dockerfile. I initially tried to use "mealkit" as a base image, but at least `array_record` seems to make some strong assumptions on where exactly the compiler toolchain is located.

For `grain` to build I had to make [some pretty heavy-handed changes](https://github.com/gspschmid/grain/commits/external-arm64-build-poc/) -- it seems like the export process from Google's internal monorepo through copybara left various parts mangled. Among other things, the bazel build currently refers to some Google-internal dependencies, and in two places the Python code just isn't syntactically valid ([#1](https://github.com/google/grain/blob/6de2270cec0407d0011721b78461b344abe07704/grain/_src/python/data_sources.py#L45-L47), [#2](https://github.com/google/grain/blob/6de2270cec0407d0011721b78461b344abe07704/grain/_src/python/data_loader.py#L181-L191)).

For both repos I'm hoping we can get the respective maintainers to take a look and eventually publish some arm64-specific wheels :-)

### DIFF
--- a/.github/container/Dockerfile.t5x.arm64
+++ b/.github/container/Dockerfile.t5x.arm64
@@ -11,13 +11,25 @@ ARG SRC_PATH_T5X=/opt/t5x
 ###############################################################################
 
 #------------------------------------------------------------------------------
+# Install bazel (needed by some source builds) and expose the python version
+#------------------------------------------------------------------------------
+
+FROM ${BASE_IMAGE} as wheel-builder
+
+RUN wget https://github.com/bazelbuild/bazelisk/releases/download/v1.17.0/bazelisk-linux-arm64 -O /usr/bin/bazel && \
+    chmod a+x /usr/bin/bazel
+
+RUN python -c 'import sys; vi = sys.version_info; print(f"{vi.major}.{vi.minor}")'  > /mealkit-python-version
+
+
+#------------------------------------------------------------------------------
 # build array_record 0.5.0 from source
 #------------------------------------------------------------------------------
 # TODO: Remove this once array_record maintainers have published an arm64 wheel.
 
 FROM linaro/tensorflow-arm64-build:2.12-multipython as array_record-builder
-# TODO: Infer PYTHON_VERSION from our own $BASE_IMAGE?
-ARG PYTHON_VERSION="3.10"
+
+COPY --from=wheel-builder /mealkit-python-version /mealkit-python-version
 
 # The bazel build of array_record makes some strong assumptions on where the cpp toolchain lives.
 # We therefore replicate (part of) its Dockerfile, which ensures everything is in the right place:
@@ -25,10 +37,11 @@ ARG PYTHON_VERSION="3.10"
 #
 # BEGIN
     ENV DEBIAN_FRONTEND=noninteractive
-    ENV PYTHON_BIN_PATH=/usr/bin/python${PYTHON_VERSION}
+    ENV PYTHON_BIN_PATH=/usr/bin/python${MEALKIT_PYTHON_VERSION}
 
     # Install supplementary Python interpreters
-    RUN ln -s ${PYTHON_BIN_PATH} /usr/local/bin/python && \
+    RUN export PYTHON_BIN_PATH=/usr/bin/python$(cat /mealkit-python-version) && \
+        ln -s ${PYTHON_BIN_PATH} /usr/local/bin/python && \
         ln -s ${PYTHON_BIN_PATH} /usr/local/bin/python3 && \
         ln -s ${PYTHON_BIN_PATH} /usr/bin/python
 
@@ -40,8 +53,8 @@ ARG PYTHON_VERSION="3.10"
             libffi-dev
 
     # Install pip dependencies needed for array_record
-    RUN ${PYTHON_BIN_PATH} -m pip install -U pip && \
-        ${PYTHON_BIN_PATH} -m pip install -U \
+    RUN /usr/bin/python -m pip install -U pip && \
+        /usr/bin/python -m pip install -U \
             absl-py \
             auditwheel \
             etils[epath] \
@@ -65,23 +78,11 @@ EOT
 
 
 #------------------------------------------------------------------------------
-# intermediate stage that adds bazel (needed by both tftext and grain)
-#------------------------------------------------------------------------------
-
-ARG BASE_IMAGE
-FROM ${BASE_IMAGE} as wheel-builder
-
-RUN wget https://github.com/bazelbuild/bazelisk/releases/download/v1.17.0/bazelisk-linux-arm64 -O /usr/bin/bazel && \
-    chmod a+x /usr/bin/bazel
-
-
-#------------------------------------------------------------------------------
 # build grain from a manually-patched version
 #------------------------------------------------------------------------------
 # TODO: Remove this once grain maintainers have published an arm64 wheel.
 
 FROM wheel-builder as grain-builder
-ARG PYTHON_VERSION="3.10"
 
 # Setup bazel
 RUN  wget https://github.com/bazelbuild/bazelisk/releases/download/v1.17.0/bazelisk-linux-arm64 -O /usr/bin/bazel && \
@@ -92,11 +93,12 @@ RUN  wget https://github.com/bazelbuild/bazelisk/releases/download/v1.17.0/bazel
 #
 # BEGIN
     # Setup python
-    RUN apt-get update && apt-get install -y \
+    RUN export MEALKIT_PYTHON_VERSION=$(cat /mealkit-python-version) && \
+        apt-get update && apt-get install -y \
             python3-dev python3-pip python3-venv && \
             rm -rf /var/lib/apt/lists/* && \
-            python${PYTHON_VERSION} -m pip install pip --upgrade && \
-            update-alternatives --install /usr/bin/python python /usr/bin/python${PYTHON_VERSION} 0
+            python${MEALKIT_PYTHON_VERSION} -m pip install pip --upgrade && \
+            update-alternatives --install /usr/bin/python python /usr/bin/python${MEALKIT_PYTHON_VERSION} 0
 
     # Install pip dependencies needed for grain
     # NOTE: We are ignoring the fact that "array_record" will actually resolve to the wrong version
@@ -132,6 +134,7 @@ git clone https://github.com/gspschmid/grain /opt/grain
 cd /opt/grain
 git checkout external-arm64-build-poc
 
+export PYTHON_VERSION=$(cat /mealkit-python-version)
 chmod a+x ./grain/oss/build_whl.sh
 ./grain/oss/build_whl.sh
 
--- a/.github/container/Dockerfile.t5x.arm64
+++ b/.github/container/Dockerfile.t5x.arm64
@@ -3,10 +3,11 @@
 #   docker buildx build -f Dockerfile.t5x.arm64 --tag t5x --build-arg BASE_IMAGE=ghcr.io/nvidia/jax:mealkit-2024-01-22 .
 
 ARG BASE_IMAGE=ghcr.io/nvidia/jax:mealkit
+ARG SRC_PATH_TFTEXT=/opt/tensorflow-text
 ARG SRC_PATH_T5X=/opt/t5x
 
 ###############################################################################
-## build array_record and grain which do not have working arm64 pip wheels
+## build several packages which do not have working arm64 pip wheels
 ###############################################################################
 
 #------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


gspschmid:
cc @yhtang

.github/container/Dockerfile.t5x.arm64
```diff
@@ -0,0 +1,226 @@
...
+#------------------------------------------------------------------------------
+# TODO: Remove this once array_record maintainers have published an arm64 wheel.
+
+# The bazel build of array_record makes some strong assumptions on where the cpp toolchain lives.
+# We therefore stick to its existing Dockerfile, which ensures everything is in the right place:
+#
+#   https://github.com/google/array_record/tree/v0.5.0/oss/build.Dockerfile.aarch64
+#
+FROM linaro/tensorflow-arm64-build:2.12-multipython as array_record-builder
+# TODO: Infer PYTHON_VERSION from our own $BASE_IMAGE?
```
yhtang, line 32: Yes please 😄

.github/container/Dockerfile.t5x.arm64
```diff
@@ -0,0 +1,226 @@
...
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV PYTHON_BIN_PATH=/usr/bin/python${PYTHON_VERSION}
+
+# Install supplementary Python interpreters
+RUN ln -s ${PYTHON_BIN_PATH} /usr/local/bin/python && \
+    ln -s ${PYTHON_BIN_PATH} /usr/local/bin/python3 && \
+    ln -s ${PYTHON_BIN_PATH} /usr/bin/python
+
+RUN --mount=type=cache,target=/var/cache/apt \
```
yhtang, line 43: Our CI runners always start in a fresh pod so it is unlikely that this cache can help.

.github/container/Dockerfile.t5x.arm64
```diff
@@ -0,0 +1,226 @@
...
+RUN wget https://github.com/bazelbuild/bazelisk/releases/download/v1.17.0/bazelisk-linux-arm64 -O /usr/bin/bazel && \
+    chmod a+x /usr/bin/bazel
+
+#------------------------------------------------------------------------------
+# build array_record 0.5.0 from source
+#------------------------------------------------------------------------------
+# TODO: Remove this once array_record maintainers have published an arm64 wheel.
+
+# The bazel build of array_record makes some strong assumptions on where the cpp toolchain lives.
+# We therefore stick to its existing Dockerfile, which ensures everything is in the right place:
```
yhtang, line 27: Can we add a bit more comments to delimit the beginning and end of this embedded dockerfile?

.github/container/Dockerfile.t5x.arm64
```diff
@@ -0,0 +1,226 @@
...
+    etils[epath] \
+    "more-itertools>=9.1.0" \
+    numpy;
+
+pip install \
+    dill \
+    jax \
+    jaxlib \
+    tensorflow \
+    tensorflow-datasets;
```
yhtang, line 112: Why do we need the trailing semicolon?

.github/container/Dockerfile.t5x.arm64
```diff
@@ -0,0 +1,226 @@
+# syntax=docker/dockerfile:1-labs
+# Example command to build manually:
+#   docker buildx build -f Dockerfile.t5x.arm64 --tag t5x --build-arg BASE_IMAGE=ghcr.io/nvidia/jax:mealkit-2023-12-12 .
+
+ARG BASE_IMAGE=ghcr.io/nvidia/jax:mealkit
+ARG REPO_T5X=https://github.com/google-research/t5x.git
```
yhtang, line 6: The new x86 Dockerfile uses the `manifest.yaml` file instead of these build args. Could you please update this Dockerfile to also use that?

.github/container/Dockerfile.t5x.arm64
```diff
@@ -0,0 +1,226 @@
...
+
+COPY --from=grain-builder /tmp/grain/all_dist/grain*.whl /opt/
+RUN echo "grain @ file://$(ls /opt/grain*.whl)" >> /opt/pip-tools.d/manifest.pax
+
+
+### --- TO BE REFACTORED? (everything below is as in @nouiz's original PR) ---
+
+# force a recent version to have latest protobuf dep
+# Install now as some build need them.
+RUN pip install "tensorflow_datasets==4.9.2"
```
yhtang, line 146: In the production Dockerfile, we should not pip install packages in separate steps.
The purpose of these installs seem to provide the dependencies for building tftext. Could you please port the tftext builder section from the arm64 paxml Dockerfile to this PR?

.github/container/Dockerfile.t5x.arm64
```diff
@@ -0,0 +1,226 @@
...
+    mv /root/.profile /root/.profile.save && \
+    grep -v cargo /root/.profile.save > /root/.profile && \
+    rm /root/.profile.save && \
+    mv /root/.bashrc /root/.bashrc.save && \
+    grep -v cargo /root/.bashrc.save > /root/.bashrc && \
+    rm /root/.bashrc.save && \
+        rm -Rf /root/.cache /tmp/*
+
+# NOTE(gspschmid): Broken? Should've been "echo t5"?
+# RUN "t5 >> /opt/pip-tools.d/manifest.t5x"
```
yhtang, line 185: This does look broken.

.github/container/Dockerfile.t5x.arm64
```diff
@@ -0,0 +1,226 @@
...
+
+./oss_scripts/run_build.sh
+find * | grep '.whl$'
+EOF
+RUN pip install /opt/text/tensorflow_text-*.whl
+#RUN echo "-e file:///opt/text/tensorflow_text-*.whl"  >> /opt/pip-tools.d/manifest.t5x
+RUN echo "tensorflow-text @  file://$(ls /opt/text/tensorflow_text-*.whl)" >> /opt/pip-tools.d/manifest.t5x
+
+# Install T5 now, Pip will build the wheel from source, it needs Rust.
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > /tmp/rustup.sh && \
```
yhtang, line 170: Could you please validate if Rust is still needed as of now? Can we install that  in a `builder` stage too and only copy out the wheel?

.github/container/Dockerfile.t5x.arm64
```diff
@@ -0,0 +1,226 @@
...
+
+ADD test-t5x.sh /usr/local/bin
+
+###############################################################################
+## Install accumulated packages from the base image and the previous stage
+###############################################################################
+
+FROM mealkit as final
+
+RUN pip uninstall -y tensorflow-text
```
yhtang, line 222: This won't be needed if we build it in a separate stage and only copy the wheel into the mealkit.

.github/container/Dockerfile.t5x.arm64
```diff
@@ -0,0 +1,226 @@
...
+
+# remove head-of-tree specs from select dependencies
+pushd ${SRC_PATH_T5X}
+sed -i "s| @ git+https://github.com/google/flax#egg=flax||g" setup.py
+# for ARM64 build
+sed -i "s/tensorflow/#tensorflow/" setup.py
+sed -i "s/t5=/#t5=/" setup.py
+sed -i "s/^jax/#jax/" setup.py
+
+sed -i "s/f'jax/#f'jax/" setup.py
```
yhtang, line 199: Are these two items above redundant?

yhtang (changes requested):

.github/container/Dockerfile.t5x.arm64
```diff
@@ -0,0 +1,226 @@
...
+ARG REPO_T5X=https://github.com/google-research/t5x.git
+ARG REF_T5X=main
+ARG SRC_PATH_T5X=/opt/paxml
+
+###############################################################################
+## build array_record and grain which do not have working arm64 pip wheels
+###############################################################################
+
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE} as wheel-builder
```
yhtang, line 15: Can we move the array_record build stage forward, before the grain build stage? I feel that gives better logical connection, as the two stages are directly related. On the contrary, the grain builder is internalized from another dockerfile and does not inherit anything from `wheel-builder`.

gspschmid: It's the other way around -- `grain-builder` can at least reuse the base image, whereas the `array_record` really requires on a completely different base image.

I now put the `array_record-builder` stage first and combined `wheel-builder` and `grain-builder` into one stage.

.github/container/Dockerfile.t5x.arm64
gspschmid: Seems to work without -- which is nice because we can elide the explicit `pip install t5` entirely now.

.github/container/Dockerfile.t5x.arm64
yhtang: A possible solution is to write the Python version number to a file in the wheel-builder stage, and then use `COPY --from=wheel-builder ...` to copy this information to the new stage.

gspschmid:
cc @olupton 

This should be ready for a (final?) review.

Actually passing the mealkit python version around was surprisingly painful. There doesn't seem to be any straightforward way to programatically declare an env variable in a container, unless all of the scripts go into a single `RUN` invocation. Appending to `~/.profile` doesn't work within the `array_record` builder stage, and I couldn't figure out why. In any case, I think the current version is reasonable clear and concise.

.github/container/Dockerfile.t5x.arm64
```diff
@@ -0,0 +1,212 @@
...
+
+COPY --from=wheel-builder /mealkit-python-version /mealkit-python-version
+
+# The bazel build of array_record makes some strong assumptions on where the cpp toolchain lives.
+# We therefore replicate (part of) its Dockerfile, which ensures everything is in the right place:
+#   https://github.com/google/array_record/tree/v0.5.0/oss/build.Dockerfile.aarch64
+#
+# BEGIN
+    ENV DEBIAN_FRONTEND=noninteractive
+    ENV PYTHON_BIN_PATH=/usr/bin/python${MEALKIT_PYTHON_VERSION}
```
olupton, line 40:
```suggestion
```

.github/container/Dockerfile.t5x.arm64
```diff
@@ -0,0 +1,212 @@
...
+
+# The following setup steps are based on
+#   https://github.com/google/grain/blob/ab557c31672cc2ac6f9b31e33aea2df00a7da5bf/grain/oss/build.Dockerfile
+#
+# BEGIN
+    # Setup python
+    RUN export MEALKIT_PYTHON_VERSION=$(cat /mealkit-python-version) && \
+        apt-get update && apt-get install -y \
+            python3-dev python3-pip python3-venv && \
+            rm -rf /var/lib/apt/lists/* && \
```
olupton, line 99: Any particular reason you have this in some places but not all?
I don't think it matters as these are just in the wheel-building steps, which we ultimately don't use any layers from.

olupton (approved):
LGTM, one small suggestion

.github/container/Dockerfile.t5x.arm64
gspschmid: This just replicates a step in https://github.com/google/grain/blob/ab557c31672cc2ac6f9b31e33aea2df00a7da5bf/grain/oss/build.Dockerfile#L34-L39 -- I wouldn't be surprised if we could tweak/minimize this given what's already installed in the mealkit base image, but out of convenience for staying in sync with the grain repo I decided to include it here.

.github/container/Dockerfile.t5x.arm64
gspschmid: Thanks!

gspschmid:
Some auth issue with uploading images from the CI job?
https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7625744699/job/20773413645?pr=484#step:8:1800

gspschmid:
Closing in favor of https://github.com/NVIDIA/JAX-Toolbox/pull/492, which is the same PR, but as a branch on this the nv org's repo. This resolves some authentication issues in CI jobs.

PR #486, opened 2024-01-19
Install newer nsight-compute and nsight-systems-cli

olupton:
Take the newest version available in https://developer.download.nvidia.com/devtools/repos.

### DIFF
--- a/.github/container/Dockerfile.base
+++ b/.github/container/Dockerfile.base
@@ -24,8 +24,10 @@ apt-get update
 apt-get install -y curl
 
 # Repo for newer nsight versions
-curl -o /usr/share/keyrings/nvidia.pub https://developer.download.nvidia.com/devtools/repos/ubuntu2204/$(dpkg --print-architecture)/nvidia.pub
-echo "deb [signed-by=/usr/share/keyrings/nvidia.pub] https://developer.download.nvidia.com/devtools/repos/ubuntu2204/$(dpkg --print-architecture)/ /" > /etc/apt/sources.list.d/devtools-ubuntu2204-$(dpkg --print-architecture).list
+export UBUNTU_ARCH=$(dpkg --print-architecture)
+export UBUNTU_VERSION=$(. /etc/os-release && echo ${ID}${VERSION_ID/./}) # e.g. ubuntu2204
+curl -o /usr/share/keyrings/nvidia.pub https://developer.download.nvidia.com/devtools/repos/${UBUNTU_VERSION}/${UBUNTU_ARCH}/nvidia.pub
+echo "deb [signed-by=/usr/share/keyrings/nvidia.pub] https://developer.download.nvidia.com/devtools/repos/${UBUNTU_VERSION}/${UBUNTU_ARCH}/ /" > /etc/apt/sources.list.d/devtools-${UBUNTU_VERSION}-${UBUNTU_ARCH}.list
 apt-get update
 
 apt-get install -y \
--- a/.github/container/Dockerfile.base
+++ b/.github/container/Dockerfile.base
@@ -40,6 +40,7 @@ apt-get install -y \
   gnupg \
   liblzma-dev \
   lld \
+  nsight-compute \
   nsight-systems-cli \
   python-is-python3 \
   python3-pip \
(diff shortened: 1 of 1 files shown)


.github/container/Dockerfile.base
```diff
@@ -17,30 +17,37 @@ ARG DEST_MANIFEST_DIR
...
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
+RUN <<"EOF" bash -ex
+export DEBIAN_FRONTEND=noninteractive
+export TZ=America/Los_Angeles
+apt-get update
+apt-get install -y curl
+
+# Repo for newer nsight versions
+curl -o /usr/share/keyrings/nvidia.pub https://developer.download.nvidia.com/devtools/repos/ubuntu2204/$(dpkg --print-architecture)/nvidia.pub
```
terrykong, line 27: nit: I know docker build has some [automatically populated build args](https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope), do you know if any of these can replace the `dpkg` invocation?

yhtang: `TARGETARCH` apparently does. Empirically we do know it's `amd64` and `arm64` on two architectures that we care about. However, I'm not able to find a formal specification of the values that it can take.

.github/container/Dockerfile.base
```diff
@@ -17,30 +17,37 @@ ARG DEST_MANIFEST_DIR
...
+echo "deb [signed-by=/usr/share/keyrings/nvidia.pub] https://developer.download.nvidia.com/devtools/repos/ubuntu2204/$(dpkg --print-architecture)/ /" > /etc/apt/sources.list.d/devtools-ubuntu2204-$(dpkg --print-architecture).list
```
yhtang, line 28: Can we parameterize the `2204` part so that we don't have to manually update it when bumping the base OS version?
That'll also be more friendly if we want to matricize the build over other Ubuntu versions.

.github/container/Dockerfile.base
```diff
@@ -17,30 +17,37 @@ ARG DEST_MANIFEST_DIR
...
+  bat \
+  build-essential \
+  checkinstall \
+  clang \
+  cmake \
+  git \
+  gnupg \
+  liblzma-dev \
+  lld \
+  nsight-systems-cli \
```
yhtang, line 41: Could you also please check if nsight-compute is installed?

.github/container/Dockerfile.base
```diff
@@ -93,12 +100,6 @@ ADD install-efa.sh /usr/local/bin
 ENV LD_LIBRARY_PATH=/opt/amazon/efa/lib:${LD_LIBRARY_PATH}
 ENV PATH=/opt/amazon/efa/bin:${PATH}
 
-###############################################################################
-## Emergency fix: nsys not in PATH
-###############################################################################
-
-RUN ln -s /opt/nvidia/nsight-compute/*/host/target-linux-x64/nsys /usr/local/cuda/bin
```
yhtang, line 100: Is this version of ncu still in the base image?

yhtang (changes requested):

.github/container/Dockerfile.base
olupton: IMO it's better to get a string from the package management machinery if it's being fed into the package management machinery, given that it's not clear whether the conventions match.

.github/container/Dockerfile.base
olupton: Done.

.github/container/Dockerfile.base
olupton: The base image contains both ncu and nsys, at versions older than the newest available in the devtools repository.

.github/container/Dockerfile.base
olupton: The base image ships a version of nsight-compute as well as a version of nsight-systems.
I've added nsight-compute to the list of packages here now, so we'll get the newest version.
Unlike nsight-systems, there is no specific motivation for picking up the slightly newer version at the moment.

yhtang (approved):

PR #494, opened 2024-01-23
Add fix for XLA custom sharding lowering bug

ashors1:


nouiz:
We already have a WAR. Why replace it with another WAR instead of waiting for it to be fixed upstream?

ashors1:
> We already have a WAR. Why replace it with another WAR instead of waiting for it to be fixed upstream?

The XLA flag that was provided as a WAR is not working as expected. With `--xla_gpu_enable_llvm_module_compilation_parallelism=true`, we are now seeing errors that look like this: 
```
jaxlib.xla_extension.XlaRuntimeError: UNKNOWN: <unknown>:0: error: loc("reduce.3890"): 'mhlo.reduce' op inferred type(s) 'tensor<3072xbf16>' are incompatible with return type(s) of operation 'tensor<3072xf32>'
<unknown>:0: error: loc("reduce.3890"): 'mhlo.reduce' op failed to infer returned types
<unknown>:0: note: loc("reduce.3890"): see current operation: 
%5154 = "mhlo.reduce"(%5151, %5153) ({
^bb0(%arg641: tensor<bf16>, %arg642: tensor<bf16>):
  %5156 = "mhlo.add"(%arg641, %arg642) : (tensor<bf16>, tensor<bf16>) -> tensor<bf16>
  "mhlo.return"(%5156) : (tensor<bf16>) -> ()
}) {dimensions = dense<1> : tensor<1xi64>} : (tensor<3072x8192xf32>, tensor<f32>) -> tensor<3072xf32>
```

terrykong:
@ashors1 Once this is merged, can you open a PR that reverts this to emphasize this is temporary?

It looks like pax tests are now functional (yay), but there is a perf regression and we will file a bug internally. I think having this fix in ASAP to have a meaningful signal from the nightly CI tests is important. Any opinions @yhtang @nouiz ?

terrykong (approved):
LGTM. I think in the future it may a good idea to move these things into the manifest and use `create_distribution.sh` to form the xla distribution for hotfixes like this.

ashors1:
Draft PR to revert this change: https://github.com/NVIDIA/JAX-Toolbox/pull/495

PR #477, opened 2024-01-16
Revert "Adds a temporary WAR to the jax container that fixes a crash observed in T5x/Pax"

yhtang:
**This is a reminder PR to be merged once the underlying XLA issue is fixed**

Reverts NVIDIA/JAX-Toolbox#472

PR #495, opened 2024-01-24
Revert custom sharding WAR

ashors1:


yhtang:
Is this just a reminder like #477?

ashors1:
> Is this just a reminder like #477?

Yes

jaro-sevcik:
The WAR is not needed anymore since a proper fix landed: https://github.com/openxla/xla/pull/8896

olupton:
Workaround should no longer be needed following https://github.com/openxla/xla/pull/8896. @jaro-sevcik checked this morning.

olupton (approved):
LGTM, merging before nightlies

.github/container/build-jax.sh
```diff
@@ -133,10 +133,6 @@ while [ : ]; do
             XLA_ARM64_PATCH_LIST=$2
             shift 2
             ;;
-        --other-xla-patches)
```
nouiz, line 136: Keeping that attribute could be useful in the future.
We can always add it back later, but having it for dev could be useful.

PR #498, opened 2024-01-25
Restores markdown table summarizing results from upstream MGMN tests that were removed in #440.

terrykong:
Instead of restoring it in .summary, it is put into .full_result_markdown so that summary can be kept concise


terrykong:
Affected workflows:
* pax: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7661372679
* t5x: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7661373840

ashors1 (approved):

PR #499, opened 2024-01-26
Use standard grain source build in t5x arm64 Dockerfile

gspschmid:
Thanks to @iindyk's recent fix in the copybara export of `grain` (see [this commit](https://github.com/google/grain/commit/b02754da440d7271c3ea687f2d3fe20c672ea91b)) we no longer have to apply an intrusive patch in order to build grain from source as part of our T5X arm64 build.

A small issue remains, since some of the tests rely on missing dependencies at runtime, so we disable those two (see [these logs](https://gist.github.com/gspschmid/edddde8bb9d807b0149ed9b7bf43d3d8)).

cc @yhtang @olupton

### DIFF
--- a/.github/container/Dockerfile.t5x.arm64
+++ b/.github/container/Dockerfile.t5x.arm64
@@ -3,6 +3,7 @@
 #   docker buildx build -f Dockerfile.t5x.arm64 --tag t5x --build-arg BASE_IMAGE=ghcr.io/nvidia/jax:mealkit-2024-01-22 .
 
 ARG BASE_IMAGE=ghcr.io/nvidia/jax:mealkit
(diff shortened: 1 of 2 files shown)


gspschmid:
Rosetta-PAX tests are red due to Pallas tests on A100 failing: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7671181458/job/20910215878?pr=499#step:8:29

gspschmid:
Given that the rosetta test failures seem unrelated (also happen on other PRs, also on x86 and afaict ultimately appear related to Pallas) I think we should be fine to merge this?

olupton (approved):
LGTM, will be added to the CI in #500.

PR #500, opened 2024-01-26
Test and publish T5X arm64 build in CI

gspschmid:
So far, T5X was only tested and published for amd64. This also adds the arm64 build, analogously to how we already do it for PAX.

cc @yhtang @nouiz


gspschmid:
Hmm, seems like the rosetta build broke -- perhaps due to my addition of `grain` to `manifest.yaml`? Will have a look.

.github/workflows/_ci.yaml
```diff
@@ -55,7 +54,7 @@ jobs:
       BUILD_DATE: ${{ needs.metadata.outputs.BUILD_DATE }}
       BASE_IMAGE: ${{ needs.build-jax.outputs.DOCKER_TAG_MEALKIT }}
       CONTAINER_NAME: upstream-t5x
-      DOCKERFILE: .github/container/Dockerfile.t5x
+      DOCKERFILE: .github/container/Dockerfile.pax.${{ inputs.ARCHITECTURE }}
```
terrykong, line 57: This says pax instead of t5x

gspschmid: Indeed, that and another line in `nightly-t5x-build.yaml` was copy-paste nonsense. Thanks!

gspschmid:
`arm64 / build-t5x / build-upstream-t5x` failed after >2 1/2 hours (https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7670821005/job/20908452910?pr=500) but without any logs -- the log archive didn't have them either.

The fact it stopped in the `Build final image` step suggests something in `pip-finalize.sh` went wrong, though I'm a bit mystified why no logs would be recorded whatsoever. Also, next to the build step it says "2h 46m", whereas the overall job on top says it failed in "26m".

gspschmid:
Started a re-run of failed jobs w/ debug logging: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7670821005/job/20966741755

gspschmid:
Oh, this time around it worked (https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7670821005/job/20966741755) and only the Pallas tests (that also seems to be failing elsewhere) didn't?

gspschmid:
Re-ran once more and the new arm64 parts succeeded. I'm wondering whether the initial delay and failure were due to some caches not being populated? In any case, I would suggest merging this as is now.
cc @yhtang

olupton (approved):
LGTM

PR #504, opened 2024-01-29
Add nightly build for the Language to Reward 2023 model from Google DeepMind

DwarKapex:
The link to the repo of Language to Reward 2023: https://github.com/google-deepmind/language_to_reward_2023


yhtang (changes requested):
Can we stack this on top of the MJX container?

.github/workflows/nightly-mjx-build-test.yaml
```diff
@@ -1,5 +1,5 @@
-name: MJX build
-run-name: MJX build - (${{ github.event_name == 'workflow_run' && format('nightly {0}', github.event.workflow_run.created_at) || github.event_name }})
+name: Nightly MJX build
```
yhtang, line 1: We don't want to prepend 'Nightly', because when it runs by `on:workflow_dispatch`, we don't want to show 'nightly' in the name, and also because the run-name will add 'nightly' in the first part of the conditional context if the workflow is indeed run as a nightly.

yhtang (changes requested):

yhtang (approved):

PR #505, opened 2024-01-30
Remove Triton patch for Pallas builds

olupton:
It seems that the `HEAD~2` trick is no longer needed, as Python bindings are still there in the tag referred to by XLA.
The `is_hip()` patch no longer applies because of https://github.com/openai/triton/pull/2887/files#diff-3d1f29795218f61553ab953426c15fa1e4162b224405b85529022293054da57aL13-L16.
Pallas is not currently working, so this may need to be re-added in some other form.


yhtang:
Could you please provide an overview of the Pallas status?
> Pallas is not currently working, so this may need to be re-added in some other form.
Could you please provide a little more detail on what is working and what is not?

olupton:
> > Pallas is not currently working, so this may need to be re-added in some other form.
> 
> Could you please provide a little more detail on what is working and what is not?

There is work ongoing on the JAX side to change how lowering to Triton is organised.
Currently some relevant parts of the `jaxlib` wheels are broken, e.g.
```python
from jaxlib.mlir import ir
from jaxlib.triton import dialect as tt_dialect
ctx = ir.Context()
tt_dialect.register_dialect(ctx)
```
reproduces [the error seen in our nightly tests](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7694716097/job/20966057436#step:8:28).

Older images, such as https://github.com/NVIDIA/JAX-Toolbox/pkgs/container/jax/168056269?tag=nightly-pallas-2023-12-16, pre-date these ongoing changes to JAX and should work.

This PR is attempting to fix the mechanics of the nightly image build (which is currently failing), and in the first instance get us back to reproducing the error linked above in our nightly builds. We should also fix the badge, as the error is included in summaries but not made obvious from the README.

yhtang (approved):

PR #506, opened 2024-01-30
Add 5b pax test with fused attention

ashors1:


terrykong:
Is it possible to use [matrix.include](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/_test_t5x_rosetta.yaml#L225-L244) to simplify the amount of jobs you have?

terrykong (changes requested):
Is it possible to use [matrix.include](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/_test_t5x_rosetta.yaml#L225-L244) to simplify the amount of jobs you have?

terrykong (dismissed):
LGTM. FFTM once CI passes

yhtang (commented):
Not an objection to the PR, but I'm a bit concerned that adding more MGMN tests will make the SLURM cluster backlog even worse. We need to discuss a bit on this offline.

yhtang (approved):

PR #509, opened 2024-01-31
Adds t5x baselines and updates the baseline creation scripts to account for arbitrary artifact names

terrykong:
Adding metric baseline checking for rosetta t5x. We now have metric checking for:

- [x] upstream-t5x
- [x] rosetta-t5x (this PR)
- [x] upstream-pax
- [x] rosetta-pax

## Notable changes
* Updates the `create_baseline.sh` script to work with the latest set of tests. The script had broken due to changes in artifact names and the sitrep changes making artifact zips collide when unzipped
* Some test cases weren't included in create_baselines.sh because of failures observed a long time ago. I've added them all now, but only ran rosetta-t5x
* Renamed metric artifact for rosetta-pax since it would've collided with rosetta-t5x
* Given that the badges on the front page are broken b/c they rely on `_publish_badge.yaml`, this PR also should fix the badges


terrykong:
FYI: @kocchop

terrykong:
Will rely on presubmit, but triggered a rosetta-t5x run for early verification: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7732838672

.github/workflows/_test_pax_rosetta.yaml
```diff
@@ -36,7 +41,7 @@ jobs:
     runs-on: ubuntu-22.04
 
     env:
-      BADGE_FILENAME_FULL: rosetta-pax-single-process-multi-device-te
+      BADGE_FILENAME_PREFIX: rosetta-pax-single-process-multi-device-te
```
yhtang, line 44: Do we miss the `badge-` prefix here?

.github/workflows/_test_t5x_rosetta.yaml
```diff
@@ -78,7 +83,7 @@ jobs:
         run: |
           IMAGE="$(echo ${{inputs.T5X_IMAGE}} | sed 's/\//#/')"
           TEST_CASE_NAME=${{ matrix.TEST_NAME }}
-          JOB_NAME=rosetta-T5X-${GITHUB_RUN_ID}-${TEST_CASE_NAME}
+          JOB_NAME=${{ inputs.FW_NAME }}-${GITHUB_RUN_ID}-${TEST_CASE_NAME}
```
yhtang, line 86: I'd prefer `${{ github.run_id }}` over `${GITHUB_RUN_ID}` since the former is substituted at 'compilation' time and hence is easier for visual examination.

yhtang (dismissed):

.github/workflows/_test_pax_rosetta.yaml
terrykong: Oops. I missed that one. Good catch. I'll fix it

.github/workflows/_test_t5x_rosetta.yaml
terrykong: Sure, since this is the case for many workflows; I can bulk search-replace in a follow up PR to minimize surface of this PR

yhtang (approved):

PR #511, opened 2024-02-02
Maxtext docker functionality tests

kocchop:
Hi, I have added and verified the functionality checks for upstream-maxtext build. Here is the test run:
https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7746294404

Please let me know if anything else is required to make it public.


kocchop:
@terrykong I have made the changes to support build and test using the same yaml file. Also verified it to be working. Please see the following run:
https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7762930813

terrykong (approved):

PR #513, opened 2024-02-02
Use Clang when building JAX via build/build.py

andportnoy:
Previously the `CC=clang CXX=clang++` variables had no effect and we were building using GCC. Now that `build/build.py` has a supported `use_clang` flag, use that instead.

This PR can be merged after https://github.com/google/jax/pull/19633 is merged.

### DIFF
--- a/.github/container/build-jax.sh
+++ b/.github/container/build-jax.sh
@@ -277,7 +277,8 @@ pushd $SRC_PATH_JAX
 # Delete old wheel if one already exist.
 rm -rf dist/j*.whl
 
-time CC=clang CXX=clang++ python build/build.py \
+time python build/build.py \
+    --use_clang \
     --enable_cuda \
     --cuda_path=$TF_CUDA_PATHS \
     --cudnn_path=$TF_CUDNN_PATHS \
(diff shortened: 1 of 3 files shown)


yhtang (commented):
LGTM. Just putting in a convo as a manual blocker while we wait for the upstream PR to merge.

.github/container/build-jax.sh
```diff
@@ -277,7 +277,8 @@ pushd $SRC_PATH_JAX
 # Delete old wheel if one already exist.
 rm -rf dist/j*.whl
 
-time CC=clang CXX=clang++ python build/build.py \
+time python build/build.py \
+    --use_clang \
```
yhtang, line 281: Pending https://github.com/google/jax/pull/19633

andportnoy:
Build failing presumably because the JAX commit is pinned in the manifest: https://github.com/NVIDIA/JAX-Toolbox/blob/bbdf7d051e4ca3da9840573109775ccd43ae5279/.github/container/manifest.yaml#L5.

.github/container/build-jax.sh
olupton: Now merged.

olupton:
> Build failing presumably because the JAX commit is pinned in the manifest:
> 
> https://github.com/NVIDIA/JAX-Toolbox/blob/bbdf7d051e4ca3da9840573109775ccd43ae5279/.github/container/manifest.yaml#L5

I bumped the manifest just now (and rebased on `main`).

PR #514, opened 2024-02-05
Enable arm64 build of rosetta-t5x-nightly

gspschmid:
Briefly discussed this with @terrykong last week -- changes of Rosetta on top of T5X are in principle platform-agnostic, so this should hopefully work out of the box.


olupton (approved):
LGTM

PR #520, opened 2024-02-06
T5X arm64: Pin grain 0.0.4 source

gspschmid:
Grain's build script recently changed, breaking the way we currently call said script from the T5X arm64 build. I didn't find a quick way to fix this on the side of our Dockerfile (short of changing the base image). Instead, in the interest of unbreaking the T5X build, I pinned the grain sha corresponding to a recent, published version of the package, which still works. I'd suggest that going forward we focus on getting an arm64 wheel of grain published.

### DIFF
--- a/.github/container/manifest.yaml
+++ b/.github/container/manifest.yaml
@@ -132,12 +132,6 @@ mujoco:
   tracking_ref: main
   latest_verified_commit: ced37ffbb237584512311b041bce3124e3b2cc2a
   mode: git-clone
-grain:
-  # Used only in ARM t5x builds
-  url: https://github.com/google/grain.git
-  tracking_ref: main
-  latest_verified_commit: 41f80fb540547ec3aefe7cec57fe43228560b1cf
-  mode: git-clone
 mujoco-mpc:
   url: https://github.com/google-deepmind/mujoco_mpc.git
   tracking_ref: main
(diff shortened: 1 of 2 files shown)


.github/container/manifest.yaml
```diff
@@ -132,12 +132,6 @@ mujoco:
   tracking_ref: main
   latest_verified_commit: ced37ffbb237584512311b041bce3124e3b2cc2a
   mode: git-clone
-grain:
-  # Used only in ARM t5x builds
-  url: https://github.com/google/grain.git
-  tracking_ref: main
-  latest_verified_commit: 41f80fb540547ec3aefe7cec57fe43228560b1cf
-  mode: git-clone
```
nouiz, line 140: Could we add something like:
checkout: hash

to be more consistent and simplify it when this case happens?

gspschmid: I guess this would be analogous to the current `mode: git-clone`, but pinning the latest verified commit?

(Though I would still merge this now to unblock)

olupton: Missed today's nightly.
I think it would be cleaner if we could just add one line of YAML, something like
```yaml
grain:
  pin: true
```
that would tell `bump.sh` not to advance the commit hash, and thereby make the nightlies use the pinned commit.
That said, that feature should be used sparingly and always temporarily.

terrykong: Does setting `tracking_ref: 41f80fb540547ec3aefe7cec57fe43228560b1cf` work?

yhtang (approved):

PR #524, opened 2024-02-08
Unify JAX and Pallas unit test workflows to overcome the 20-workflow limit in CI

yhtang:
This is currently done in a sort of 'brute-force' way by stitching together the two test files and use a workflow input to conditionally run one of the tests. A more systematic approach can be done in the future.

### DIFF
--- a/.github/workflows/_ci.yaml
+++ b/.github/workflows/_ci.yaml
@@ -201,9 +201,12 @@ jobs:
   test-jax:
     needs: build-jax
     if: inputs.ARCHITECTURE == 'amd64' # arm64 runners n/a
-    uses: ./.github/workflows/_test_jax.yaml
+    uses: ./.github/workflows/_test_unit.yaml
     with:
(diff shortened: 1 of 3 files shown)


olupton (approved):
Code LGTM. In https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7825968984/job/21354245329?pr=524 the Pallas test jobs are shown as skipped, but I'm not sure if this is because the JAX build failed on the first attempt and the Pallas build from the retry has not succeeded yet. Hopefully that will change in the next few mins and I can merge before the nightlies.

yhtang:
Just for documentation purpose, this is what the actual run looks like
image
So there will be two 'cancelled' jax/pallas test each spawning from the workflow from for the other software. For example, there will be a 'cancelled' pallas test as part of the jax test, and vice versa. These can be distinguished from the actual ones by checking if it contains the GPU model suffix.

PR #527, opened 2024-02-08
Cancel CI workflow if it is triggered by a draft PR

yhtang:
Example of the workflow that is cancelled when this draft PR is opened: 
- draft: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7835375902/attempts/1
- rerun after ready for review: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7835375902/attempts/2

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,6 +4,11 @@ on:
   schedule:
     - cron: '30 9 * * *'  # Pacific Time 01:30 AM in UTC
   pull_request:
+    types:
+      - opened
+      - reopened
+      - ready_for_review
+      - synchronize
     paths-ignore:
       - '**.md'
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


yhtang:
One issue appears to be that we **cannot** rerun a cancelled workflow after flipping the draft mode, as it seems to be tied with the state of the PR at the time the workflow is created. Hence, a dummy commit must be pushed if a PR is marked as ready for review to trigger a full CI.

terrykong:
Is there any reason we can't use: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request ?

Looks like it supports things like:
- ready_for_review
- review_requested

terrykong:
For a future PR, perhaps we can also make use of https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_comment-use-issue_comment

where I imagine you can check the payload for for things like:

```
/ci jax
/ci build
/ci t5x
/ci t5x/test
```

populate a "metadata" step and then each job can run if its selector matches whatever the comment was

and run stages that match the key we expect

yhtang:
> Is there any reason we can't use: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request ?
> 
> Looks like it supports things like:
> 
> * ready_for_review
> * review_requested

That's cool. I just didn't realize it's there :grin:

yhtang:
I ended up using a combination of `on.pull_request.types` and the RESTful API.

yhtang:
Let's merge and see how it works. If not, please give me feedback and I'd be happy to iterate quickly on this.

PR #533, opened 2024-02-13
updated patch for new maxtext software version

kocchop:
updated patch for new maxtext software version

### DIFF
--- a/.github/container/test-maxtext.sh
+++ b/.github/container/test-maxtext.sh
@@ -35,7 +35,7 @@ if [[ $? -ne 0 ]]; then
 fi
 
 # Default arguments
-MULTIPROCESS=false
+HARDWARE='tpu'
 OUTPUT=$(mktemp -d)
 
 BATCH_PER_GPU=2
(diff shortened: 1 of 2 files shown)


.github/container/test-maxtext.sh
```diff
@@ -35,7 +35,7 @@ if [[ $? -ne 0 ]]; then
 fi
 
 # Default arguments
-MULTIPROCESS=false
+HARDWARE='tpu'
```
terrykong, line 38: Do you mind leaving a comment that this is essentially a no-op? As it stands it kind of looks like we test tpu by default

terrykong (changes requested):

.github/container/test-maxtext.sh
kocchop: Found another way around and also updated files @terrykong

terrykong (approved):

olupton:
Just wondering: is there a reason this is being managed via a patch file in the repo instead of via external pr/branch like the others (rosetta pax etc.)?

kocchop:
Hi @olupton, I am working towards that, will soon convert the patching similar to the rest

PR #534, opened 2024-02-13
Use Clang to build MJX

DwarKapex:
I got a compilation issue while compiling mujoco-mpc with GNU:
```
#10 912.6         /opt/mujoco-mpc/mjpc/../mjpc/spline/spline.h:106:9: error: expected unqualified-id before ‘const’
#10 912.6           106 |         const IteratorT<SplineType, NodeType>& other) = default;
#10 912.6               |         ^~~~~
#10 912.6         /opt/mujoco-mpc/mjpc/../mjpc/spline/spline.h:105:37: error: expected ‘)’ before ‘const’
#10 912.6           105 |     IteratorT<SplineType, NodeType>(
#10 912.6               |                                    ~^
#10 912.6               |                                     )
#10 912.6           106 |         const IteratorT<SplineType, NodeType>& other) = default;
#10 912.6               |         ~~~~~
#10 912.6         /opt/mujoco-mpc/mjpc/../mjpc/spline/spline.h:109:68: error: expected ‘)’ before ‘&&’ token
#10 912.6           109 |     IteratorT<SplineType, NodeType>(IteratorT<SplineType, NodeType>&& other) =
#10 912.6               |                                    ~                               ^~
#10 912.6               |                                                                    )
```

The developers of `mujoco-mpc` use Clang to compile (https://github.com/google-deepmind/mujoco_mpc/blob/main/.github/workflows/build.yml#L19).
From now on, the default compiler for MJX is Clang, but can be changed via docker arguments.

### DIFF
--- a/.github/container/Dockerfile.mjx
+++ b/.github/container/Dockerfile.mjx
@@ -31,6 +31,8 @@ EOF
 
 # Specify installation targets
 RUN <<"EOF" bash -ex
+export CC=/usr/bin/clang
+export CXX=/usr/bin/clang++
 get-source.sh -l mujoco -m ${MANIFEST_FILE} -b $(dirname ${SRC_PATH_MUJOCO})
 get-source.sh -l mujoco-mpc -m ${MANIFEST_FILE}
 get-source.sh -l language-to-reward-2023 -m ${MANIFEST_FILE}
--- a/.github/container/Dockerfile.mjx
+++ b/.github/container/Dockerfile.mjx
@@ -4,7 +4,8 @@ ARG BASE_IMAGE=ghcr.io/nvidia/jax:mealkit
 ARG SRC_PATH_MUJOCO=/opt/mujoco
 ARG SRC_PATH_MUJOCO_MPC=/opt/mujoco-mpc
 ARG SRC_PATH_L2R=/opt/language-to-reward-2023
-
+ARG CC=/usr/bin/clang
+ARG CXX=/usr/bin/clang++
 
 ###############################################################################
 ## Download source and add auxiliary scripts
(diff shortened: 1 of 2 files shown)


.github/container/Dockerfile.mjx
```diff
@@ -14,6 +15,11 @@ FROM ${BASE_IMAGE} as mealkit
 ARG SRC_PATH_MUJOCO
 ARG SRC_PATH_MUJOCO_MPC
 ARG SRC_PATH_L2R
+ARG CC
+ARG CXX
+
+ENV CC=${CC}
```
yhtang, line 21: Instead of enforcing the compiler toolchain globally, can we set it only for the MPC installation step?

yhtang (changes requested):

.github/container/Dockerfile.mjx
```diff
@@ -47,4 +50,8 @@ EOF
 
 FROM mealkit as final
 
-RUN pip-finalize.sh
+# compile MuJoCo-MPC with Clang separately 
+RUN CC=${CC_COMPILER} CXX=${CXX_COMPILER} pip-finalize.sh /opt/pip-tools.d/requirements-mpc.in
+
+# compiler the rest dependencies
+RUN pip-finalize.sh /opt/pip-tools.d/requirements-mjx.in /opt/pip-tools.d/requirements-l2r.in
```
yhtang, line 57: Why do we need to run `pip-finalize` twice here? Due to the use of `pip-sync` in the script, the second run will erase what the first run installed.

yhtang (changes requested):

yhtang:
The plan from an offline discussion between @DwarKapex and me is to use multi-stage build to create the MPC wheel using clang and then bring it into the mealkit/final image to avoid interfering with the default compiler toolchain. This is similar to how we build the jaxlib wheel in the JAX core image and also how we build the tensorflow-text/lingvo packages in the T5X/Pax arm64 images.

PR #536, opened 2024-02-13
nsys/ncu: move update logic into separate script

olupton:
As requested by @yhtang.

### DIFF
--- a/.github/container/install-nsight.sh
+++ b/.github/container/install-nsight.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+set -ex -o pipefail
+
+# Repo for newer nsight versions
+export UBUNTU_ARCH=$(dpkg --print-architecture)
+export UBUNTU_VERSION=$(. /etc/os-release && echo ${ID}${VERSION_ID/./}) # e.g. ubuntu2204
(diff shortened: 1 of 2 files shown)


yhtang:
Pasting some offline discussion for documentation purposes:

> There is of course a cost to this; we do more fetching/deleting apt caches the more we make these into atomic scripts.

Indeed. A more systematic solution might have to do with multi-staging: we can create a stage just for storing the apt cache, and bind mount that cache to later stages where the apt installations actually happen.

.github/container/install-nsight.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -ex -o pipefail
+
+# Repo for newer nsight versions
+export UBUNTU_ARCH=$(dpkg --print-architecture)
+export UBUNTU_VERSION=$(. /etc/os-release && echo ${ID}${VERSION_ID/./}) # e.g. ubuntu2204
+curl -o /usr/share/keyrings/nvidia.pub https://developer.download.nvidia.com/devtools/repos/${UBUNTU_VERSION}/${UBUNTU_ARCH}/nvidia.pub
```
yhtang, line 7: Nitpick: the `https://developer.download.nvidia.com/devtools/repos/${UBUNTU_VERSION}/${UBUNTU_ARCH}` thing gets repeated twice here. Maybe use a variable to hold it and reuse?

olupton: Changed.

yhtang (approved):

PR #542, opened 2024-02-15
update manifest to support mlperf-logging for maxtext

kocchop:
Added the mlperf-logging to the manifest.yaml and also verified it using this [run](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7907539388)

### DIFF
--- a/.github/container/manifest.yaml
+++ b/.github/container/manifest.yaml
@@ -148,3 +148,8 @@ language-to-reward-2023:
   tracking_ref: main
   latest_verified_commit: abb8e5125e4ecd0da378490b73448c05a694def5
   mode: git-clone
+logging:
+  url: https://github.com/mlcommons/logging.git
+  tracking_ref: main 
+  latest_verified_commit: c94564a66a6b13b4da6248847f3799b463283454
+  mode: git-clone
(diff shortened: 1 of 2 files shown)


.github/container/manifest.yaml
```diff
@@ -148,3 +148,8 @@ language-to-reward-2023:
   tracking_ref: main
   latest_verified_commit: abb8e5125e4ecd0da378490b73448c05a694def5
   mode: git-clone
+mlperf-logging:
+  url: https://github.com/mlcommons/logging.git
+  tracking_ref: main 
+  latest_verified_commit: 38709131757a786d7f66150243b49eb0365324d7
+  mode: git-clone
```
terrykong, line 155: You can set it to `pip-vcs` to document that this isn't source that we clone, but used just during pip-compile + pip-sync
```suggestion
  mode: pip-vcs
```

terrykong (changes requested):
Review was a standalone comment

.github/container/manifest.yaml
kocchop: changed it to pip-vcs

terrykong (approved):
Lgtm

Issue #525, opened 2024-02-08
Unique name for SLURM-launched A100 runners


yhtang:
[This workflow](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7825933884) run exposed an issue with our current workflow: both JAX and Pallas unit test calls the `_runner_ondemand_slurm.yaml` workflow to create A100 runners. If two such calls happens in fast succession, they ended up creating two runners that may be scheduled by the SLURM cluster at the same time while having identical names (`A100-${{ github_run_id }}`), thus causing issue for the actual job to properly landed in the runner (more detail to be discovered here).

To fix potential conflicts between runners launched this way, the runner need to have different names, i.e. having a UUID as part of the name, etc.

PR #544, opened 2024-02-15
Append unique number to SLURM-based runner jobs

yhtang:
This is to avoid multiple A100 runners within the same run to overwrite the log of each other.

### DIFF
--- a/.github/workflows/_runner_ondemand_slurm.yaml
+++ b/.github/workflows/_runner_ondemand_slurm.yaml
@@ -44,7 +44,7 @@ jobs:
         id: meta
         shell: bash -x -e {0}
         run: |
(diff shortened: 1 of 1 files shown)


yhtang:
> Update the description with the issue number that this close? I think I saw an issue for this.

It is related to #525, but does not actually address the real problem there. This current PR would help to provide insight into the impact of the problem inside #525, i.e. whether having jobs cross-use runners would cause real problem.

yhtang:
In particular, I'm seeing sometimes our CI run would launch 3 A100 runners, but only 2 of the 3 test jobs that requires A100 would be able to find a runner. One of the A100 runner somehow disappeared. The runners are tied to each run but not tied to individual jobs within the run. I'm still trying to figure out what exactly is going on.

Issue #540, opened 2024-02-14
Report SLURM runner status as `launch-slurm-runner` job outcome


yhtang:
We provision GitHub Actions runners for running JAX/Pallas/Levanter A100 unit tests using SLURM jobs that launch ephemeral runner instances inside the job. Currently, the [launch-slurm-runner](https://github.com/NVIDIA/JAX-Toolbox/blob/c318902ee5f1a66dfc94e445fdf93bec4ae6e2b1/.github/workflows/_runner_ondemand_slurm.yaml#L22) job will succeed as long as it is able to submit the job and wait for the job to finish/terminate. If the runner fails to register, e.g. due to network issue, then `launch-slurm-runner` would succeed, while the actual unit test job that needs to use the runner will wait forever. Even cancelling and then restarting the workflow cannot work around this, as the launcher job is marked as successfully and won't be rerun.

To fix that, we need the `launch-slurm-runner` job to query the status of the runner running inside the job. If the runner errors out without being able to handle a job, then the CI job should also fail.

Note the difference between
1) `launch-slurm-runner` successfully starts an A100 runner, which picks up a job that run and failed, and
2) `launch-slurm-runner` fails to start an A100 runner.

The latter is what we need to address in this issue.

PR #554, opened 2024-02-15
[WIP] Address issue #540

DwarKapex:


.github/workflows/_runner_ondemand_slurm.yaml
```diff
@@ -98,9 +99,26 @@ jobs:
           echo "SLURM_JOB_ID=$(cat ${SLURM_JOB_ID_FILE})" >> $GITHUB_OUTPUT
 
       - name: Wait for SLURM job to complete
+        id: job-completion
```
yhtang, line 102: Is this job id used?

.github/workflows/_runner_ondemand_slurm.yaml
```diff
@@ -98,9 +99,26 @@ jobs:
...
           while ssh ${{ secrets.CLUSTER_LOGIN_USER }}@${{ vars.HOSTNAME_SLURM_LOGIN }} squeue -j${{ steps.submit.outputs.SLURM_JOB_ID }} > /dev/null 2>&1; do echo "wait"; sleep 15; done
+          SLURM_JOB_COMPLETION_STATUS_FILE=$(mktemp)
+          ssh ${{ secrets.CLUSTER_LOGIN_USER }}@${{ vars.HOSTNAME_SLURM_LOGIN }} >${SLURM_JOB_COMPLETION_STATUS_FILE} \
+            sacct -XP -j ${{ steps.submit.outputs.SLURM_JOB_ID }}
+
+          echo "JOB COMPLETION STATUS"
+          cat ${SLURM_JOB_COMPLETION_STATUS_FILE}
+
+          SLURM_JOB_COMPLETION_STATE_FILE=$(mktemp)
+          ssh ${{ secrets.CLUSTER_LOGIN_USER }}@${{ vars.HOSTNAME_SLURM_LOGIN }} >${SLURM_JOB_COMPLETION_STATE_FILE} \
```
yhtang, line 114: Why repeat this? Can we `cut -d| -f<N>` from the previous result?

yhtang (changes requested):
Please also configure the submitted job script so that the job would fail if an individual command in it fails. Otherwise, the job will be successful from the eyes of SLURM regardless of whether the runner did its work.

.github/workflows/_runner_ondemand_slurm.yaml
DwarKapex: Removed

.github/workflows/_runner_ondemand_slurm.yaml
DwarKapex: Done

DwarKapex:
> Please also configure the submitted job script so that the job would fail if an individual command in it fails. Otherwise, the job will be successful from the eyes of SLURM regardless of whether the runner did its work.

IIUC, slurm job [state](https://slurm.schedmd.com/squeue.html#lbAG) shows what happens with the job. The state covers all scenarios including failure of individual commands, allocation failures, timeout and etc. IMO, we are good here.

yhtang:
> > Please also configure the submitted job script so that the job would fail if an individual command in it fails. Otherwise, the job will be successful from the eyes of SLURM regardless of whether the runner did its work.
> 
> IIUC, slurm job [state](https://slurm.schedmd.com/squeue.html#lbAG) shows what happens with the job. The state covers all scenarios including failure of individual commands, allocation failures, timeout and etc. IMO, we are good here.

That doc concerns the overall status of the job script, which can be different from the exit status of individual commands in the script. For example:
```
sbatch --parsable <<EOF
#!/bin/bash
#SBATCH -p <partition>
cat non-existing.txt
EOF
```
leads to
```
$ sacct -XP -j <job_id> -o State
State
FAILED
```
while
```
sbatch --parsable <<EOF
#!/bin/bash
#SBATCH -p <partition>
cat non-existing.txt
echo Hello
EOF
```
leads to
```
$ sacct -XP -j <job_id> -o State
State
COMPLETED
```

The difference is that in the second example, the exit status of the subsequent `echo` command replaces the exit status of the failing `cat` command.

A simple fix for this situation is to add `set -x -o pipefail` to the batch script.

**What is more important** is that the docker run command, which starts a GitHub Actions runner instance, may gracefully handle a registration/runtime error, report it in the log, and then exits without an explicit non-zero status code. For this, I'm not 100% what the specific behavior is so please check the documentation of [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners) and figure out a solution accordingly.

PR #566, opened 2024-02-16
[WIP] Automatic all green bump

terrykong:
This change introduces a mechanism to automatically bump the manifest & patches in the triggering branch. So for nightlies that trigger on `main`, this will bump if all tests have succeeded.

There is now a new input for the unified workflow: `MERGE_BUMPED_MANIFEST`, which controls whether or not to make a PR for the manifest bump.
* In scheduled pipelines (nightlies), `MERGE_BUMPED_MANIFEST` is forced to be `true` and will take a bumped manifest branch, PR it (important in case there's a merge conflict that requires manual intervention), and then merge/delete the manifest branch.
* If `BUMP_MANIFEST=false`, `MERGE_BUMPED_MANIFEST=true`, there will be an error.

Our two common uses cases are presubmits and nightlies.
* For nightlies, BUMP_MANIFEST/MERGE_BUMPED_MANIFEST are forced to be `true`.
* For presubmit, the default is taken, which is `false` for both.

Example runs:
| BUMP_MANIFEST | MERGE_BUMPED_MANIFEST | run |
| --- | --- | ---|
| Yes|Yes|https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7934518919|
|Yes|No|https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7934526041|
|No|No|https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7934701030|
|No|Yes|https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7934712287 (errors)|

## Changes to current workflow
Prior to this PR, the manifest was bumped separately in amd64 and arm64 jobs. This could produce potentially two separate versions of the manifest, as well as make it difficult to decide which manifest to merge back in.

The solution taken is to reintroduce the "trial branch" mechanism, but name it `MANIFEST_BRANCH` to better reflect its contents. The manifest branch is created in the top level `ci.yaml` workflow and passed to the reusable `_ci.yaml` workflow.

## Notes about the manifest branch
The manifest branch is always deleted to avoid cluttering the git branches. Information should not be lost since a PR will be open for nightlies so that the changeset on the manifest can be recovered and manually merged.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -141,9 +141,9 @@ jobs:
     secrets: inherit
 
   # Only bump if everything succeeds
-  commit-new-manifest:
+  merge-new-manifest:
     runs-on: ubuntu-22.04
-    if: needs.metadata.outputs.MERGE_BUMPED_MANIFEST == 'true'
+    if: needs.metadata.outputs.MERGE_BUMPED_MANIFEST == 'true' && needs.metadata.outputs.MANIFEST_BRANCH != github.sha
     needs:
       - metadata
       - amd64
@@ -156,7 +156,8 @@ jobs:
           route: POST /repos/{owner_and_repo}/pulls
           owner_and_repo: ${{ github.repository }}
           head: ${{ needs.metadata.outputs.MANIFEST_BRANCH }}
-          base: automatic-all-green-bump
+          # Always try to merge back into the branch that triggered this workflow
+          base: ${{ github.ref }}
           body: |
             https://github.com/NVIDIA/JAX-Toolbox/actions/runs/${{ github.run_id }}
           title: Nightly Manifest Bump (${{ needs.metadata.outputs.BUILD_DATE }})
@@ -164,6 +165,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: "Merging PR #${{ fromJson(steps.create_pr.outputs.data).number }}"
+        id: merge_pr
         uses: octokit/request-action@v2.x
         with:
           route: PUT /repos/{owner_and_repo}/pulls/${{ fromJson(steps.create_pr.outputs.data).number }}/merge
@@ -172,14 +174,15 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         
-  # Optionally clean up manifest branch
+  # Always clean up manifest branch since PRs hold the changeset
   cleanup-manifest-branch:
     runs-on: ubuntu-22.04
-    if: always() && needs.metadata.outputs.MERGE_BUMPED_MANIFEST == 'false'
+    if: always() && needs.metadata.outputs.MANIFEST_BRANCH != github.sha
     needs:
       - metadata
       - amd64
       - arm64
+      - merge-new-manifest
     steps:
       - name: Check that the branch matches znightly- prefix
         run: |
@@ -188,6 +191,8 @@ jobs:
             exit 1
           fi
       - name: Deleting MANIFEST_BRANCH=${{ needs.metadata.outputs.MANIFEST_BRANCH }}
+        # Delete can fail if branch was already deleted or not created, e.g., if the PR successfully merges, then branch is also already deleted.
+        continue-on-error: true
         uses: octokit/request-action@v2.x
         with:
           route: DELETE /repos/{owner_and_repo}/git/refs/heads/${{ needs.metadata.outputs.MANIFEST_BRANCH }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -106,6 +106,9 @@ jobs:
           bash bump.sh --input-manifest manifest.yaml
           git diff
 
+      # In the unusual situation where the manifest is the same even after bumping,
+      # we will produce an empty commit with --allow-empty, which allows a PR to be
+      # made and merged even with no changeset.
       - name: Push manifest branch
         shell: bash -x -e {0}
         run: |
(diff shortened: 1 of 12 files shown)


.github/workflows/ci.yaml
```diff
@@ -65,30 +77,158 @@ jobs:
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ needs.metadata.outputs.MANIFEST_ARTIFACT_NAME }}
+          path: |
+            .github/container/manifest.yaml
+            .github/container/patches
+
+  amd64:
+    needs: [metadata, maybe-bump-manifest]
+    if: "!cancelled()"
```
yhtang, line 131: Why do we need this if-not-cancelled logic here?

.github/workflows/ci.yaml
```diff
@@ -65,30 +77,158 @@ jobs:
...
       ARCHITECTURE: arm64
       BUILD_DATE: ${{ needs.metadata.outputs.BUILD_DATE }}
-      BUMP_MANIFEST: ${{ needs.metadata.outputs.BUMP_MANIFEST == 'true' }}      
+      MANIFEST_ARTIFACT_NAME: ${{ needs.metadata.outputs.MANIFEST_ARTIFACT_NAME }}
     secrets: inherit
 
+  # Only merge if everything succeeds
+  merge-new-manifest:
+    runs-on: ubuntu-22.04
+    if: ${{ !cancelled() && needs.metadata.outputs.MERGE_BUMPED_MANIFEST == 'true' && needs.metadata.outputs.MANIFEST_BRANCH != github.sha }}
```
yhtang, line 152: Would it make better same to remove the `!cancelled()` part so that we can cancel a workflow as a 'emergency-stop' button to stop merging?

.github/workflows/ci.yaml
terrykong: Accidentally introduced that, probably when I was rebasing. Will remove

.github/workflows/ci.yaml
terrykong: My understanding is I need this `!cancelled()` b/c I want `mege-new-manifest` to run regardless of the `needs` jobs succeeding. If any of them fail, then there's an implicit "skip descendants" logic so if arm64 failed, I would skip this entire job. The reason I want the job to run regardless is so I can create a PR.

Also, doesn't `!cancelled()` respect emergency stop? Isn't that why it's preferred over `always()`?

terrykong:
@yhtang I updated the bumping logic to always happen, similar to what you introduced in main. I needed to add a new flag for patches since those were always updated in place as well as the manifest.

.github/workflows/ci.yaml
```diff
@@ -23,15 +23,24 @@ on:
...
-  contents: read  # to fetch code
-  actions:  write # to cancel previous workflows
-  packages: write # to upload container
+  contents: write       # to fetch code and push branch
+  actions:  write       # to cancel previous workflows
+  packages: write       # to upload container
+  pull-requests: write  # to make pull request for manifest bump
+
+env:
+  DEFAULT_MANIFEST_ARTIFACT_NAME: maybe-bumped-manifest
```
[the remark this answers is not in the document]
terrykong: done

.github/workflows/ci.yaml
yhtang: > Also, doesn't !cancelled() respect emergency stop? Isn't that why it's preferred over always()?

You are right 😂 

> I want mege-new-manifest to run regardless of the needs jobs succeeding.

Is this supposedly the 'all-green' merge so that this would not run if the whole run is not successful? Or is it because you want the PR to be created anyway, and just not *merge* it if it does not succeed?

.github/workflows/ci.yaml
```diff
@@ -65,30 +77,164 @@ jobs:
...
+          git config user.email "jax@nvidia.com"
+          manifest_branch=${{ needs.metadata.outputs.MANIFEST_BRANCH }}
+          git switch -c $manifest_branch
+          git status
+          git add .github/container/patches/
+          git status
+          git commit --allow-empty -a -m "Nightly Manifest Bump (${{ needs.metadata.outputs.BUILD_DATE }}) for run_id=${{ github.run_id }}: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/${{ github.run_id }}"
+          git push --set-upstream origin $manifest_branch
+
+      - name: Creating PR for MANIFEST_BRANCH=${{ needs.metadata.outputs.MANIFEST_BRANCH }}
```
yhtang, line 194: Can we prevent presubmit CI from running on this PR, since the PR itself is the result of a full CI?

However, in the future, when we do partial bumps, it might sense to run presubmit CI.

.github/workflows/ci.yaml
```diff
@@ -65,30 +77,164 @@ jobs:
...
+          head: ${{ needs.metadata.outputs.MANIFEST_BRANCH }}
+          # Always try to merge back into the branch that triggered this workflow
+          base: ${{ github.ref }}
+          body: |
+            https://github.com/NVIDIA/JAX-Toolbox/actions/runs/${{ github.run_id }}
+          title: Nightly Manifest Bump (${{ needs.metadata.outputs.BUILD_DATE }})
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: "Merging PR #${{ fromJson(steps.create_pr.outputs.data).number }}"
```
yhtang, line 209: For future considerations, can we condition the merge on presubmit success?

.github/workflows/ci.yaml
yhtang: It would be fine for the time being, but in the long run I would rather we do not create PRs from failed runs. Instead, (as we discussed offline last week), we should create a separate workflow with a manual `workflow_dispatch` trigger to turn a workflow run into a PR.

terrykong:
Converting to draft to address offline comments from @yhtang :
* manually merge via `git` and if fail, then open draft PR

yhtang (approved):
LGTM. Thanks!

yhtang (changes requested):
Actually, seems like there are some minor bugs to be fixed.

terrykong:
@yhtang Yea, there was the code path of the presubmit itself, which I missed. This should address the bug: https://github.com/NVIDIA/JAX-Toolbox/pull/566/commits/a4b875acb4ed1dffd7db327e67ae166aa95858cd

yhtang (approved):

olupton:
It looks like this broke the handling of patches to `openxla-triton`: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/8062509528/job/22023406454

PR #570, opened 2024-02-20
Apply XLA's patches to XLA's tag of Triton

olupton:
XLA maintains a list of patches: https://github.com/openxla/xla/blob/e83773328ac3b327e4925b877f2988bb45aa7bf1/third_party/triton/workspace.bzl#L16-L18 that should be applied on top of its specific tag of Triton.

Previously we did not apply these when building the Pallas containers. This PR applies them.

I did initially try to use more of the Rosetta machinery for applying patches, but this was a bit painful. Note that Google's patches do not contain headers like the commit author/date/message, and that confuses the `git am` command that Rosetta uses.

https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7972292328 was launched with the manifest bump enabled, to test that side of the changeset.

### DIFF
--- a/.github/container/bump-openxla-triton.sh
+++ b/.github/container/bump-openxla-triton.sh
@@ -6,7 +6,7 @@ usage() {
 cat <<EOF
 This script is a utility for updating the commit reference for openxla-triton
 in a manifest YAML file used to build JAX-Toolbox images. The commit is derived
-from the commit for xla contained in the manifest.
+from the commit for xla contained in the manifest, along with the patches.
 
 Usage: $0 [OPTION]...
   -h, --help      Print usage.
@@ -50,6 +50,7 @@ if [[ -z "${MANIFEST:-}" ]]; then
 fi
 
 set -eou pipefail
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
 xla_url=$(yq e ".xla.url" $MANIFEST)
 xla_tracking_ref=$(yq e ".xla.tracking_ref" $MANIFEST)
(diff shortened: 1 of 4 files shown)


.github/container/bump-openxla-triton.sh
```diff
@@ -58,8 +59,17 @@ xla_commit=$(yq e ".xla.latest_verified_commit" $MANIFEST)
...
 (cd "${xla_repo}" && git checkout "${xla_commit}")
 # Extract the openxla/triton tag used by XLA. Even though it is called
-# TRITON_COMMIT it is a tag. In principle we should also account for the
-# patches in this .bzl file, but skip that for now.
-openxla_triton_tag=$(sed -n -e 's#\s\+TRITON_COMMIT = "\(cl[0-9]\+\)"#\1#p' "${xla_repo}/third_party/triton/workspace.bzl")
+# TRITON_COMMIT it is a tag.
+workspace_file="${xla_repo}/third_party/triton/workspace.bzl"
+openxla_triton_tag=$(sed -n -e 's#\s\+TRITON_COMMIT = "\(cl[0-9]\+\)"#\1#p' "${workspace_file}")
+# Extract Triton patch files applied by XLA
+patch_files=$(python3 -c 'import ast, sys; tree = ast.parse(sys.stdin.read()); print(" ".join(elem.value.removeprefix("//third_party/triton:").removesuffix(".patch") for node in ast.walk(tree) if isinstance(node, ast.keyword) and node.arg == "patch_file" for elem in node.value.elts))' < "${workspace_file}")
```
andportnoy, line 66: Nice!

andportnoy (approved):
Thank you!

.github/container/bump-openxla-triton.sh
yhtang: Can we do a naive `ls third_party/triton/*.patch` as a more readable alternative solution?

olupton: Unfortunately not. For example, at this moment in history: https://github.com/openxla/xla/blob/d0ca61b6bdd3d8e2020436b1c3073226429e72f0/third_party/triton/workspace.bzl there were three patch files in the directory, two were listed in the workspace file, and one of those was commented out. I'm also not sure it's guaranteed that the patches should/can be applied in sorted-by-filename order.

PR #578, opened 2024-02-21
Remove chex pin from pax build causing failure due to pax using later than 0.1.7 now

terrykong:
Failing build: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7971250761/job/21763211927#step:10:240


ashors1 (approved):
Assuming build passes, LGTM

PR #581, opened 2024-02-22
Fix steps/sec computation in Pax tests

ashors1:
There are currently two issues affecting the reported step times in pax:
1. First `steps/sec` value includes compilation time
2. `steps/sec` computation includes eval step compilation and eval step time

To fix (1), we exclude the first `steps/sec` value from the average step time. To fix (2), we disable eval.


.github/container/test-pax.sh
```diff
@@ -247,6 +247,7 @@ class GPT126MPP(TransformerLmSpmdPipelineAdam):
   ## dropout
   DROPOUT_PROB = dropout
 
+  EVAL_INTERVAL_STEPS = 100000
```
terrykong, line 250: Can a comment be added here to explain why it is set this high?

terrykong (changes requested):

terrykong (approved):
Assuming CI passes; adding comment shouldn't need retrigger CI

ashors1:
> This PR now has a merge conflict due to updates to manifest.yaml, where every package seems to have been updated. Is it necessary for this PR? If not, we can handle version bumps in another PR.

I bumped the manifest because we were seeing a dependency issue with Seqio, so all Pax tests were failing. I tried selectively bumping Seqio, but that did not solve the problem. Since it was not clear to me where exactly the dependency issue was introduced, I tried instead bumping every package in the manifest to see whether that would fix the problem. But if you're wary of this approach, I can investigate more and bump just those packages that fix the Seqio dependency issue.

yhtang:
> > This PR now has a merge conflict due to updates to manifest.yaml, where every package seems to have been updated. Is it necessary for this PR? If not, we can handle version bumps in another PR.
> 
> I bumped the manifest because we were seeing a dependency issue with Seqio, so all Pax tests were failing. I tried selectively bumping Seqio, but that did not solve the problem. Since it was not clear to me where exactly the dependency issue was introduced, I tried instead bumping every package in the manifest to see whether that would fix the problem. But if you're wary of this approach, I can investigate more and bump just those packages that fix the Seqio dependency issue.

The downside of bumping manifests alongside other changes is that, as what has happened here, that the PR may remain active for a few days while the upstream manifest evolved. Ideally, if we can fix Pax tests via another PR and then only do changes related to the steps/second measurements in this PR that'll be much more secure.

yhtang (approved):

terrykong (approved):

olupton:
FWIW this PR contained an incomplete update to the manifest: it updated the `xla` commit without updating the dependent `openxla-triton` commit and patches. `bump.sh` is supposed to call `bump-openxla-triton.sh` and take care of this. I don't know what happened in this instance.

PR #582, opened 2024-02-23
Set environment variables to guard against hangs with Triton gemms

ashors1:


yhtang (approved):

nouiz:
Should we enable triton_gemm at the same time?
If not, there is no need to enable it right now. It will only slow down compilation without any gain.

yhtang (approved):

nouiz:
We should document them.

PR #586, opened 2024-02-26
Add documentation about profiling JAX programs with NSight Systems

olupton:
This should complement the main JAX documentation and provide useful resources for people profiling JAX programs on GPU.

cc: @sbhavani @yhtang @nouiz


yhtang (commented):
Cool work :grin:

Nitpick: maybe change the tone of the following paragraph a bit? I see 'likely' get used twice when we are giving recommendations and also am not sure why ncu is considered 'invasive'.


> The two tools that are most likely to be relevant are [Nsight Systems](https://developer.nvidia.com/nsight-systems) and [Nsight Compute](https://developer.nvidia.com/nsight-compute).
> 
> Nsight Systems provides a high level overview of activity on the CPU and GPU, and is likely to be the best place to start investigating the performance of your program. It has small overheads and should not significantly affect the execution time of your program.
> 
> Nsight Compute, on the other hand, enables detailed performance analysis of individual GPU kernels. It is very invasive and repeatedly re-runs the kernel(s) in question to collect different metrics.

nouiz (approved):
Already reviewed earlier draft. LGTM. Didn't re-read it.

nouiz (approved):

PR #590, opened 2024-02-27
Add gemma dockerfile

DwarKapex:
Add a new docker file for [GEMMA](https://github.com/google-deepmind/gemma)


yhtang (changes requested):
Let's wait for Abhinav to provide the MaxText version instead of this vanilla version.

sbhavani:
Could we add Jupyter Lab to this image? There's some notebook examples I think would be good to include

sbhavani (approved):
looks good to me

DwarKapex:
> Let's wait for Abhinav to provide the MaxText version instead of this vanilla version.

Done

.DS_Store
olupton, line 1: This should be removed.

rosetta/Dockerfile.gemma
```diff
@@ -0,0 +1,71 @@
...
+optax
+protobuf
+tfds-nightly
+tensorflow
+tensorflow-text
+tensorflow-gan
+" >> /opt/pip-tools.d/requirements-gemma.in
+EOF
+
+ENV PYTHONPATH "/opt/big_vision/:${PYTHONPATH}"
```
olupton, line 60: It's a bit unfortunate that they didn't make this big_vision project installable.
First, this should be `${SRC_PATH_BIG_VISION}` rather than hardcoded.
Second, should we be linking https://github.com/google-research/big_vision/blob/70b22bb60c46023e50fd0f5d42e0f48f8dfe2fdb/big_vision/requirements.txt into `/opt/pip-tools.d`? Maybe then some of the above hardcoded requirements can be dropped?

rosetta/rosetta/projects/paligemma/Finetune_PaliGemma.ipynb
olupton, line 1: Should we run `nb-clean` on this? Is it desirable to have so much output content embedded in the repo: https://github.com/NVIDIA/JAX-Toolbox/blob/8b7609a2d4a7cc170ef342d8f74cf6a8280f2db2/rosetta/rosetta/projects/paligemma/Finetune_PaliGemma.ipynb ?

rosetta/Dockerfile.gemma
```diff
@@ -0,0 +1,71 @@
...
+ADD ./rosetta/rosetta/projects/paligemma/Finetune_PaliGemma.ipynb ${SRC_PATH_GEMMA}/examples/Finetune_PaliGemma.ipynb
+ADD ./rosetta/rosetta/projects/paligemma/test_gemma.py ${SRC_PATH_GEMMA}/tests/test_gemma.py
+
+###############################################################################
+## Install accumulated packages from the base image and the previous stage
+###############################################################################
+
+FROM mealkit as final
+
+RUN pip-finalize.sh
```
olupton, line 71: Missing newline

.github/container/manifest.yaml
```diff
@@ -150,3 +155,23 @@ grok-1:
...
+flaxformer:
+  url: https://github.com/google/flaxformer.git
+  tracking_ref: main
+  latest_verified_commit: 399ea3a85e9807ada653fd0de1a9de627eb0acde
+  mode: git-clone
+panopticapi:
+  url: https://github.com/akolesnikoff/panopticapi.git
+  tracking_ref: mute
+  latest_verified_commit: a698a12deb21e4cf0f99ef0581b2c30c466bf355
+  mode: git-clone
```
olupton, line 177: Missing newline

rosetta/rosetta/projects/paligemma/README.md
```diff
@@ -0,0 +1,55 @@
...
+You can pull a container that includes JAX and all dependencies needed for this notebook with the following:
+
+```bash
+docker pull nvcr.io/nvidia/jax:gemma
+```
+
+The best way to run this notebook is from within the container. You can do that by launching the container with the following command
+
+```bash
+docker run --gpus all -it --rm -p 8888:8888 nvcr.io/nvidia/jax:gemma bash -c 'source /usr/local/nvm/nvm.sh && jupyter lab'
```
olupton, line 54: What is this `nvm.sh` script and where does it come from?

sbhavani: Sorry that was copied from NGC container docs, sourcing nvm.sh avoids some node error. We don't have that file in our containers. We also have root as the default user so the command should be:

docker run --gpus all -it -p 8888:8888 nvcr.io/nvidia/jax:gemma bash -c 'jupyter lab --allow-root'

rosetta/rosetta/projects/paligemma/Finetune_PaliGemma.ipynb
sbhavani: I think it's fine to keep it rendered as an easy reference since we don't have the content in the README

rosetta/rosetta/projects/paligemma/README.md
```diff
@@ -0,0 +1,55 @@
+# PaliGemma
+
+PaliGemma is a vision language model (VLM). These models are well-suited for a variety of tasks that require visually-situated text understanding and object localization. On a high level, the architecture follows a vision encoder (ViT) that encodes an image into tokens which together with text input (question, prompt, instruction), are passed to a decoder-only transformer (Gemma) that generates a text output.
+
+For more details, refer to the [PaliGemma model card](https://ai.google.dev/gemma/docs/model_card) released by Google.
```
sbhavani, line 5: Correct model card link: https://ai.google.dev/gemma/docs/paligemma/model-card

rosetta/Dockerfile.gemma
DwarKapex: 1. Done
2. Unfortunately, it does not work, b/c Big Vision requirements contains packages we cannot process in our pip-compile approach (for example, `git+https://github.com/google/flaxformer`).

sbhavani (dismissed):
thanks for updating the readme!

rosetta/Dockerfile.gemma
olupton: What's wrong with the second one? We already have explicit handling for pip vcs installations like that?

.github/container/manifest.yaml
olupton: It's still missing 🙃

.DS_Store
olupton: It still seems to be there. Did you forget to push a commit?

.github/workflows/_ci.yaml
```diff
@@ -465,6 +489,30 @@ jobs:
...
+  test-gemma:
+    needs: build-gemma
+    uses: ./.github/workflows/_test_unit.yaml  
+    if: inputs.ARCHITECTURE == 'amd64'
+    with:
+      TEST_NAME: gemma
+      EXECUTE: |
+        docker run --shm-size=1g --gpus all ${{ needs.build-gemma.outputs.DOCKER_TAG_FINAL }} \
+        bash -ec \
+        "pip install -e /opt/gemma[test] && pytest /opt/gemma/ && python /opt/gemma/tests/test_paligemma.py" | tee test-gemma.log
```
olupton, line 502: The pip install seems odd here. I guess the point is to install extra test dependencies?

.DS_Store
DwarKapex: Looks like yes, I was distracted and forgot to push =). Thanks for noticed it.

.github/workflows/_ci.yaml
DwarKapex: Yes ([proof1](https://github.com/google-deepmind/gemma/blob/main/pyproject.toml#L31), [proof2](https://github.com/google-deepmind/gemma?tab=readme-ov-file#running-the-unit-tests)).

olupton (approved):

PR #593, opened 2024-02-27
added maxtext tests to CI

kocchop:
Added maxtext test scripts to CI


terrykong (approved):
Looks like tests ran, but failed b/c the slurm cluster had no space. Since this doesn't change the test content, I think this is fine to merge since this was a coverage gap.

PR #594, opened 2024-02-27
remove NCCL_IB_SL flag from JAX dockerfile

ashors1:


nouiz:
Do we need to also remove some documentation?

ashors1:
> Do we need to also remove some documentation?

Done. Thanks for the reminder

nouiz (approved):

yhtang (approved):

PR #600, opened 2024-02-29
Install latest llvm/clang using script from https://apt.llvm.org/

yhtang:


.github/container/Dockerfile.base
```diff
@@ -37,7 +36,14 @@ apt-get install -y \
   python3-pip \
   rsync \
   vim \
-  wget
+  wget \
+  lsb-release software-properties-common  # for llvm.sh
+
+# Install LLVM/Clang 17
```
andportnoy, line 42: Pedantic nit: the comment says 17 but the code is generic.

andportnoy (dismissed):

.github/container/Dockerfile.base
yhtang: Will fix before merging (if I don't forget :grin:)

PR #601, opened 2024-02-29
Change haliax tracking mode in manifest.yaml

yhtang:
Fixes [build failure](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/8093829709/job/22129756262#step:10:667) in today's nightly.


dlwh:
Is there something I should be doing to make this kind of thing less likely/easier for you all?

yhtang:
> Is there something I should be doing to make this kind of thing less likely/easier for you all?

No immediate action needed as the build error should be fixed now.

In the longer run, should the need arises for including an editable version of haliax in the container, then having VCS specs in `pyproject.toml` such as
```
haliax @ git+https://github.com/stanford-crfm/haliax.git@dev
```
could cause difficulty as pip would regard any local haliax as incompatible with that spec.

Out of curiosity, how do you install the two packages locally without triggering pip to always pull halix from GitHub? Do you use pip at all or maybe some other dependency manager?

dlwh:
i just pip install -e haliax again after installing levanter when I need to do that (which I don't need to do that often)... is the best way to do this to just publish nightly builds of haliax?

yhtang:
> i just pip install -e haliax again after installing levanter when I need to do that (which I don't need to do that often)... is the best way to do this to just publish nightly builds of haliax?

Yes nightly build is much easier to handle since it is not a moving target such as a git branch.

PR #602, opened 2024-03-01
Add flash attention relnote to pax readme

ashors1:
Recommend that users disable FA using the XLA flag `--set_xla_gpu_enable_cudnn_fmha=False` if they observe divergence with 126M

### DIFF
--- a/rosetta/rosetta/projects/pax/README.md
+++ b/rosetta/rosetta/projects/pax/README.md
@@ -126,6 +126,9 @@ Note that packing is currently not supported when using TE. All configs disable 
 ### Native FP8
 Rosetta Pax containers also provide support for native FP8 through XLA. Enabling FP8 can be done by adding the following command-line flag to `paxml/contrib/gpu/scripts_gpu/run_pile_singlenode.sh`: `--fdl.USE_FP8=True`. When using native FP8, TE must be disabled. For a detailed explanation of native FP8 support in Pax, as well as a comparison between native FP8 and TE FP8, please refer to the [NATIVE_FP8](https://github.com/NVIDIA/JAX-Toolbox/blob/main/rosetta/docs/NATIVE_FP8.md) documentation.
 
+### Flash Attention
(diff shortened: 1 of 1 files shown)


ashors1:
Note that there are plans to deprecate this flag: https://github.com/openxla/xla/blob/92dd29b80b6a57087daa6fe88079fc5bb2dbe39e/xla/debug_options_flags.cc#L89-L90 so this may not be a permanent solution

nouiz:
> Note that there are plans to deprecate this flag: https://github.com/openxla/xla/blob/92dd29b80b6a57087daa6fe88079fc5bb2dbe39e/xla/debug_options_flags.cc#L89-L90 so this may not be a permanent solution

FYI: @Ayanmol This option can never be removed.

AyanmoI:
> > Note that there are plans to deprecate this flag: https://github.com/openxla/xla/blob/92dd29b80b6a57087daa6fe88079fc5bb2dbe39e/xla/debug_options_flags.cc#L89-L90 so this may not be a permanent solution
> 
> FYI: @Ayanmol This option can never be removed.

Yes that makes sense.

rosetta/rosetta/projects/pax/README.md
```diff
@@ -126,6 +126,9 @@ Note that packing is currently not supported when using TE. All configs disable
 ### Native FP8
 Rosetta Pax containers also provide support for native FP8 through XLA. Enabling FP8 can be done by adding the following command-line flag to `paxml/contrib/gpu/scripts_gpu/run_pile_singlenode.sh`: `--fdl.USE_FP8=True`. When using native FP8, TE must be disabled. For a detailed explanation of native FP8 support in Pax, as well as a comparison between native FP8 and TE FP8, please refer to the [NATIVE_FP8](https://github.com/NVIDIA/JAX-Toolbox/blob/main/rosetta/docs/NATIVE_FP8.md) documentation.
 
+### Flash Attention
+CuDNN flash attention is now enabled by default via XLA. Divergence has been observed with the GPT 126M model with flash attention enabled. If you observe divergence when running GPT 126M, you can disable flash attention using the following XLA flag: `--set_xla_gpu_enable_cudnn_fmha=False`.
```
nouiz, line 130: Can you add the date when our container has it enabled by default?

nouiz (approved):

PR #609, opened 2024-03-04
Run Pallas tests on JAX containers

olupton:
Now the JAX containers should be capable of this.


yhtang:
Could we also disable the current 'Pallas' build and update the associated README entry?

yhtang (approved):

PR #613, opened 2024-03-04
added PGLE instructions

kocchop:
Added instructions to run with PGLE and also provided the protobuf generation script


kocchop:
Hi @ashors1 , I have moved the readme. I have also pointed to the pgo script directly from jax. Please check and let me know if you have other suggestions.

ashors1 (approved):

rosetta/docs/PGLE.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Recommended XLA Flags
+
+In order to get the best performance with PGLE, here is a list of all recommended XLA flags:
+```
+export XLA_FLAGS="--xla_gpu_enable_latency_hiding_scheduler=true
+--xla_gpu_enable_async_all_gather=true
+--xla_gpu_enable_async_reduce_scatter=true
+--xla_gpu_enable_triton_gemm=false
+--xla_gpu_simplify_all_fp_conversions
```
nouiz, line 70: This list is out dated. @abhinavgoel95 Do you have an up to date list?

rosetta/docs/PGLE.md
```diff
@@ -0,0 +1,89 @@
...
+--xla_gpu_enable_while_loop_double_buffering=true
+--xla_gpu_enable_triton_softmax_fusion=false
+--xla_gpu_enable_all_gather_combine_by_dim=false
+--xla_gpu_enable_reduce_scatter_combine_by_dim=false
+--xla_disable_hlo_passes=rematerialization
+--xla_gpu_pgle_profile_file_or_directory_path=path\to\generated\pbtxt"
+```
+### About Combine thresholds
+
+One last thing to add is regarding the **"combine thresholds"**. Ideally, a higher combining threshold for all-gather, reduce-scatter kernels will ensure the best use of the bandwidth. However, they might also incur some dependencies as XLA would try to combine several async collectives and we might see degradation in overall overlap. On the other hand, using very small combining thresholds would create more number of individual collective ops. This makes overlap easier but might under-utilize the bandwidth. For the best performance, these values may need to be tuned based on the individual model and number of devices used for training.
```
nouiz, line 89: at the end add something like: `and hardware` the differnt kind of GPUs and networking gear used can influance that.

nouiz (commented):
Some small late comments.

PR #618, opened 2024-03-05
Add 9.0a capability to build-jax.sh

andportnoy:
This gets translated to `sm_90a` by bazel and is used to compile Hopper-only kernels such as custom fusions in XLA implemented using CUTLASS.


yhtang (approved):
LGTM

PR #622, opened 2024-03-06
Fix metrics for all MGMN tests

terrykong:
After https://github.com/NVIDIA/JAX-Toolbox/pull/509 was merged, the metrics were broken since there was an unexpected number of hyphens. This obtains the job name more robustly by splitting on the GITHUB_RUN_ID, which all experiment dirs should have.

* Introduces `--output_json_path` to be more explicit
* Fixes rosetta-t5x which still used v3 actions, which caused some issues download artifacts
* Adds size_guidance to the TB EventAccumulator to ensure all steps are loaded


terrykong:
Moved to draft as I am debugging why rosetta tests aren't finding metric artifacts

terrykong:
Found the issue with rosetta-t5x, turns out it was still using v3 actions. I also found that we need to add size_guidance to the TB api, otherwise we can have failures b/c TB tries to be efficient and not load all events

yhtang (approved):

PR #624, opened 2024-03-07
Re-enable environment variables to guard against hangs with Triton gemms

ashors1:
From some local experiments, it appears that setting `JAX_SHARE_BINARY_BETWEEN_HOSTS=True` and `JAX_SHARE_AUTOTUNE_CONFIG_BETWEEN_HOSTS=True` no longer leads to failures. Attempting to re-enable them in the JAX container.

### DIFF
--- a/.github/container/manifest.yaml
+++ b/.github/container/manifest.yaml
@@ -1,31 +1,31 @@
 jax:
   url: https://github.com/google/jax.git
   tracking_ref: main
-  latest_verified_commit: 75cdef7626b92b8b6563ea68ae4747fd6994db2e
+  latest_verified_commit: c94ec17c0d255b560e948e1b6bb98f1a1b8be45f
   mode: git-clone
 xla:
   url: https://github.com/openxla/xla.git
   tracking_ref: main
-  latest_verified_commit: 831e9cef85493ff7ee2e24fd4cc64377d682aecc
+  latest_verified_commit: e025d9f3ed4d5b7ed80b2352c613e3706d917416
   mode: git-clone
 flax:
   url: https://github.com/google/flax.git
   mirror_url: https://github.com/nvjax-svc-0/flax.git
   tracking_ref: main
-  latest_verified_commit: aaf130c90eb46160a3234c258a48bf1b932d7829
+  latest_verified_commit: e4282ee187efbefbde268c6873c592a352f56313
   mode: git-clone
   patches:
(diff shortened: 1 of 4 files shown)


terrykong:
Is that praxis PR update required for this JAX-Toolbox PR?

ashors1:
> Is that praxis PR update required for this JAX-Toolbox PR?

It might not be, but I figured it was easier to update everything since it's unclear whether updating JAX without updating praxis would lead to dependency issues. Also yesterday's rosetta pax tests passed (https://github.com/NVIDIA/JAX-Toolbox/actions/runs/8185876141), so I figured the praxis bump was safe

.github/container/manifest.yaml
```diff
@@ -93,54 +93,54 @@ jestimator:
...
   tracking_ref: main
-  latest_verified_commit: e31af8c1a11f749edeac512f34d148b9933f863f
+  latest_verified_commit: 11706e4a1e01a81ea6b3e02c5ad147028d5b94bb
   mode: pip-vcs
 # used by Pallas
 openxla-triton:
   url: https://github.com/openxla/triton.git
   tracking_ref: llvm-head
-  latest_verified_commit: cl608559313
+  latest_verified_commit: c33bbb01d087129e3714cfc2a0bcd53b80dd2013
```
olupton, line 107: What is the workflow that resulted in this change?
The `openxla-triton` commit is derived from the `xla` commit.
In this case, the old value is still valid: https://github.com/openxla/xla/blob/e025d9f3ed4d5b7ed80b2352c613e3706d917416/third_party/triton/workspace.bzl#L8
The `bump.sh` script should get this right:
https://github.com/NVIDIA/JAX-Toolbox/blob/9c2ae8fcde0cb2e0d14273ed0107634df0cbc5a7/.github/container/bump.sh#L114-L120
so I guess these changes came from something else?

olupton (dismissed):

.github/container/manifest.yaml
ashors1: I did use `bump.sh` to update the manifest. My command was simply
```
bash bump.sh -i manifest.yaml
```

ashors1:
Looks like I was incorrect; the problems that exist with `JAX_SHARE_BINARY_BETWEEN_HOSTS=True` and `JAX_SHARE_AUTOTUNE_CONFIG_BETWEEN_HOSTS=True` are still present. Converting this PR to a draft while we wait for these issues to be resolved.

PR #635, opened 2024-03-12
Draft: Save -status.json (recreated)

CliveUnger:
Adds the exit status json file to the metrics artifact for convenience. Helps get the full picture of a job pass/fail, then metrics all from one artifact. Downloading all artifacts is slow since they are large files.

### DIFF
--- a/.github/workflows/_test_maxtext.yaml
+++ b/.github/workflows/_test_maxtext.yaml
@@ -363,6 +363,7 @@ jobs:
           path: |
             report.jsonl
             *_metrics.json
+            *-status.json
 
   sitrep:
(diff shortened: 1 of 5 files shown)


yhtang (changes requested):
Note that the *status.json files will be retired after #626 gets merged. The plan is to encode such information in a similar file called `sitrep.json`, which are meant to be produced by each and all jobs in the workflow.

CliveUnger:
> Note that the *status.json files will be retired after #626 gets merged. The plan is to encode such information in a similar file called `sitrep.json`, which are meant to be produced by each and all jobs in the workflow.

I see you already merged #626, so I should just close this?

PR #637, opened 2024-03-13
Update profiling.md

olupton:
Add extra wording around `--capture-range-end`.


docs/profiling.md
```diff
@@ -116,6 +116,11 @@ If we then tell `nsys` to listen to the CUDA profiler API, with a command like:
 $ PYTHONPATH=/opt/jax nsys profile --capture-range=cudaProfilerApi --cuda-graph-trace=node --capture-range-end=stop python /opt/jax/examples/mnist_vae.py
 ```
 then the resulting profile will only contain 3 iterations of the loop (5 total - 2 skipped).
+
+With `--capture-range-end=stop`, `nsys` will stop collecting profile data at `cudaProfilerStop()`, but it will not kill the application.
```
nouiz, line 120: Can you say that 'stop' you can't restart it? That was my expectation.

nouiz: Maybe also specify that each restart will be a different nsys report?

nouiz (approved):

PR #641, opened 2024-03-14
Pax2TEConverter support for LLaMa2 checkpoints

denera:
**Changes:**
- New option flags `--pax-split-qkv` and `--te-qkv-layout={qkv_packed,kv_packed}` for identifying QKV layouts for both the Pax source and TE target. Default behavior without new flags remains unchanged.
- New option flag `--use-gated-activations` to support gated GELU in LLaMa2.
- Bias terms are omitted from the checkpoint map when `--weight-only` is used.

**Notes:**
- Changes are verified with a 70b LLaMa2 checkpoint previously converted with @mingxu1067's hard-coded LLaMa2-only converter script.
```
python converter/main.py \
    --input-path=<...> \
    --output-path=<...> \
    --fw=pax --direction=fw2te \
    --num-of-layer=80 \
    --num-of-head=64 \
    --head-dim=128  \
    --mlp-intermediate-dim=28672 \
    --kernel-chunk-size=512 \
    --weight-only \
    --use-gated-activations \
    --pax-split-qkv \
    --te-qkv-layout=kv_packed
```

@ashors1 @sharathts @nluehr @terrykong for viz.

### DIFF
--- a/rosetta/utils/te_pax_t5x_ckpt_converter/README.md
+++ b/rosetta/utils/te_pax_t5x_ckpt_converter/README.md
@@ -112,6 +112,7 @@ python  converter/main.py \
     --head-dim=128 \
     --mlp-intermediate-dim=28672 \
     --kernel-chunk-size=512 \
+    --skip-bias \
     --weight-only \
     --use-gated-activations \
     --pax-split-qkv \
--- a/rosetta/utils/te_pax_t5x_ckpt_converter/README.md
+++ b/rosetta/utils/te_pax_t5x_ckpt_converter/README.md
@@ -2,29 +2,42 @@
 
 ### Arguments
 ```bash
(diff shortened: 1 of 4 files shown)


ashors1:
Could we add some quick documentation to explain how to convert the llama checkpoints? Maybe something along these lines:


"To convert Pax LLaMA checkpoints to TE format, be sure to enable gated activations using the `--use-gated-activations` flag. If converting the 70B model, set `--te-qkv-layout kv_packed`. This accounts for the fact that the `query` and `key/value` projections are unfused when grouped query attention is enabled."

ashors1 (dismissed):

ashors1 (approved):

ashors1 (approved):

PR #649, opened 2024-03-19
Save package URLs to Gist

olupton:
This uploads a tiny markdown file with the latest package link, like this: https://gist.github.com/olupton/7921cf0da6a420f8091b8fabc91fd095#file-final-base-md, and links to it from the README.


yhtang (approved):
Merging as an temporary remedy for the container url issue. Hopefully, we will no longer need it soon once a more thorough solution is landed.

PR #651, opened 2024-03-19
Update T5X baselines and create_baselines script

ashors1:
Note: There is currently a bug in the e2e time which causes reported e2e time to be obsolete. https://github.com/NVIDIA/JAX-Toolbox/pull/654 fixes this bug, but until the baselines are updated accordingly, I have commented out the `test_e2e_time` code block to avoid testing an irrelevant metric. This PR fixes the t5x loss baselines, which are currently causing many t5x tests to fail.


.github/workflows/baselines/download_artifacts.sh
```diff
@@ -15,18 +15,20 @@ fi
 
 [ "$#" -ge "1" ] || usage
 
+EXPECTED_NUM_PAGES=2
```
yhtang, line 18: Is this safe to hard-code this value here?

ashors1: I think it's okay for now because the number of pages will only increase once the number of jobs surpasses 200, and we're only at 107 right now, but this could definitely lead to errors in the future. I will update this to compute the number of pages based on the artifact count

yhtang (approved):
Note that I'm refactoring all of the MGMN jobs and that includes renaming test case names. Let's sync next week to make sure I update the baselines correctly.

terrykong (approved):

PR #658, opened 2024-03-22
Update NCCL version on demand

DwarKapex:
Add ability to build base container with custom NCCL version.
2 main scenarios are under consideration:
1. Custom build with user defined `JAX_NCCL_VERSION`: **libnccl2** and **libnccl-dev** will be installed with the provided version of NCCL if NCCL not in the system or updated if NCCL arrives in base image.
2. Default build (`JAX_NCCL_VERSION` is not defined): JAX_NCCL_VERSION will be assigned to the value of an existing NCCL_VERSION env var of the base image. If NCCL_VERSION is not defined in the base image, then JAX_NCCL_VERSION will be assigned to the version of `libnccl-dev` library if it exists, or left empty.

Installation of NCCL happens if and only if either JAX_NCCL_VERSION or NCCL_VERSION is defined.


.github/container/Dockerfile.base
```diff
@@ -5,13 +5,15 @@ ARG GIT_USER_EMAIL=jax@nvidia.com
...
 ARG CLANG_VERSION=17
+ARG NCCL_VERSION=""
 
 FROM ${BASE_IMAGE}
 ARG GIT_USER_EMAIL
 ARG GIT_USER_NAME
 ARG SRC_MANIFEST_FILE
 ARG DEST_MANIFEST_DIR
 ARG CLANG_VERSION
+ARG NCCL_VERSION
```
yhtang, line 16: We can also override the value of the `NCCL_VERSION` environment variable, which is set in the base image, to avoid confusion.

.github/workflows/_ci.yaml
```diff
@@ -16,6 +16,11 @@ on:
         description: Artifact name in current run w/ manifest/patches. Leaving empty uses manifest/patches in current branch
         default: ''
         required: false
+      NCCL_VERSION:
+        type: string
+        description: NCCL driver version to be installed
```
yhtang, line 21: NCCL is not a driver

.github/workflows/ci.yaml
```diff
@@ -28,6 +28,12 @@ on:
         description: "(used if BUMP_MANIFEST=true) If true: attempt to PR/merge manifest branch"
         default: false
         required: false
+      NCCL_VERSION:
+        type: string
+        description: NCCL driver version to be installed in the format <version>+cuda<version> (for example, 2.20.5+cuda12.3)
```
yhtang, line 33: NCCL is not a driver

yhtang (changes requested):

terrykong (changes requested):

.github/container/Dockerfile.base
```diff
@@ -12,6 +12,15 @@ ARG GIT_USER_NAME
...
 ARG DEST_MANIFEST_DIR
 ARG CLANG_VERSION
+ARG JAX_NCCL_VERSION
+
+###############################################################################
+## Update NCCL version
+###############################################################################
+
+ENV NCCL_VERSION=${JAX_NCCL_VERSION}
+ENV NV_LIBNCCL_DEV_PACKAGE_VERSION=${JAX_NCCL_VERSION}
```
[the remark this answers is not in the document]
DwarKapex: Both `NV_LIBNCCL_DEV_PACKAGE_VERSION` and `NCCL_VERSION` are defined in CUDA base container and they match each other. Here I need to update both of them to keep them consistent and have the same value after any change.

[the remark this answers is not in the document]
terrykong: I see, I didn't realize there were other variables. Looks like there are a few others:

```
NV_LIBNCCL_PACKAGE_NAME=libnccl2
NV_LIBNCCL_PACKAGE_VERSION=2.19.3-1
NCCL_VERSION=2.19.3-1
NV_LIBNCCL_PACKAGE=libnccl2=2.19.3-1+cuda12.3
NV_LIBNCCL_DEV_PACKAGE_NAME=libnccl-dev
NV_LIBNCCL_DEV_PACKAGE_VERSION=2.19.3-1
NV_LIBNCCL_DEV_PACKAGE=libnccl-dev=2.19.3-1+cuda12.3
NCCL_NVLS_ENABLE=0
```
Do these need to be updated as well? In particular NV_LIBNCCL_PACKAGE_VERSION?

.github/workflows/_build_base.yaml
```diff
@@ -105,6 +110,24 @@ jobs:
           driver-opts: |
             image=moby/buildkit:v0.12.1
 
+      - name: Get NCCL version
```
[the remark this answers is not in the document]
DwarKapex: > Is it just so you can populate ENV NCCL_VERSION with the appropriate value at build time?

Yes, this is the purpose of this job.

I will refactor it with `get_remote_env`

.github/workflows/_build_base.yaml
```diff
@@ -57,8 +62,41 @@ permissions:
...
+          if [[ -z $JAX_LIBNCCL_PACKAGE ]]; then
+            BASE_IMAGE=${{ inputs.BASE_IMAGE }}
+            if [[ $BASE_IMAGE == latest ]]; then
+              BASE_IMAGE=$(cat .github/container/Dockerfile.base | sed -n "s/^ARG BASE_IMAGE=\(.*\)$/\1/p")
+            fi
+
+            # try to get NCCL version from provided BASE_IMAGE of ARM64-arch
+            source .github/workflows/scripts/get_remote_env.sh
+            JAX_LIBNCCL_PACKAGE=$(get_remote_env ${BASE_IMAGE:-ubuntu} linux amd64 | jq -r '.[]' | egrep '^NV_LIBNCCL_PACKAGE')
+            JAX_NCCL_VERSION=$(get_remote_env ${BASE_IMAGE:-ubuntu} linux amd64 | jq -r '.[]' | egrep '^NCCL_VERSION=' | cut -d= -f2-)
```
terrykong, line 90: Is `ubuntu` the right default? Does the ubuntu base even have NCCL installed? I feel like there should be a check like
```
if [[ -z "$BASE_IMAGE" ]]; then
  echo "Need to pass non-empty BASE_IMAGE variable"
  exit 1
fi
```
earlier (maybe above line 83)

.github/container/install-nccl.sh
```diff
@@ -5,31 +5,43 @@ set -ex -o pipefail
...
     echo "Could not find compatible NCCL version for CUDA ${cuda_version}"
     exit 1
 fi
 
-apt-get install -y \
+apt-get install -y --allow-change-held-packages \
     libnccl2=${libnccl2_version} \
     libnccl-dev=${libnccl_dev_version}
 
+dpkg -s libnccl2 libnccl-dev
```
terrykong, line 44: Is this a smoke test? Is this being done because the apt-get above doesn't print the version?

DwarKapex: Sort of smoke test. And you're right, this is done for debugging purposes: to have all information about install packages in one place.

.github/workflows/_build_base.yaml
terrykong:
```suggestion
            JAX_LIBNCCL_PACKAGE=$(get_remote_env $BASE_IMAGE linux amd64 | jq -r '.[]' | egrep '^NV_LIBNCCL_PACKAGE')
            JAX_NCCL_VERSION=$(get_remote_env $BASE_IMAGE linux amd64 | jq -r '.[]' | egrep '^NCCL_VERSION=' | cut -d= -f2-)
```

How about removing this too since `ubuntu:latest` doesn't seem to have what we need in the image?

DwarKapex: I removed it, but forgot to push. Done.

PR #672, opened 2024-04-01
Added TCPx check when running on GCP

chaserileyroberts:
This should help direct users on GCP toward a better solution.


.github/container/gcp-tcpx-check.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+if [ -z "${DISABLE_TCPX_CHECK}" ]; then
```
yhtang, line 3:
```
if [ "${DISABLE_TCPX_CHECK}" == 1 ]; then 
  exit 0
fi
```
may be better since it narrows the scope under the branching condition.

.github/container/gcp-tcpx-check.sh
```diff
@@ -0,0 +1,36 @@
...
+# Colors
+YELLOW='\033[0;33m'
+NOCOLOR='\033[0m'
+
+# Paths
+FILE="libnccl.so"
+DIR1="/var/lib/tcpx/lib64"
```
yhtang, line 10: How about a PATH-style format as a more extendable solution?
```
TCPX_PLUGIN_PATH="$TCPX_PLUGIN_PATH:/var/lib/tcpx/lib64:/usr/local/tcpx/lib64"  # allow custom path to be added from outside script
...
for path in $(echo $TCPX_PLUGIN_PATH | tr ':'); do
...
done
```

.github/container/gcp-tcpx-check.sh
```diff
@@ -0,0 +1,36 @@
...
+DIR2="/usr/local/tcpx/lib64"
+
+# Attempt to retrieve the instance ID from the GCP metadata server
+INSTANCE_ID=$(curl  -m 1 -s -H "Metadata-Flavor: Google" "http://metadata.google.internal/computeMetadata/v1/instance/id" || true)
```
yhtang, line 14: Can we use the existence of the metadata server as the predicate here?

.github/container/gcp-tcpx-check.sh
```diff
@@ -0,0 +1,36 @@
...
+
+if [ -n "$INSTANCE_ID" ]; then
+    if [ -f "$DIR1/$FILE" ]; then
+        :
+    elif [ -f "$DIR2/$FILE" ]; then
+        :
+    else
+            echo -e "${YELLOW}
+WARNING: It looks like you are running on GCP, but we did not find libnccl.so
+at either /var/lib/tcpx/lib64 or /usr/local/tcpx/lib64. You main end up defaulting
```
yhtang, line 24: Avoid hardcoding (and repeating) the paths here.

yhtang (changes requested):

.github/container/Dockerfile.base
```diff
@@ -6,6 +6,17 @@ ARG SRC_MANIFEST_FILE=manifest.yaml
 ARG DEST_MANIFEST_DIR=/opt/manifest.d
 ARG CLANG_VERSION=17
 
+###############################################################################
+## Install TCPx for running on GCP
+###############################################################################
+
+FROM us-docker.pkg.dev/gce-ai-infra/gpudirect-tcpx/nccl-plugin-gpudirecttcpx as tcpx_installer
+RUN install --install-nccl
```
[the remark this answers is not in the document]
yhtang: This runs in a separate container stage, so it shouldn't matter whether it installs NCCL there. Later on, we copy the tcpx plugin into the final container.

yhtang (commented):
Shall we expect subsequent PRs for:
- adding the NCCL sanity check
- locality-aware detection of NICs
?

chaserileyroberts:
I can try and do a follow up of automatically detecting nics (or setting known values for certain vm configs), but doing a nccl test on every startup doesn't sound like a good idea to me. We don't even know if users will have multiple GPUs at that time.

.github/container/Dockerfile.base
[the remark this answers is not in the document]
chaserileyroberts: Yeah, what Yu-Hang said. Even if it does reinstall all of NCCL, we are only grabbing the tcpx pluggin and discarding the rest.

.github/container/gcp-tcpx-check.sh
chaserileyroberts: What do you mean? Like just check if `google.internal` gets a dns hit? Probably a good idea actually...

chaserileyroberts:
One thing I want to try before merging this in is setting up two nodes from scratch and only relying on this container.

.github/container/gcp-tcpx-check.sh
```diff
@@ -0,0 +1,33 @@
...
+    and is located at /usr/local/tcpx in this container. 
+    
+    However, there are additional steps you will need to take. Mainly, you'll need to run a 
+    separate receive-datapath-manager daemon on each of your nodes, and correctly configure your 
+    networks and NICs.
+
+    For more information, please see the guide at: 
+    https://cloud.google.com/compute/docs/gpus/gpudirect#provide-access
+
+    (To disable this message, set DISABLE_TCPX_CHECK=1)
```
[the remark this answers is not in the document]
chaserileyroberts: Hm good point. I think originally I wanted to just disable the message, but now I think it might be more important to give users a way to just turn off the setup entirely if it ends up breaking things for them. 

I'll keep the if statement where it is, but I'll update the text of the message.

.github/container/gcp-tcpx-check.sh
chaserileyroberts: Done.

.github/container/gcp-tcpx-check.sh
chaserileyroberts: Done.

yhtang (dismissed):

PR #679, opened 2024-04-03
[maxtext] fixed the attention and default model config

kocchop:
Fixes TE flash attention in maxtext's default model's `quantize_kvcache` config. This introduces a patch that is applied on top of upstream maxtext

### DIFF
--- a/.github/container/Dockerfile.maxtext.amd64
+++ b/.github/container/Dockerfile.maxtext.amd64
@@ -37,5 +37,6 @@ ADD test-maxtext.sh /usr/local/bin
 FROM mealkit as final
 
 RUN pip-finalize.sh
+RUN pip install typing-extensions==4.10.0
 
 WORKDIR ${SRC_PATH_MAXTEXT}
--- a/.github/container/Dockerfile.maxtext.amd64
+++ b/.github/container/Dockerfile.maxtext.amd64
@@ -40,6 +40,5 @@ ADD test-maxtext.sh /usr/local/bin
 FROM mealkit as final
 
 RUN pip-finalize.sh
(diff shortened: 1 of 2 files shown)


kocchop:
Hi @terrykong , the build and tests are now working. Please take a look

.github/container/Dockerfile.maxtext.amd64
```diff
@@ -30,5 +40,6 @@ ADD test-maxtext.sh /usr/local/bin
 FROM mealkit as final
 
 RUN pip-finalize.sh
+RUN pip install typing-extensions==4.10.0
```
terrykong, line 43: It is odd that you need to do it as it defeats the purpose of using `pip-compile`. Can you elaborate the conflict?

Also, if this strict pinning is truly the version you need, then it should go in `/opt/pip-tools.d/requirements-maxtext.in`. If for some reason it causes a conflict because you put it in there, we should resolve that

kocchop: @terrykong I removed the `typing-extensions` pin. It should be fine with the current container config

terrykong: Just commenting about the resolution: the typing-extensions pin wasn't needed b/c pip-compile found the right version. @kocchop found that it was needed in building maxtext in a non-jax-toolbox dockerfile

terrykong (approved):

terrykong:
Looks like I'm not able to merge, I'll try again later, but if someone beats me to it, please give it a try!

image

olupton:
I see the same error when trying to merge.

terrykong:
Thanks @olupton !

PR #683, opened 2024-04-04
External test trigger

DwarKapex:
_test_unit.yaml does not accept `IMAGE` as a parameter.


terrykong (approved):

yhtang (changes requested):
Why does the changeset include maxtext stuff? Is your branch based off main or 24.04-devel?

yhtang (changes requested):
Please rebase it on top of `24.04-devel`

PR #674, opened 2024-04-01
Add GPU performance tips for high performance LLMs with JAX/XLA

instinct79:
Documenting all JAX and XLA flags that we use for high performance LLMs with JAX/XLA.

Will update this periodically, and also link to 
- https://github.com/google/jax/blob/main/docs/gpu_performance_tips.md

Per Yu-Hang's suggestion, will update https://github.com/NVIDIA/JAX-Toolbox/blob/main/README.md#environment-variables to this page as well, once the PR is merged.

PR #684, opened 2024-04-04
Update front-page readme with link to XLA flag doc

terrykong:
Followup for #674 that links to this new document

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ We will update this table as new models become available, so stay tuned.
 
 ## Environment Variables
 
-The [JAX image](https://github.com/NVIDIA/JAX-Toolbox/pkgs/container/jax) is embedded with the following flags and environment variables for performance tuning:
+The [JAX images](https://github.com/NVIDIA/JAX-Toolbox/pkgs/container/jax) are embedded with the following flags and environment variables for performance tuning:
 
 | XLA Flags | Value | Explanation |
 | --------- | ----- | ----------- |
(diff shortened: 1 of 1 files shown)


instinct79:
Should we delete all xla flags from the front page readme, and link to this document and maybe the scripts for PAX or T5X for benchmark specific flags ?

nouiz:
> Should we delete all xla flags from the front page readme, and link to this document and maybe the scripts for PAX or T5X for benchmark specific flags ?

We need to tell which flag are enabled in the container. But maybe we can just list them and tell to see that page for their description?

instinct79:
I think we will point to our scripts that enable these flags. Otherwise, there are duplicate documentation. Perhaps, the environment variables can stay here.

terrykong:
Since the flags in there can change over time, what about instead including instructions to view the flags. Example:

```
docker run --rm quay.io/skopeo/stable inspect docker://ghcr.io/nvidia/jax:upstream-pax | jq -r '.Env[]' | grep '^XLA_FLAGS='

# which returns

XLA_FLAGS= --xla_gpu_enable_latency_hiding_scheduler=true --xla_gpu_enable_async_all_gather=true --xla_gpu_enable_async_reduce_scatter=true --xla_gpu_enable_triton_gemm=false
```

yhtang:
> Since the flags in there can change over time, what about instead including instructions to view the flags. Example:
> 
> ```
> docker run --rm quay.io/skopeo/stable inspect docker://ghcr.io/nvidia/jax:upstream-pax | jq -r '.Env[]' | grep '^XLA_FLAGS='
> 
> # which returns
> 
> XLA_FLAGS= --xla_gpu_enable_latency_hiding_scheduler=true --xla_gpu_enable_async_all_gather=true --xla_gpu_enable_async_reduce_scatter=true --xla_gpu_enable_triton_gemm=false
> ```

I would consider it a less-readable approach as a reader will have to break from the mental flow and do scripting to obtain these flags.

A more user-friendly but also much heavier solution is to set up a proper docs building process (e.g. readthedocs) so that the docs page is automatically populated with the flags set in the containers.

instinct79:
Maybe provide a sample output from one of our latest containers for now ? The flags are not going to change so rapidly and will atleast provide the users some idea of what is enabled for our benchmarks.

yhtang:
> Maybe provide a sample output from one of our latest containers for now ? The flags are not going to change so rapidly and will atleast provide the users some idea of what is enabled for our benchmarks.

Good point. @terrykong what do you think?

terrykong:
Done 👍

PR #686, opened 2024-04-05
Adds bump*.sh & create-distribution.sh into base container

terrykong:
This helps the base container be self-contained since it now has all the
scripts and patches to build downstream containers like the rosetta
containers.

Prior to this change, if you wanted to build a rosetta container, you
would need to check out JAX-Toolbox and use these build scripts at a
commit that may not align with the base container.

After this change we can use the container to build the rosetta
container. As an example, you would be able to do something like this:

```sh
id=$(docker create ghcr.io/nvidia/jax:upstream-pax)
docker cp $id:/opt /local

bash /local/bump.sh ...

docker buildx build --build-context jax-toolbox=/local -f /local/rosetta/Dockerfile.pax
```

## Note:
mirrors #685


DwarKapex:
LGTM

terrykong:
Closing since #685 will just be cherry-picked in

PR #699, opened 2024-04-08
fix patch for upstream maxtext

kocchop:
Most of the patch change PRs got merged to maxtext main. So removing them from patch. Only change require now is the tf pin

### DIFF
--- a/.github/container/manifest.yaml
+++ b/.github/container/manifest.yaml
@@ -114,7 +114,7 @@ jax-triton:
 maxtext:
   url: https://github.com/google/maxtext.git
   tracking_ref: main
-  latest_verified_commit: 02d83aa564df985512b2e3b110c3deae5b89c2c9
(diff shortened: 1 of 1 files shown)


terrykong (changes requested):
Does the changes to the dockerfile need to be reverted as well? https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/container/Dockerfile.maxtext.amd64#L23

olupton:
> Does the changes to the dockerfile need to be reverted as well? https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/container/Dockerfile.maxtext.amd64#L23

Note that the patch file is not being deleted here, there is just a large block being removed from the top of it.

kocchop:
@terrykong @olupton I bumped the maxtext commit id in `manifest.yaml`. We still need patch for the tensorflow pin.

olupton (approved):

terrykong (approved):

PR #700, opened 2024-04-09
Fix tensorflow-text source build on LLVM 17

gspschmid:
Yesterday's update from LLVM 14 to LLVM 17 broke the source builds of tensorflow-text on arm64 ([example](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/8598999090/job/23561812951#step:10:861)).

~~Let's see if this is resolved by upgrading to a newer version (2.13.0 -> 2.16.1). (We might even just want to track main instead?)~~
Actually, bumping tensorflow-text from 2.13.0 -> 2.16.1 doesn't work, because it's in sync with tensorflow, and lingvo (used in pax) is only compatible with 2.13.0 (even on the [most recent main](https://github.com/tensorflow/lingvo/blob/05a076b0783a8bbf4a770095966c472bb37bbf65/WORKSPACE#L8)). Luckily, merely disabling the `--no-undefined-version` check in `lld` alone seems to do the trick.

### DIFF
--- a/.github/container/Dockerfile.pax.arm64
+++ b/.github/container/Dockerfile.pax.arm64
@@ -18,15 +18,26 @@ RUN wget https://github.com/bazelbuild/bazelisk/releases/download/v1.17.0/bazeli
     chmod a+x /usr/bin/bazel
 
 #------------------------------------------------------------------------------
-# build tensorflow-text 2.13.0 from source
+# build tensorflow-text from source
 #------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


gspschmid:
@olupton This should be good to go (pax fixed; t5x still broken, but now for unrelated reasons)

olupton (approved):
LGTM; it's a shame we need this...

yhtang (approved):

PR #704, opened 2024-04-09
nsys-jax: add profiling wrapper script.

olupton:
This wraps `nsys profile` and writes a .zip archive containing additional information for offline analysis of JAX/XLA profiles:
- All Python source files referred to in the JAX/XLA metadata.
- XLA metadata exported in .pb and .pbtxt format.
- .proto schema files needed to read the above.

In addition, the .nsys-rep file is compressed inside the .zip archive, and a .parquet format export of the NVTX trace is included.

The wrapper also ensures that the metadata does not truncate Python stack traces, and enables CUDA graph node tracing -- although at present disabling CUDA graphs is recommended to avoid limitations in metadata correlation.


.github/container/install-nsight.sh
```diff
@@ -30,3 +30,5 @@ if [[ "${UBUNTU_ARCH}" == "amd64" && -d "${NSYS202411}" ]]; then
   ln -s "${LIBCUPTI123}" "${NSYS202411}/target-linux-x64/libcupti.so.12.3"
   mv "${NSYS202411}/target-linux-x64/libcupti.so.12.4" "${NSYS202411}/target-linux-x64/_libcupti.so.12.4"
 fi
+
+ln -s $(dirname $(realpath $(command -v nsys)))/python/packages/nsys_recipe/requirements/common.txt /opt/pip-tools.d/requirements-nsys-recipe.in
```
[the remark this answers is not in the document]
olupton: Done

.github/container/jax_nsys/install.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
olupton: This is for the post-processing part (i.e. potentially on a laptop, workstation, ...). It creates a new virtual environment in which to run Jupyter Lab, yes.

I added a comment block to the top of the file, and also added a section here describing the new wrapper/workflow: https://github.com/NVIDIA/JAX-Toolbox/blob/olupton/nsys-wrapper/docs/profiling.md#nsys-jax-wrapper-for-nsight-systems

.github/container/Dockerfile.base
```diff
@@ -54,6 +54,7 @@ apt_packages=(
   liblzma-dev
   python-is-python3
   python3-pip
+  python3-venv
```
yhtang, line 57: This could conflict with pip-installed venv packages. Can we use pip to install it? The rule of thumb here is to install with pip as much as possible and avoid distro-repos.

.github/container/nsys-jax-gather-src-files
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+set -e -o pipefail
+
+# Handle commandline arguments
+proto_dir="$1"
+shift
+
+# Work in a temporary directory
+tmp_dir=$(mktemp -d --tmpdir protoc-install.XXXXXXXX)
+pushd "$tmp_dir" 1>&2
```
yhtang, line 10: Is there a way to globally redirect stdout to stderr?

.github/container/jax_nsys/install.sh
```diff
@@ -0,0 +1,31 @@
...
+  exit 1
+fi
+VIRTUALENV="${PWD}/nsys_jax_venv"
+if [[ ! -d "${VIRTUALENV}" ]]; then
+  ${PYTHON} -m venv nsys_jax_venv
+  . ./nsys_jax_venv/bin/activate
+  ${PYTHON} -m pip install -U pip
+  ./nsys-jax-ensure-protobuf
+  ${PYTHON} -m pip install jupyterlab numpy pandas protobuf pyarrow
+  curl -o "${VIRTUALENV}/bin/flamegraph.pl" https://raw.githubusercontent.com/brendangregg/FlameGraph/master/flamegraph.pl
```
yhtang, line 25: Do we want to ship the flamegraph binary in the system path instead of downloading it every time?

yhtang (changes requested):

.github/container/jax_nsys/install.sh
olupton: This `install.sh` script is embedded in the output `.zip` archive and is intended to be run on the end-user's system, not inside the containers we build.

.github/container/nsys-jax-gather-src-files
olupton: Note that it's not global; the Python command with the `EOL` delimiters prints to stdout.

.github/container/Dockerfile.base
olupton: OK, I switched to `virtualenv` both for the parts that run inside the container and the script (`install.sh`) that the user runs on their machine.

yhtang (approved):

PR #707, opened 2024-04-10
Fix pax commits to known good versions

ashors1:
Commits were from last known working nightly on 3/14 where rosetta-pax was all green: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/8278396687/job/22659250931


terrykong:
Failure with arm is due to h5py 3.11 (released very recently) not having arm pre-built wheels. I'll address in a subsequent PR

terrykong (approved):

PR #712, opened 2024-04-11
Combine all workflow metrics into single summary (Recreated)

CliveUnger:
Combine sitrep, status, and metric_summary for each test into a single summary json at the end.

The goal is capturing a complete picture of a Workflow Run: each build success/fail, each test success/fail, the metrics for each test and also metadata like what image the tests ran on and such. This data should be all pulled from a single artifact file.


yhtang (changes requested):
Thanks for this PR to make the CI result more usable by other automated systems!

However, I noticed that the human-readable summary section (e.g. [this](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/8686774674/attempts/1#summary-23834304593)) of the CI run is no longer there. Although the section looks a bit ugly at this time, I would still wish we could improve on it instead of dropping it.

Could you please add it back, potentially using the newly combined metrics?

CliveUnger:
Fixed the comment about human readable summary, now this will create a new file instead of replacing an existing one.

.github/workflows/_sandbox.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
           a smooth integration process once the changes are ready to be merged.
 
           Usage:
-          
+
```
olupton, line 28: If you're pushing another change anyway, revert changes to this file.

.github/workflows/_sitrep_mgmn.yaml
```diff
@@ -142,10 +142,20 @@ jobs:
 
           echo "STATUS=${status}" >> ${GITHUB_OUTPUT}
 
+      - name: Check and display metrics summary
```
olupton, line 145: I'm a bit puzzled because it seems like this .json file is not JSON?
https://github.com/NVIDIA/JAX-Toolbox/actions/runs/8917232922/job/24497043508?pr=712#step:7:1

CliveUnger: Hmm yea that is clearly a markdown file (also I have a typo, metric_summary.json instead of metrics_summary.json).

I made an update to write both a markdown and json file.

olupton (approved):

PR #709, opened 2024-04-11
Fixes Pax & T5x ARM build by adding packages needed to build h5py

terrykong:

PR #713, opened 2024-04-12
Fixes Pax & T5x ARM build by adding packages needed to build h5py

terrykong:
Copy of #709 but to main instead of 24.04

### DIFF
--- a/.github/container/Dockerfile.base
+++ b/.github/container/Dockerfile.base
@@ -22,21 +22,28 @@ export DEBIAN_FRONTEND=noninteractive
 export TZ=America/Los_Angeles
 apt-get update
 
-apt-get install -y \
-  bat \
-  build-essential \
-  checkinstall \
-  cmake \
-  curl \
-  git \
-  gnupg \
-  liblzma-dev \
-  python-is-python3 \
-  python3-pip \
-  rsync \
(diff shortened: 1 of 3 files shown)


.github/container/Dockerfile.base
```diff
@@ -22,21 +22,28 @@ export DEBIAN_FRONTEND=noninteractive
...
+  python3-pip
+  rsync
+  vim
+  wget
+  # llvm.sh
+  lsb-release software-properties-common
+  # h5py: The newest release of of h5py (3.11.0) does not include ARM wheels and causes pip to build h5py.
+  #       These installs ensure that 3.11.0, or any future version missing ARM, can be built.
+  #       Related: https://github.com/h5py/h5py/issues/2408
+  pkg-config libhdf5-dev
```
yhtang, line 44: Do we need it in the base image if they are only used by the pax/t5x arm64 dockerfiles?

yhtang (changes requested):

.github/container/Dockerfile.base
terrykong: No, it's only an issue for ARM. I'll make it platform dep

yhtang (approved):

PR #716, opened 2024-04-12
Automatically config NICs used by NCCL on GCP

yhtang:
Each A3 instance on GCP has 5 NICs. The first one is attached to the same PCI bridge with 8 NVMe disks and is hence best for I/O. Each of the other 4 NICs is attached to the same PCI bridge with 2 GPUs, and is hence best for inter-GPU/inter-node communications.

This information needs to be passed into NCCL so that it uses the correct NICs in order to achieve best performance.

This script discovers this information by inspecting the hardware topology of the node using `hwloc`.

Note that this PR does not contain additional NCCL finetuning variables, which are needed for max perf and will be added via a subsequent PR.

Example output from the autoconfig script:
```
[GCP autoconfig slurm0-a3-ghpc-0] NCCL_SOCKET_IFNAME=enp0s12 NCCL_GPUDIRECTTCPX_CTRL_DEV=enp0s12 NCCL_GPUDIRECTTCPX_SOCKET_IFNAME=enp6s0,enp12s0,enp134s0,enp140s0
```


.github/container/gcp-autoconfig.sh
```diff
@@ -0,0 +1,59 @@
...
+      jq -r '.. | objects | select(if .info | type == "array" then .info[1].value | contains("Ethernet") else false end ) | .object.name'
+    )
+    export NCCL_GPUDIRECTTCPX_CTRL_DEV=$NCCL_SOCKET_IFNAME
+    # find NICs that are NUMA-aligned with the NVIDIA GPUs
+    export NCCL_GPUDIRECTTCPX_SOCKET_IFNAME=$(
+      lstopo  --no-caches --no-smt --filter core:none --of xml |\
+      yq -p=xml --xml-attribute-prefix='' --xml-skip-directives --xml-skip-proc-inst -o=json |\
+      jq -c '.. | objects | select(.type == "Bridge" and (.object | type == "array"))' |\
+      grep -i nvidia |\
+      jq -rs '[.. | objects | select(if .info | type == "array" then .info[1].value | contains("Ethernet") else false end) | .object.name] | join(",")'
```
chaserileyroberts, line 51: This all completely looks like black magic to me, so I trust that this is correct.

.github/container/gcp-autoconfig.sh
```diff
@@ -0,0 +1,59 @@
...
+the plugin at ${TCPX_LIBRARY_PATH}, which is already added to LD_LIBRARY_PATH by
+this script $2.
+
+For A3 instances, this script also automatically determines the best NICs to use
+by NCCL for maximum performance.
+
+However, there are additional steps you will need to take. Mainly, you'll need
+to run a separate receive-datapath-manager daemon on each of your nodes, and
+correctly configure your networks and NICs. For more information, please see the
+guide at: https://cloud.google.com/compute/docs/gpus/gpudirect#provide-access
```
chaserileyroberts, line 27: We should also info users that if they are running on DGX, this is already done for them.

chaserileyroberts (approved):
Looks great! Thanks for getting this automated fully Yu-Hang. Happy to hopefully never how to think about this again. :)

.github/container/gcp-autoconfig.sh
yhtang: I'll figure out a way to add some comments. It's a bit tricky inside the `$(...)` block to do so.

.github/container/gcp-autoconfig.sh
yhtang: Issue created for follow-up: https://github.com/NVIDIA/JAX-Toolbox/issues/722

PR #721, opened 2024-04-16
Re-enable t5 in pax arm64 build

yhtang:


nouiz (approved):
LGTM. Did you manually verify that the build works and there isn't any glitch?
Doing the install later vs here could trigger issues.

yhtang:
> LGTM. Did you manually verify that the build works and there isn't any glitch? Doing the install later vs here could trigger issues.

CI looks all good. Do you have any additional concerns?

PR #730, opened 2024-04-17
Revert XLA PR #9778 in 24.04

yhtang:
Reverts https://github.com/openxla/xla/pull/9778 to fix the paxml flash attention convergence issue.


ashors1 (approved):
Assuming CI shows that loss issues are fixed, LGTM

PR #732, opened 2024-04-18
Enable TE flash attention by default in Pax container

ashors1:


yhtang (approved):
Approving but let's don't merge until #730 and #728 lands so that this can be properly tested.

PR #738, opened 2024-04-18
feat: Add rosetta-maxtext

ko3n1g:


.github/container/manifest.yaml
```diff
@@ -113,9 +113,12 @@ jax-triton:
   mode: git-clone
 maxtext:
   url: https://github.com/google/maxtext.git
+  mirror_url: https://github.com/nvjax-svc-0/maxtext.git
   tracking_ref: main
   latest_verified_commit: 78daad198544def8274dbd656d122fbe6a0e1129
   mode: git-clone
+  patches:
+    mirror/patch/test_rosetta_maxtext: file://patches/maxtext/mirror-patch-rosetta-maxtext.patch
```
terrykong, line 121: Just leaving a reminder that this can be cleaned up if not needed

.github/workflows/_sandbox.yaml
```diff
@@ -1,41 +1,150 @@
-name: "~Sandbox"
-
+name: Sandbox
```
terrykong, line 1: Just a reminder to revert before merging

.gitignore
```diff
@@ -0,0 +1 @@
+local/
```
terrykong, line 1: Is this a debug line? Does it need to be committed?

rosetta/Dockerfile.maxtext
```diff
@@ -0,0 +1,77 @@
+# syntax=docker/dockerfile:1-labs
+ARG BASE_IMAGE=ghcr.io/nvidia/jax-mealkit:upstream-pax
```
terrykong, line 2: Should this be
```suggestion
ARG BASE_IMAGE=ghcr.io/nvidia/jax-mealkit:upstream-maxtext
```

rosetta/Dockerfile.maxtext
```diff
@@ -0,0 +1,77 @@
...
+if [[ "${UPDATE_PATCHES}" != "true" && "${UPDATE_PATCHES}" != "false" ]]; then
+  echo "UPDATE_PATCHES can only be true or false"
+  exit 1
+fi
+if [[ "${UPDATE_PATCHES}" == "true" ]]; then
+  cp -r /mnt/jax-toolbox/.github/container/patches ${MANIFEST_DIR}/
+  cp /mnt/jax-toolbox/.github/container/manifest.yaml ${MANIFEST_DIR}/manifest.yaml
+  cp /mnt/jax-toolbox/.github/container/create-distribution.sh ${MANIFEST_DIR}/create-distribution.sh
+  # TODO: remove
+  cp /mnt/jax-toolbox/.github/container/pip-finalize.sh /usr/local/bin/
```
terrykong, line 39: I think it's okay to remove this moving forward. Initially I added it since `pip-finalize.sh` was in flux, but it's stabilized

.github/workflows/_ci.yaml
```diff
@@ -171,7 +180,7 @@ jobs:
       - build-jax
       - build-triton
       - build-equinox
-      - build-maxtext
+      - build-upstream-maxtext
```
terrykong, line 183: I think we want to publish both so two rows, similar to how we have `build-rosetta-pax` and `build-upstream-pax` here.

.github/workflows/_ci.yaml
```diff
@@ -91,6 +90,16 @@ jobs:
       DOCKERFILE: .github/container/Dockerfile.maxtext.amd64
```
terrykong, line 90: I think CONTAINER_NAME should be upstream-maxtext to match the other rosetta frameworks

terrykong (changes requested):
Thanks for the contribution! I left a few comments

.gitignore
ko3n1g: We can remove it. I found it convenient to have local testing directory which isn't checked into git.

rosetta/Dockerfile.maxtext
ko3n1g: I guess we can then also we also remove it from `rosetta-t5x` and `rosetta-pax`?

terrykong: Yup

.github/container/manifest.yaml
kocchop: I could not see the patch file in the repo. Are we using it?

rosetta/Dockerfile.maxtext
kocchop, line 1: Why are we again building TE here? The base image should be `jax-mealkit:jax`, right? And then we apply patch and build the final

.github/workflows/_ci.yaml
kocchop, line 1: we did not run the tests for rosetta-maxtext yet, right? we should check the validity of the rosetta build

kocchop (changes requested):
Left a few comments, please take a look

PR #739, opened 2024-04-18
Fix jax and jaxlib version string for 24.04

yhtang:
So that they appear as `jax-0.4.26` and `jaxlib-0.4.26` instead of jax-0.4.26.dev20240404+165361ee8 and jaxlib-0.4.26.dev20240418.


yhtang:
Manual verification: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/8744507879/job/23997723048#step:12:627

PR #748, opened 2024-04-22
ci: Rename `_build` to `_build_upstream`

ko3n1g:
To improve consistency of our naming conventions


yhtang:
`_build.yaml` is meant to be generic --- it doesn't care if the build is for an 'upstream' version of some software or not. Hence I don't think it is useful to rename it which hints at specializing its purpose.

PR #750, opened 2024-04-22
Install ray on all images

chaserileyroberts:
Install's the [Ray](https://github.com/ray-project/ray) library on all of the base images.


yhtang (changes requested):
Please pin a specific version so that we can have reproducible builds.

The best way to do it consistently is to dump it to a `requirements-ray.in` file and then let `pip-finalize` handle it. This way, the dependencies are handled together with all other packages that we want to install.

PR #751, opened 2024-04-22
Check compute capability at runtime

DwarKapex:
It was noticed, that if you run jax on mixture of different GPUs, it may fail if run at least `pmap()` with the following error:

```
E0418 23:30:54.309306    2770 pjrt_stream_executor_client.cc:2809] Execution of replica 4 failed: INVALID_ARGUMENT: executable is built for device CUDA:0 of type "NVIDIA A100-SXM4-80GB"; cannot run it on device CUDA:4 of type "NVIDIA T1000"
jax.errors.SimplifiedTraceback: For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.
```

I added a docker runtime check for this case: it checks that all available and/or specified thru CUDA_VISIBLE_DEVICES GPUs are identical


DwarKapex:
I'm only sure about `jax.pmap()` function. Maybe there are more to include.

.github/container/check-cc.sh
```diff
@@ -0,0 +1,20 @@
...
+    YELLOW='\033[0;33m'
+    NOCOLOR='\033[0m'
+
+    echo -e "${YELLOW}
+
+WARNING: If you're going to use jax.pmap(), be aware that it requires all of the 
+participating devices to be identical. You can filter in the identical GPUs by using
+the following cmd:  
+
+    $ docker run ... --gpus=0,1,2,...,N  ...
```
nouiz, line 16: This fix doesn't always work. If the user do: `docker --privileged` then all the GPUs will be mounted even if you try to limit it like this...
CUDA_VISIBLE_DEVICES seem to always work to my knowledge.

yhtang (commented):
Instead of relying on JAX, can we go from first principles and use nvidia-smi to query for GPU specs? Even a simple check of GPU name string should help.

yhtang:
> > Instead of relying on JAX, can we go from first principles and use nvidia-smi to query for GPU specs? Even a simple check of GPU name string should help.
> 
> Note that `nvidia-smi` does not respect `CUDA_VISIBLE_DEVICES`, so it might take some effort to implement a warning that goes away when `CUDA_VISIBLE_DEVICES` is set correctly.
> 
> Seems to me that giving a better error when an unsupported combination of GPUs is visible is something that belongs upstream, as an extension of [google/jax@9fff9ae](https://github.com/google/jax/commit/9fff9aeb69f586d36bdb6422f40b548aa41bd1e3).

The warning doesn't have to go away, since CUDA_VISIBLE_DEVICES may either be enforced at container scope or application scope. We can just add wordings like "if you already set it, please ignore this message"...

nouiz:
> That's a good point, but I'm not sure how much length upstream would like to go in instructing users about container and NVIDIA variable specifics 🤣

We don't need to do that. Just warn. This isn't frequent. They could just point to CUDA_VISIBLE_DEVICES. It can be simple.

sitrep_0.zip
gspschmid, line 1: Is this zip intended to be part of the PR?

.github/container/check-gpus.sh
```diff
@@ -0,0 +1,84 @@
...
+    diff_gpu_count=$(nvidia-smi --query-gpu=gpu_name --format=csv,noheader | sort | uniq | wc -l)
+    if [[ $diff_gpu_count -ge 2 ]]; then
+            echo -e "${YELLOW}
+
+WARNING: There are $diff_gpu_count different GPUs in the system:
+
+$(nvidia-smi --query-gpu=gpu_name --format=csv,noheader | sort | uniq)
+
+Using all of them with JAX may
+cause the JAX failure. To avoid that you can either specify CUDA_VISIBLE_DEVICES env variable:
```
chaserileyroberts, line 19:
```suggestion
Using all of them at once with JAX may cause JAX to fail. 
To avoid that you can either specify CUDA_VISIBLE_DEVICES env variable:
```

.github/container/check-gpus.sh
```diff
@@ -0,0 +1,84 @@
...
+        fi
+    done
+    done <<< "$CUDA_VISIBLE_DEVICES"
+
+    if [[ "$issue_warning" == "true" ]]; then 
+            echo -e "${YELLOW}
+
+WARNING: There are different types of GPU specified by CUDA_VISIBLE_DEVICES. Using them with JAX may
+cause the JAX failure. To avoid that please restrict only to identical GPUs.
+List of devices in use:
```
chaserileyroberts, line 79:
```suggestion
WARNING: There are different types of GPUs specified by CUDA_VISIBLE_DEVICES.
Using them all at once with JAX may cause JAX to fail. To avoid this, please restrict to only using identical GPUs.

List of devices in use:
```

sitrep_0.zip
chaserileyroberts, line 1: Please remove this.

chaserileyroberts (changes requested):
Some minor wording edits.

Please remove the zip.

chaserileyroberts:
Has someone at Nvidia checked that this works as intended?

sitrep_0.zip
DwarKapex: Nope, removed it.

DwarKapex:
> Has someone at Nvidia checked that this works as intended?

I tested it on our DGXstation

.github/container/check-cc.sh
DwarKapex: Ack

.github/container/check-gpus.sh
```diff
@@ -0,0 +1,83 @@
...
+
+
+# Check that all devices are identical
+if [[ -z "$CUDA_VISIBLE_DEVICES" ]]; then
+    # CUDA_VISIBLE_DEVICES is not set. Check what GPUs are available
+    diff_gpu_count=$(nvidia-smi --query-gpu=gpu_name --format=csv,noheader | sort | uniq | wc -l)
+    if [[ $diff_gpu_count -ge 2 ]]; then
+            echo -e "${YELLOW}
+
+WARNING: There are $diff_gpu_count different GPUs in the system:
```
yhtang, line 14: 'different GPUs' -> 'different GPU models present'?

.github/container/check-gpus.sh
```diff
@@ -0,0 +1,83 @@
...
+
+$(nvidia-smi --query-gpu=gpu_name --format=csv,noheader | sort | uniq)
+
+Using all of them at once with JAX may cause the JAX to fail. To avoid that you can either specify CUDA_VISIBLE_DEVICES env variable:
```
yhtang, line 18: 'the JAX' -> 'JAX'

.github/container/check-gpus.sh
```diff
@@ -0,0 +1,83 @@
...
+# Check that all devices are identical
+if [[ -z "$CUDA_VISIBLE_DEVICES" ]]; then
+    # CUDA_VISIBLE_DEVICES is not set. Check what GPUs are available
+    diff_gpu_count=$(nvidia-smi --query-gpu=gpu_name --format=csv,noheader | sort | uniq | wc -l)
+    if [[ $diff_gpu_count -ge 2 ]]; then
+            echo -e "${YELLOW}
+
+WARNING: There are $diff_gpu_count different GPUs in the system:
+
+$(nvidia-smi --query-gpu=gpu_name --format=csv,noheader | sort | uniq)
```
yhtang, line 16: This info is already available when computing number of GPU models. We should reuse it instead of repeating.

.github/container/check-gpus.sh
yhtang: 'specify CUDA_VISIBLE_DEVICES env variable' -> 'specify the CUDA_VISIBLE_DEVICES environment variable'

yhtang: 'variable:' -> 'variable via:'

.github/container/check-gpus.sh
```diff
@@ -0,0 +1,83 @@
...
+
+Using all of them at once with JAX may cause the JAX to fail. To avoid that you can either specify CUDA_VISIBLE_DEVICES env variable:
+   1. inside container:
```
yhtang, line 19: inside a container:

.github/container/check-gpus.sh
```diff
@@ -0,0 +1,83 @@
...
+
+        $ export CUDA_VISIBLE_DEVICES=<coma-separated GPU ids>
+
+     or
+
+        $ CUDA_VISIBLE_DEVICES=<coma-separated GPU ids> <your-script-to-run>
+
+    2. at docker run:
```
yhtang, line 27: when starting a container:

yhtang (changes requested):

.github/container/check-gpus.sh
```diff
@@ -0,0 +1,83 @@
...
+
+Example:
+
+        docker run ... -e CUDA_VISIBLE_DEVICES=0,1,4,6 ...
+
+${NOCOLOR}"
+    fi
+else
+    # CUDA_VISIBLE_DEVICES env variable is set. Check only GPUs that are mentioned in it. 
+    available_gpus=$(nvidia-smi --query-gpu=index,gpu_name --format=csv,noheader | sort)
```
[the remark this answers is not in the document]
nouiz: There is an env variable to change the order nvidia-smi list the GPUs. But I suppose this isn't needed here.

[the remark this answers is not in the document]
yhtang: What is that variable (just out of curiosity)?

.github/container/check-gpus.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+
+YELLOW='\033[0;33m'
+NOCOLOR='\033[0m'
+
+
+# Check that all devices are identical
+if [[ -z "$CUDA_VISIBLE_DEVICES" ]]; then
+    # CUDA_VISIBLE_DEVICES is not set. Check what GPUs are available
+    diff_gpu_count=$(nvidia-smi --query-gpu=gpu_name --format=csv,noheader | sort | uniq | wc -l)
```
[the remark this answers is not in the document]
DwarKapex: No need after refactoring

.github/container/check-gpus.sh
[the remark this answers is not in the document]
DwarKapex: @yhtang Yes, I noticed that. That's why I print out all GPUs in use for customer visibility
@nouiz  I'd like to learn this variable too =)

PR #770, opened 2024-04-29
Reintroduce build args for source URLs in Dockerfiles

yhtang:
This is a replacement of #532.

Summary of changes:
- `ci.yaml` now takes an workflow_dispatch input `SOURCE_OVERRIDES` for specifying alternative origin for fetching the source code of the software that we build. The arg takes the form of a comma-separated PACKAGE=URL#REF list to override sources used by build, where PACKAGE∊{JAX,XLA,Flax,transformer-engine,T5X,paxml,praxis,maxtext,levanter,haliax,grok,mujuco,mujuco-mpc} (case-insensitive).
  for example:
  ```
  XLA=https://github.com/yhtang/xla.git#main,transformer-engine=https://github.com/yhtang/TransformerEngine.git#main
  ```
- All dockerfiles now takes in build args of form `URLREF_<SOFTWARE>=<URL>#<GIT-REF>`. These build args have default values that points to the upstream main/master branches.
- Removed `get-source.sh` and replaced it with `git-clone.sh`, which does not implicitly depend on `manifest.yaml`. Instead, it's git clone behavior is completely specified by its arguments. In addition, it will log/append the software checked out to `/opt/manifest.d/git-clone.yaml`. This is in preparation for the next stage of work where we will stop pinning versions and go back to running the presubmit on the head of everything.

Note:
- Triton build is disabled due to the intricacy of specifying default URLREF values per discussion with @olupton; this is re-enabled in https://github.com/NVIDIA/JAX-Toolbox/pull/772

Closes #532


.github/workflows/ci.yaml
```diff
@@ -139,13 +148,32 @@ jobs:
...
             .github/container/patches
 
+      - name: Create URL ref build args
+        id: source-urlrefs
+        shell: bash -x -e {0}
+        run: |
+          # converts manifest yaml to a json object of {SOFTWARE_NAME: URL#REF, ...}
+          urlrefs=$(
+            cat .github/container/manifest.yaml |\
+            yq -o=json 'to_entries | .[] | select(.value.mode == "git-clone") | {( .key | upcase | sub("-", "_") ): .value.url + "#" + .value.latest_verified_commit}' |\
```
olupton, line 158: IIUC it would just take renaming a few keys (that contain `-`) in `manifest.yaml` and we could consistently use `_` delimiters everywhere?

olupton (commented):
Generally LGTM, a few minor things to be fixed before merging.

.github/container/git-clone.sh
```diff
@@ -0,0 +1,98 @@
...
+git checkout ${GIT_REF}
+COMMIT_SHA=$(git rev-parse HEAD)
+git submodule update --init --recursive
+popd
+
+## update the manifest file
+
+mkdir -p $(dirname ${MANIFEST})
+touch ${MANIFEST}
+yq eval --inplace ". += {\"${DESTINATION}\": {\"url\": \"${GIT_REPO}\", \"ref\": \"${GIT_REF}\", \"commit\": \"${COMMIT_SHA}\"}}" ${MANIFEST}
```
olupton, line 98: This also doesn't match the `manifest.yaml` format, `ref` vs. `tracking_ref` and `commit` vs. `latest_verified_commit`.

olupton:
https://github.com/NVIDIA/JAX-Toolbox/blob/330c61b4c0246ee41a165d22e767a059b7226291/.github/container/Dockerfile.t5x.arm64 also still uses `get-source.sh`.

.github/workflows/ci.yaml
olupton: https://github.com/NVIDIA/JAX-Toolbox/blob/330c61b4c0246ee41a165d22e767a059b7226291/.github/container/pip-finalize.sh#L17 is the reason this cannot be done, I think

.github/container/git-clone.sh
olupton: I made the output format match the input format, even though "latest verified" doesn't quite make sense. Also changed the key from e.g. `/opt/transformer-engine` to `transformer-engine`, again to try and have only one YAML format.

olupton:
> https://github.com/NVIDIA/JAX-Toolbox/blob/330c61b4c0246ee41a165d22e767a059b7226291/.github/container/Dockerfile.t5x.arm64 also still uses `get-source.sh`.

Changed in 0365a79.

olupton (dismissed):
@terrykong please take a look. You might also like to review the diff between @yhtang's last commit and the current HEAD of the branch.

.github/workflows/_ci.yaml
```diff
@@ -8,17 +8,22 @@ on:
...
         default: NOT SPECIFIED
       MANIFEST_ARTIFACT_NAME:
         type: string
-        description: Artifact name in current run w/ manifest/patches. Leaving empty uses manifest/patches in current branch
+        description: 'Artifact name in current run w/ manifest/patches. Leaving empty uses manifest/patches in current branch'
         default: ''
         required: false
+      SOURCE_URLREFS:
+        type: string
+        description: 'A JSON object containing git url+refs for softwares to be built'
```
terrykong, line 21: nit: Should the description include that these are case-sensitive?

terrykong (dismissed):
Thanks for reintroducing this feature

It looks like the presubmit shows that all rosetta builds would fail b/c they depended on the [MANIFEST_FILE](https://github.com/NVIDIA/JAX-Toolbox/blob/main/rosetta/Dockerfile.t5x#L29) env var being set. Can that be included in this change-set and make sure at least rosetta-pax succeeds?

Also, upstream-t5x fails to build b/c [optax](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/8879805167/job/24379146824#step:12:213); not due to this PR, but because optax top-of-tree requires `>=0.4.26` and `0.4.26.dev*` does not satisfy. This error would also be present in JAX-Toolbox main so I don't see a reason it'd block this PR

yhtang:
> https://github.com/NVIDIA/JAX-Toolbox/blob/330c61b4c0246ee41a165d22e767a059b7226291/.github/container/Dockerfile.t5x.arm64 also still uses `get-source.sh`.

I left this out intentionally so as to not mess with Georg's ongoing PR.

terrykong:
> Yeah we need to fix that. Can we just hard code the path since we will need to rework that anyway in the follow-up PR?

@ko3n1g put together this PR to fix it: https://github.com/NVIDIA/JAX-Toolbox/pull/773/files

If it looks okay to you, please go ahead and merge

.github/container/Dockerfile.base
```diff
@@ -203,12 +199,5 @@ RUN ln -s /opt/jax_nsys/nsys-jax-ensure-protobuf /usr/local/bin/
...
-ENV MANIFEST_FILE=${DEST_MANIFEST_DIR}/${SRC_MANIFEST_FILE}
-# Copy all required files for manifestation
-ADD ${SRC_MANIFEST_FILE} \
-    bump.sh \
-    bump-openxla-triton.sh \
-    create-distribution.sh \
-    ${DEST_MANIFEST_DIR}/
-COPY patches/ ${DEST_MANIFEST_DIR}/patches/
+ADD manifest.yaml /opt/manifest.d/
+COPY patches/ /opt/manifest.d/patches/
```
DwarKapex, line 203: Also we need sort of flag whether to apply patch or not to a custom source. Let's say, you wanna build `flax` from custom repo, but you don't want to apply our patches.

.github/container/Dockerfile.grok
```diff
@@ -1,24 +1,22 @@
 # syntax=docker/dockerfile:1-labs
 
 ARG BASE_IMAGE=ghcr.io/nvidia/jax-mealkit:jax
-ARG SRC_PATH_GROK=/opt/grok-1
+ARG URLREF_GROK_1=https://github.com/xai-org/grok-1.git#main
+ARG SRC_PATH_GROK_1=/opt/grok-1
```
DwarKapex, line 5: Why `_1`?

.github/container/Dockerfile.jax
```diff
@@ -15,17 +19,21 @@ ARG BUILD_DATE
 ###############################################################################
 
 FROM ${BASE_IMAGE} as builder
+ARG URLREF_JAX
+ARG URLREF_XLA
```
DwarKapex, line 23: Maybe, add `SRC_PATH_TRANSFORMER_ENGINE`?

.github/container/git-clone.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+
+## Parse command-line arguments
+
+usage() {
```
DwarKapex, line 5: Update the `usage()`

.github/container/git-clone.sh
```diff
@@ -0,0 +1,87 @@
...
+args=$(getopt -o hm: --long help,manifest: -- "$@")
+if [[ $? -ne 0 ]]; then
+    exit 1
+fi
+
+## Set default arguments
+
+MANIFEST="/opt/manifest.d/git-clone.yaml"
+
+eval set -- "$args"
```
DwarKapex, line 32: Update or remove `while-loop`

.github/container/Dockerfile.base
yhtang: The patches are only applied when building the Rosetta container. Maybe @terrykong can answer if this is an immediate need or not.

.github/container/Dockerfile.grok
yhtang: This is the Grok-1 model and there may be future models named Grok-2, Grok-3, etc... 🤣

.github/container/git-clone.sh
yhtang: What do we need to update here?

.github/container/git-clone.sh
yhtang: Why shall we remove the while loop?

.github/container/git-clone.sh
```diff
@@ -0,0 +1,98 @@
...
+
+echo "Fetching $GIT_REPO#$GIT_REF to $DESTINATION"
+
+set -ex -o pipefail
+
+git clone ${GIT_REPO} ${DESTINATION}
+pushd ${DESTINATION}
+git checkout ${GIT_REF}
+COMMIT_SHA=$(git rev-parse HEAD)
+git submodule update --init --recursive
```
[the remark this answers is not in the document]
yhtang: Yeah we can make this change when the situation arises.

.github/container/pip-finalize.sh
```diff
@@ -14,7 +14,7 @@ IFS=$'\n'
 for line in $(cat requirements.pre | egrep '^[^#].+ @ git\+' || true); do
   # VCS installs are of the form "PACKAGE @ git+..."
   PACKAGE=$(echo "$line" | awk '{print $1}')
-  ref=$(yq e ".${PACKAGE}.latest_verified_commit" ${MANIFEST_FILE})
+  ref=$(yq e ".${PACKAGE}.latest_verified_commit" /opt/manifest.d/manifest.yaml)
```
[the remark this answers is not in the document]
yhtang: For pip-installed VCS packages, this is still the mechanism in play.

.github/workflows/_ci.yaml
yhtang: Unlike `ci.yaml`, `_ci.yaml` is an internal workflow so I'm not too worried. Let me know if you think otherwise :grin:

olupton (approved):

.github/container/Dockerfile.base
DwarKapex: @terrykong If we apply patches only for Rosetta, maybe, let's move it to rosetta docker files? WDYT?

DwarKapex (approved):
LGTM after Terry's comments on `patch` folder

.github/container/Dockerfile.base
olupton: Moving rosetta-specific stuff out of the base container sounds reasonable to me, but I don't think it's directly linked to this PR so I'm marking this resolved.

PR #772, opened 2024-04-29
Fix Triton container builds

olupton:
#770 disabled some `triton` container CI, this updates the `triton` containers to build and be tested again.
Instead of manually checking out and patching XLA's fork of Triton, take the patched repository from Bazel.


.github/container/Dockerfile.triton
```diff
@@ -1,9 +1,10 @@
 # syntax=docker/dockerfile:1-labs
 ARG BASE_IMAGE=ghcr.io/nvidia/jax-mealkit:jax
-ARG URLREF_OXLA_TRITON=https://github.com/openxla/triton.git#llvm-head
 ARG URLREF_JAX_TRITON=https://github.com/jax-ml/jax-triton.git#main
-ARG SRC_PATH_OXLA_TRITON=/opt/openxla-triton
+ARG SRC_PATH_JAX=/opt/jax
 ARG SRC_PATH_JAX_TRITON=/opt/jax-triton
```
ko3n1g, line 5: Is `openxla-triton` deprecated or why is it removed completely?

.github/container/Dockerfile.triton
```diff
@@ -17,48 +18,62 @@ ENV TRITON_NVDISASM_PATH=/usr/local/cuda/bin/nvdisasm
...
-sed -i '1s|^|include_directories(${CMAKE_SOURCE_DIR}/third_party/nvidia/backend/include)\n|' "${SRC_PATH_OXLA_TRITON}/lib/Conversion/TritonGPUToLLVM/CMakeLists.txt"
-pip wheel --wheel-dir="${SRC_PATH_OXLA_TRITON}/dist" "${SRC_PATH_OXLA_TRITON}/python"
+mkdir -p "${SRC_PATH_TRITON}/dist"
+sed -i -e 's|^add_subdirectory(unittest)|# unit tests disabled|' "${SRC_PATH_TRITON}/CMakeLists.txt"
+sed -i -e 's|BackendInstaller.copy(\["nvidia", "amd"\])|BackendInstaller.copy(["nvidia"])|g' "${SRC_PATH_TRITON}/python/setup.py"
+# Extra patches to Triton maintained in XLA. These are already applied in the checkout stage.
+XLA_TRITON_PATCHES="${SRC_PATH_XLA}/third_party/triton"
+# This patch only works with a newer LLVM than is pinned in the branch.
+PATCH_FILE="${XLA_TRITON_PATCHES}/llvm_integration/cl623185214.patch"
+if [[ $(cat "${SRC_PATH_TRITON}/cmake/llvm-hash.txt") == "6f44bb7717897191be25aa01161831c67cdf5b84" && -f "${PATCH_FILE}" ]]; then
```
ko3n1g, line 49: Can we store `6f44bb7717897191be25aa01161831c67cdf5b84` into a more declarative variable?

.github/container/Dockerfile.triton
```diff
@@ -17,48 +18,62 @@ ENV TRITON_NVDISASM_PATH=/usr/local/cuda/bin/nvdisasm
...
+  pushd "${SRC_PATH_TRITON}"
+  patch -p1 -R < "${PATCH_FILE}"
+  popd
+fi
+# This patch adds two files that are not known to CMake
+if [[ -f "${XLA_TRITON_PATCHES}/xla_extensions/sparse_dot_base.patch" ]]; then
+  sed -i -e '/ConvertLayoutOpToLLVM.cpp/a ConvertLayoutOpToLLVM/SharedToSparseDotOperand.cpp' "${SRC_PATH_TRITON}/third_party/nvidia/lib/TritonNVIDIAGPUToLLVM/CMakeLists.txt"
+  sed -i -e '/DotOpToLLVM.cpp/a DotOpToLLVM/Sparse.cpp' "${SRC_PATH_TRITON}/third_party/nvidia/lib/TritonNVIDIAGPUToLLVM/CMakeLists.txt"
+fi
```
ko3n1g, line 58: Maybe it is time for a rosetta-triton build? Adding patches to an upstream diverges a bit from our best practice. Unless I'm missing some context here which is definitely not unlikely (this is not to block this PR but as a note for ourselves, we can tackle this later too)

ko3n1g (commented):
Please take my review with a grain of salt, specifics of installing jax-trition are not super clear to me yet so I can just assess it from high-level. Generally looks good to me, with minor small remarks

.github/container/Dockerfile.triton
olupton: I just thought it would be clearer to label what used to be called `openxla-triton` as `triton`.
https://github.com/openxla/triton is a fork of https://github.com/openai/triton that supports using LLVM versions that are closer to HEAD.
JAX/XLA already contains a build using a tag from that fork, but it doesn't expose the Python bindings.
In these containers we build ~the same tag from the same fork, but as a CMake-driven build, including Python bindings, that spits out a wheel.

.github/container/Dockerfile.triton
olupton: This Dockerfile is trying to replicate the patched tag of openxla/triton described by this file: https://github.com/openxla/xla/blob/2aa2b0fec91c77805928ff368d1711e74697d171/third_party/triton/workspace.bzl#L11-L19.
The inconsistency that `sed` is trying to bridge here is that that version is intended to be built with Bazel, but in this Dockerfile we are trying to build a Python wheel using CMake: when patches add entirely new files, CMake is unaware of them and needs to be told.

gspschmid (approved):

PR #773, opened 2024-04-29
fix: Re-add `${MANIFEST_FILE}`

ko3n1g:


yhtang (changes requested):
LGTM. However, it looks like the changeset is bloated when indentation change while the actual changes are hidden and less obvious. Could you please revert the lines where the only change is indentation?

ko3n1g:
> LGTM. However, it looks like the changeset is bloated when indentation change while the actual changes are hidden and less obvious. Could you please revert the lines where the only change is indentation?

I cleaned up the commit history by moving linting issues into a separate commit. It's now easier to review if you browse changes-by-commit than changes-by-file AND it would also be easy to revert these changes by dropping a single commit. Let me know how you feel about it

olupton (approved):

olupton:
> LGTM. However, it looks like the changeset is bloated when indentation change while the actual changes are hidden and less obvious. Could you please revert the lines where the only change is indentation?

FWIW GitHub's hide whitespace option worked fine for me here.

yhtang:
> > LGTM. However, it looks like the changeset is bloated when indentation change while the actual changes are hidden and less obvious. Could you please revert the lines where the only change is indentation?
> 
> FWIW GitHub's hide whitespace option worked fine for me here.

My point is that we shouldn't let the commit history (e.g. when run git blame) be bloated by different people all applying their personal linter, especially if the linting change is not directly related to the PR.

ko3n1g:
@yhtang: It's now disabled. I was using [shell-fmt](https://marketplace.visualstudio.com/items?itemName=foxundermoon.shell-format) in VSCode which was standard in my past projects. Sorry for polluting the recent PRs

PR #774, opened 2024-04-29
Add simple script to test AWS EFA installation

DwarKapex:
A simple script to check if AWS EFA installation (thru `install-efa.sh`) succeeded.

### DIFF
--- a/.github/container/test-aws-efa.sh
+++ b/.github/container/test-aws-efa.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+# 1. Check if AWS EFA installation script completed successfully
+check=$(/opt/amazon/efa/bin/fi_info --version | grep "libfabric")
+if [[ -z "$check" ]]; then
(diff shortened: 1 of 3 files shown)


.github/container/test-aws-efa.sh
terrykong, line 1: What do you think about moving this logic into `install-efa.sh`? It looks like there's already a [smoke test](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/container/install-efa.sh#L33) in that script to check if efa is installed, so there's at least precedent.

DwarKapex: @terrykong QA wants to have a dedicated test to automate the testflow. But good point, I will remove the smote test from installation script.
Also this the first part of the test list, that checks the correctness of EFA installation process. I plan to add more tests to check that JAX does use the EFA and does not fell back to some default stuff.

terrykong (approved):

Issue #255, opened 2023-09-20
Triggering a subset of the pre-submit CI (ci.yaml)


terrykong:
The pre-submit CI [ci.yaml](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/ci.yaml) is extremely helpful for testing a change E2E, but in many cases we don't need to run all of the CI to test every change. Addressing this will speed up our development velocity since we don't have to wait for a very long CI pipeline to pass in order merge in a change.

Here is a summary of some of our options:

# Option 1: minimal pre-submit; reverts on failures
@yhtang initially proposed the idea of slimming down our pre-submit CI so that only minimal CI is run for PRs, and then the full ci.yaml can be run once the commit is in `main`. To handle the case where commits in `main` can cause failures, there can be added logic that will reverts the PR if ci.yaml fails

# Option 2: selectively running tests
We can add another "command" similar to /assistant where members of JAX-Toolbox can enumerate the tests needed. So assuming there's four stages:
```
base -> jax -> paxml
                     |
                     -> t5x
```
We can comment on a PR something like:
> /ci jax

which would jax build and test steps and any descendant jobs (t5x/pax). 

> /ci t5x paxml

which would run only t5x and paxml

> /ci
> /ci base

would run everything.

ko3n1g:
I would be curious whether using GH workflows `paths` filter as well as a docker caching strategy could help you in this regard? 

Since you have quite clearly separated dependencies, I suppose you could quite easily (after re-arranging some files into folders) condition the execution of a workflow to changes of a specified set of files.

At that point docker caching might not even be relevant anymore but using the inline cache together with `cache-from` could in some situations maybe still speed up your ci pipeline. To still fetch security updates you could build the image nightly/weekly w/o using the cache. 

Having said that, those custom commands you have implemented are admittedly quite cool (plus of course provide the benefit of having explicit control over the execution of your test suite).

ko3n1g:
I would like to share my perspective on this issue:

Option 1, so a triage-mode, sounds initially like a quicker solution but I'm afraid of some edge cases where it could backfire. The easiest I can think of is that between two nightly runs multiple PRs got merged. The auto-triage would revert all of these PRs even though potentially only a single was malicious. While we can definitely more quickly merge PRs into `main`, I am not sure if the 24hrs delay induced by nightly workflows will allows us to increase our velocity. 

[...]

PR #775, opened 2024-04-30
ci: Selective triggering

ko3n1g:
tl;dr: #255

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,7 +30,7 @@ on:
           PACKAGE∊{JAX,XLA,Flax,transformer-engine,T5X,paxml,praxis,maxtext,levanter,haliax,grok,mujuco,mujuco-mpc,gemma,big-vision,common-loop-utils,flaxformer,panopticapi} (case-insensitive)
         default: ''
       TEST_SUBSET:
-        type: string
+        type: choice
         options:
           - base
           - jax


.github/workflows/_ci.yaml
```diff
@@ -236,9 +286,10 @@ jobs:
       - name: Run integration test ${{ matrix.TEST_SCRIPT }}
         run: bash rosetta/tests/${{ matrix.TEST_SCRIPT }}
 
+  # If $TEST_SUBSET IN ("base", "jax")
   test-jax:
     needs: build-jax
-    if: inputs.ARCHITECTURE == 'amd64' # arm64 runners n/a
+    if: contains(fromJSON('["base", "jax"]'), inputs.TEST_SUBSET) && inputs.ARCHITECTURE == 'amd64' # arm64 runners n/a
```
yhtang, line 292: Is there a way to resolve this dependency from the job graph itself, since this is essentially the same information encoded in `needs`?

yhtang (changes requested):

.github/workflows/_ci.yaml
terrykong: I think it's fine to allow test-upstream-t5x to run if we're testing rosetta-t5x. Does that allow you to simplify the logic? 

It's still a big improvement if we're able to selectively trigger all t5x related nodes

ko3n1g: Unfortunately not, it was just one example but we have this issue within multiple slices of our deps graph. Whenever a node splits into several leaves, we have to limit which leaves should fire. `needs` just tells us that potentially all leaves could fire, but we need to narrow that down by our own selection.

I.e. now we build `jax` to be able to selectively test `t5x`, but don't want to run `test-jax`. The only way to really simplify it would be to get rid of `if: contains` but we can't do that since we have a tree. We can prune the set of triggers of course by grouping upstream/rosetta together, but so far I'm not creative enough to see a solution that doesn't require `contains`.

DwarKapex: [ Better readability] What do you think to create another job (or add this functionality to `pre-flight`) and move all logic there?
in `preflight`:
```
function contains_in() {
   if [[...]]; then return 'true';
   else [[...]]; then return 'false'
}
<...>
echo "RUN_JAX_TEST=$(contains_in base jax ${{ inputs.TEST_SUBSET}} )" >> $GITHUB_OUTPUT
```

in `test-jax`:
```
test-jax:
    needs: build-jax
    if: pre-flight.outputs.RUN_JAX_TEST == 'true'
<...>
```

ko3n1g: Great suggestion @DwarKapex ! I added it to pre-flight, I like it a little more if the caller decides whether to call a function rather than the callee rejecting its call.

DwarKapex: Another approach I was think of was to have a dummy jobs with name like `build-jax-4-test`. Then make `test-jax` depends on both `build-jax` and `build-jax-4-test`. The purpose of `build-jax-4-test` is to fail it if you don't want to run `test-jax`. Using dummy job helps to get rid of `if: contains(..)` in test, but seems not an elegant solution

ko3n1g:
hello, world!

ko3n1g:
test

ko3n1g:
Hi everyone! (@terrykong and @DwarKapex and @yhtang),
I've updated the watchdog to respond to subsets. You can see some examples (and also test it yourself) [in my own fork](https://github.com/ko3n1g/JAX-Toolbox/pull/28).

.github/workflows/_ci.yaml
```diff
@@ -21,20 +21,107 @@ on:
...
+      TEST_UPSTREAM_PAX: ${{ steps.run-conditions.outputs.TEST_UPSTREAM_PAX }}
+      TEST_UPSTREAM_T5X: ${{ steps.run-conditions.outputs.TEST_UPSTREAM_T5X }}
+      TEST_ROSETTA_PAX: ${{ steps.run-conditions.outputs.TEST_ROSETTA_PAX }}
+      TEST_ROSETTA_T5X: ${{ steps.run-conditions.outputs.TEST_ROSETTA_T5X }}
+    steps:
+      - name: Check out the repository under ${GITHUB_WORKSPACE}
+        uses: actions/checkout@v4
+
+      - name: Validate input `TEST_SUBSET`
+        shell: bash
```
DwarKapex, line 66: For debugging purposes:
```
shell: bash -x -e {0}
```

.github/workflows/_ci.yaml
```diff
@@ -21,20 +21,107 @@ on:
...
+          if_testset_in() {
+            local list_of_items="$1"  # The list of items as a string
+            IFS=' ' read -a test_subsets <<< "${{ steps.validate.outputs.test_subset }}"
 
+            contains=0
+            for test_subset in "${test_subsets[@]}"; do     
+              if [[ $test_subset == "" || "$list_of_items" != *"$test_subset"* ]]; then
+                contains=$(( $contains | 0 ))
+              else
+                contains=$(( $contains | 1 ))
```
DwarKapex, line 97: You actually don't need to continue when you find it. Can you do something like this?
```
echo "true"
return
```
?

.github/workflows/_ci.yaml
```diff
@@ -21,20 +21,107 @@ on:
...
+          echo BUILD_TRITON=$(if_testset_in "base jax triton") >> $GITHUB_OUTPUT
+          echo BUILD_EQUINOX=$(if_testset_in "base jax equinox") >> $GITHUB_OUTPUT
+          echo BUILD_MAXTEXT=$(if_testset_in "base jax maxtext") >> $GITHUB_OUTPUT
+          echo BUILD_LEVANTER=$(if_testset_in "base jax levanter") >> $GITHUB_OUTPUT
+          echo BUILD_T5X=$(if_testset_in "base jax upstream-t5x rosetta-t5x") >> $GITHUB_OUTPUT
+          echo BUILD_PAX=$(if_testset_in "base jax upstream-pax rosetta-pax") >> $GITHUB_OUTPUT
+          echo BUILD_GROK=$(if_testset_in "base jax grok") >> $GITHUB_OUTPUT
+
+          echo TEST_JAX=$(if_testset_in "base jax") >> $GITHUB_OUTPUT
+          echo TEST_TRITON=$(if_testset_in "base jax triton") >> $GITHUB_OUTPUT
```
DwarKapex, line 113: Looks like you don't need to do the comparison again: if you build something, you always want to test it. IMHO, you can do this:
```
echo "TEST_TRITON=${BUILD_TRITON}" >> $GITHUB_OUTPUT
```
Same for other tests.

.github/workflows/_ci.yaml
```diff
@@ -21,20 +21,107 @@ on:
...
+          echo TEST_MAXTEXT=$(if_testset_in "base jax maxtext") >> $GITHUB_OUTPUT
+          echo TEST_LEVANTER=$(if_testset_in "base jax levanter") >> $GITHUB_OUTPUT
+          echo TEST_UPSTREAM_PAX=$(if_testset_in "base jax upstream-pax") >> $GITHUB_OUTPUT
+          echo TEST_UPSTREAM_T5X=$(if_testset_in "base jax upstream-t5x") >> $GITHUB_OUTPUT
+          echo TEST_ROSETTA_PAX=$(if_testset_in "base jax upstream-pax rosetta-pax") >> $GITHUB_OUTPUT
+          echo TEST_ROSETTA_T5X=$(if_testset_in "base jax upstream-t5x rosetta-t5x") >> $GITHUB_OUTPUT
+
+          cat $GITHUB_OUTPUT
```
DwarKapex, line 121: Will it print current state of GITHUB_OUTPUT in this step or everything from all steps above?

.github/workflows/ci.yaml
```diff
@@ -60,16 +75,17 @@ jobs:
...
-      - name: Cancel workflow run if the trigger is a draft PR
-        id: cancel-if-draft
-        if: github.event_name == 'pull_request' && github.event.pull_request.draft == true
-        run: |
-          echo "Cancelling workflow for draft PR"
-          curl -X POST -H "Authorization: token ${{ github.token }}" \
-                -H "Accept: application/vnd.github.v3+json" \
-                "https://api.github.com/repos/${{ github.repository }}/actions/runs/${{ github.run_id }}/cancel"
-          while true; do sleep 1; done  # blocks execution in case workflow cancellation takes time
+      # - name: Cancel workflow run if the trigger is a draft PR
```
DwarKapex, line 80: Remove dead lines?

.github/workflows/ci.yaml
```diff
@@ -416,7 +440,7 @@ jobs:
 
   finalize:
     needs: [metadata, amd64, arm64, publish-containers]
-    if: "!cancelled()"
+    if: '!cancelled()'
```
DwarKapex, line 443: Let's make it consistent:
```
if: ${{ !cancelled() }}
```
WDYT?

.github/workflows/_ci.yaml
ko3n1g: You are right, some of these checks we can simplify. However, if differs for rosetta-tests. Here, we need to build upstream but want to test rosetta only. But I will still simplify it where possible

.github/workflows/_ci.yaml
ko3n1g: good point!

.github/workflows/ci.yaml
ko3n1g: shouldn't be dead in the first place

.github/workflows/ci.yaml
ko3n1g: There are already a lot of consistencies, like https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/_ci.yaml#L11 vs https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/workflows/ci.yaml#L419

I think we should address it in a different PR to not bloat this one. Maybe we should think about some pre-commit checks for linting and formatting?

.github/workflows/ci.yaml
DwarKapex: You're going to have another PR for this step?

DwarKapex:
Looks good to try =)

ashors1 (dismissed):
LGTM! Thanks a lot Oliver!

terrykong (approved):
Other than my one comment. LGTM

.github/workflows/ci.yaml
```diff
@@ -34,6 +29,22 @@ on:
           A comma-separated PACKAGE=URL#REF list to override sources used by build.
           PACKAGE∊{JAX,XLA,Flax,transformer-engine,T5X,paxml,praxis,maxtext,levanter,haliax,grok,mujuco,mujuco-mpc,gemma,big-vision,common-loop-utils,flaxformer,panopticapi} (case-insensitive)
         default: ''
+      TEST_SUBSET:
+        type: string
+        options:
```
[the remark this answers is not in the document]
ko3n1g: good catch! Probably some dev/debugging artifact, but tested in my fork that switching `type: choice` doesn't break the workflow so changed it

PR #784, opened 2024-05-01
Set default vocab_path in `test-pax.sh`

ashors1:
Automatically set the default `VOCAB_PATH` in `test-pax.sh` if not provided by the user. This eliminates the need for the user to pass in a vocab path that ultimately ends up being unused

### DIFF
--- a/.github/workflows/_test_pax_rosetta.yaml
+++ b/.github/workflows/_test_pax_rosetta.yaml
@@ -98,7 +98,6 @@ jobs:
           #SBATCH --gpus-per-node=${{ steps.meta.outputs.GPUS_PER_NODE }}
           #SBATCH --time=00:30:00
           #SBATCH --output=${{ steps.meta.outputs.LOG_FILE }}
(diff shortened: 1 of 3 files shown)


.github/workflows/_test_pax_rosetta.yaml
```diff
@@ -98,7 +98,6 @@ jobs:
           #SBATCH --gpus-per-node=${{ steps.meta.outputs.GPUS_PER_NODE }}
           #SBATCH --time=00:30:00
           #SBATCH --output=${{ steps.meta.outputs.LOG_FILE }}
-          #SBATCH --export="VOCAB_PATH=gs://t5-data/vocabs/cc_all.32000.100extra/sentencepiece.model,ENROOT_PASSWORD=${{ secrets.GITHUB_TOKEN }}"
```
terrykong, line 101: I think `ENROOT_PASSWORD=${{ secrets.GITHUB_TOKEN }}` should still be set

ashors1: Good catch. I'll fix that

olupton (approved):

PR #798, opened 2024-05-03
triton: build LLVM to match XLA

olupton:
The current HEAD of XLA points to an openxla/triton tag that has [code](https://github.com/openxla/triton/blob/cl623533461/lib/Target/LLVMIR/LLVMDIScope.cpp#L94) that requires a [newer](https://github.com/llvm/llvm-project/commit/6f6336858e4588ebd113ebcc930f6384a4edca54) LLVM [than the tag claims](https://github.com/openxla/triton/blob/cl623533461/cmake/llvm-hash.txt). XLA builds with [a newer LLVM](https://github.com/openxla/xla/blob/7777ebe402cb6238c73ba00428421c548ef4aaea/third_party/llvm/workspace.bzl#L7), which is why this works there.

This PR makes the `triton` container build stage compile LLVM at the same commit as XLA and use that, ignoring `llvm-hash.txt`.

Also, by making the layer of the builder stage much smaller, this will hopefully stop the `final` stage from repeating the build work when run in CI. Interestingly just making the stage smaller didn't help, but using a `large` runner did.

Fixing the build exposes that the more recent Triton (https://github.com/openxla/triton/commit/c9800840489fbe04619ad2d4e0e46b868c589d48) requires `import torch` in a code path used by JAX-Triton tests.

### DIFF
--- a/.github/container/Dockerfile.triton
+++ b/.github/container/Dockerfile.triton
@@ -18,53 +18,64 @@ ENV TRITON_NVDISASM_PATH=/usr/local/cuda/bin/nvdisasm
 RUN [ -x "${TRITON_PTXAS_PATH}" ] && [ -x "${TRITON_CUOBJDUMP_PATH}" ] && [ -x "${TRITON_NVDISASM_PATH}" ]
 
 ###############################################################################
-## Check out Triton source
+## Check out LLVM and Triton sources that match XLA, build them.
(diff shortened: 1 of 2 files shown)


gspschmid (dismissed):
lgtm

gspschmid (approved):

PR #802, opened 2024-05-06
test-maxtext.sh: do not compare strings as integers

olupton:
Also `set -eo pipefail` earlier, which should have caught this.

Error message in 2024-05-05 is
```
/usr/local/bin/test-maxtext.sh: line 148: [: dot_product: integer expression expected
```


kocchop (approved):
Thanks, LGTM!

PR #804, opened 2024-05-06
Add missing python packages for levanter tests

DwarKapex:


ko3n1g:
Why are out levanter test displayed as successful when in fact they are not?

ko3n1g (approved):

PR #807, opened 2024-05-07
ci: Add `watchdog`

ko3n1g:
We'd need to merge this before working on #775 as workflows only run if they are present in `main`. Therefore, this PR does not reflect the additional inputs required for selective triggering, I would add these in #775

### DIFF
--- a/.github/workflows/_watchdog.yaml
+++ b/.github/workflows/_watchdog.yaml
@@ -1,0 +1,68 @@
+name: watchdog
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  woof:
+    runs-on: ubuntu-22.04
+    if: >
+      github.event_name == 'issue_comment' &&
+      github.event.issue.pull_request &&
+      startsWith(github.event.comment.body, '/ci') &&
+      contains('ko3n1g,nouiz,joker-eph,yhtang,terrykong,hemildesai,chaserileyroberts,CliveUnger,DwarKapex,mingxu1067,kocchop,SahilJain314,ashors1,maanug-nv,hmonishN,gspschmid', format('{0},', github.actor))
+    steps:
+      - name: Get PR number
+        shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/_watchdog.yaml
```diff
@@ -0,0 +1,43 @@
...
+              pull_number: ${{ steps.get-pr-num.outputs.pr_number }}
+            });
+            console.log('Pull Request Information:', pr.data);
+            return pr.data.head.ref;
+
+      - name: Trigger `CI` workflow on subset of tests          
+        uses: aurelien-baudet/workflow-dispatch@v2.1.1
+        with:
+          workflow: "CI"
+          token: ${{ secrets.TOKEN }}
```
terrykong, line 41: We spoke about this, but just leaving the upshot from offline. We should try to reuse the automatically generated one `secrets.GITHUB_TOKEN`

.github/workflows/_watchdog.yaml
```diff
@@ -0,0 +1,43 @@
...
+          wait-for-completion: false
+          ref: ${{ steps.get-pr-branch.outputs.result }}
```
terrykong, line 43: I think it would be great to print out the URL for the dispatch CI run somewhere, either:

1. `GITHUB_STEP_SUMMARY`
2. within a step name itself or
3. as a PR comment

Perhaps (1) b/c it's clickable. WDYT?

terrykong (changes requested):

.github/workflows/_watchdog.yaml
ko3n1g: Implemented (1)

ko3n1g:
Hey @terrykong, added the step summary and a check that limits the watchdog to certain actors.

terrykong (approved):

DwarKapex (approved):

PR #808, opened 2024-05-07
Skip downloading model checkpoints

DwarKapex:


ko3n1g (approved):
I'm approving this as the CI is green everywhere.

But I looked into yesternights nightly and was also there not able to find checkpoints, so I'm a bit confused about this task. Could you point me to an example workflow/job/step where this change becomes effective?

DwarKapex:
> But I looked into yesternights nightly and was also there not able to find checkpoints, so I'm a bit confused about this task. Could you point me to an example workflow/job/step where this change becomes effective?

Sure. PaxML team added a new feature to save checkpoints in case of failures (for instance, for LLaMA models) for further analysis. And they save them as a part of artifacts. The size of the checkpoint can be huge (up to 15 Gb [example](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/8957256120/job/24602019902#step:3:11)). We have only 14Gb disk space available on `ubuntu-22.04` runners (the one we use for metrics, for example), meaning that the job fails with OOM issue and so does the whole workflow, leaving us without test run stats.
Checkpoints are the part of uploaded artifacts, so it is simply impossible to download the artifact and them remove checkpoints (because of limited disk space). So one of the solutions is simply save checkpoints and test logs separately.

PR #810, opened 2024-05-08
Always remove eval checkpoints in test-pax.sh

ashors1:
Currently, eval checkpoints are only removed if the eval tests succeed. In this PR, we remove the checkpoints in the case of failure as well.

### DIFF
--- a/.github/container/test-pax.sh
+++ b/.github/container/test-pax.sh
@@ -388,6 +388,9 @@ else
 fi
 
 if [[ ${EVALUATE} -ne 0 ]]; then
+
+  trap "rm -rf ${OUTPUT}/checkpoints" ERR
+
   ## train for 0 steps to generate an initial checkpoint
   python -m paxml.main \
     --fdl_config=${CONFIG} \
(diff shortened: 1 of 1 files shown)


.github/container/test-pax.sh
```diff
@@ -388,6 +388,9 @@ else
 fi
 
 if [[ ${EVALUATE} -ne 0 ]]; then
+
+  trap "rm -rf ${OUTPUT}/checkpoints" ERR
```
terrykong, line 392: Should we cleanup if this runs successfully as well?

I was thinking maybe:

```suggestion
  trap "rm -rf ${OUTPUT}/checkpoints" ERR INT HUP TERM EXIT
```

I think if you add `EXIT`, you shouldn't need the `rm -rf` that appears below

terrykong (changes requested):
Generally LGTM. Left a single comment

.github/container/test-pax.sh
ashors1: Done, thanks!

terrykong (approved):

PR #813, opened 2024-05-09
Skip checkpoints downloading

DwarKapex:


ashors1:
@DwarKapex https://github.com/NVIDIA/JAX-Toolbox/pull/810 was merged, and that should fix this problem. The finalize step of the presubmit CI passed. So I don't think this PR is necessary anymore

PR #815, opened 2024-05-10
Remove all checkpoints download verification

DwarKapex:
After merging #810 we don't need explicit workaround with checkpoints.


terrykong (approved):
Approving assuming CI looks good

olupton (approved):

PR #818, opened 2024-05-13
Add NCCL sanity check

gspschmid:
Adds a small utility `nccl-sanity-check` that is intended to serve as a basic diagnostic for the NCCL setup. The utility connects all the ranks and performs several single-float all-reduces. When the minimum timing of these attempts is below a certain configurable thresholds (1ms, by default) the utility exits successfully.

The utility takes `world-size world-rank local-rank coordinator-address` as arguments. To simplify invocations a wrapper script `nccl-sanity-check.sh` uses `jax.distributed` to extract the configuration from slurm or openmpi environmental variables.

This was tested manually on multiple nodes via slurm using an invocation like
```
srun --container-image=ghcr.io/nvidia/jax:jax-2024-05-05 --container-mounts=... -N 4 --tasks-per-node 8 bin/bash -c "cd ... && ./nccl-sanity-check.sh"
```

### DIFF
--- a/.github/container/Dockerfile.base
+++ b/.github/container/Dockerfile.base
@@ -174,6 +174,14 @@ ADD install-efa.sh /usr/local/bin
 ENV LD_LIBRARY_PATH=/opt/amazon/efa/lib:${LD_LIBRARY_PATH}
 ENV PATH=/opt/amazon/efa/bin:${PATH}
 
+##############################################################################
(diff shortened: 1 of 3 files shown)


olupton (approved):
LGTM, better in than out 👍

nouiz:
That looks great. I think this should be documented somewhere. Right now, users won't know that it exists.

PR #822, opened 2024-05-14
nsys-jax: handle parallel compilation

olupton:
Improve handling of parallel compilation. Now, range hierarchies in worker threads are spliced in under the main thread ranges that launch them. This allows a meaningful summary of where compilation time is spent even when parallelism is enabled.

Also support profiles from runs with multiple GPUs per process.


.github/container/jax_nsys/python/jax_nsys/jax_nsys/visualization.py
```diff
@@ -29,6 +29,8 @@ def create_flamegraph(
     flat_data = ""
     for loc, value in data.items():
         assert not any(";" in x for x in loc)
+        if value < 0.0:
```
gspschmid, line 32: Nit: I would write this as `assert value >= 0.0, f'Negative value ...'`

.github/container/jax_nsys/python/jax_nsys/jax_nsys/data_loaders.py
```diff
@@ -33,44 +34,34 @@ def is_communication(row):
...
-    compute_df = thunk_df[~thunk_df["Communication"]]
-    for compute_thunk in compute_df.itertuples():
-        # This should just find the thunk itself
-        mask = (compute_df["ProjEndNs"] > compute_thunk.ProjStartNs) & (
-            compute_df["ProjStartNs"] < compute_thunk.ProjEndNs
-        )
-        assert mask.sum() == 1
+                compute_df["ProjStartNs"].iloc[1:].array
+                >= compute_df["ProjEndNs"].iloc[:-1].array
+            ).all()
```
gspschmid, line 56: Is `compute_df` known to be sorted by `ProjStartNs`? (I assume so as otherwise this is unlikely to ever have succeeded?)

.github/container/jax_nsys/python/jax_nsys/jax_nsys/utils.py
```diff
@@ -0,0 +1,23 @@
...
+    mask &= df["NumChild"] != 0
+    for row in df[mask].itertuples():
+        child_mask = df["RangeStack"].str.startswith(f"{row.RangeStack}:")
+        to_remove = child_mask if to_remove is None else child_mask | to_remove
+        df.loc[row.Index, ["NumChild", "DurChildNs"]] = 0
+        df.loc[row.Index, "DurNonChildNs"] = row.DurNs
+    if to_remove is None:
+        return df
+    else:
+        return df[~to_remove].copy()
```
gspschmid, line 23: Given that the procedure has already mutated the original `df` at this point, is there any good reason to still `.copy()`?

.github/container/jax_nsys/python/jax_nsys/jax_nsys/analysis.py
```diff
@@ -0,0 +1,91 @@
...
+        # Find all child ranges; some may still be in the main thread. Assume for now
+        # that the sequence will just be something like:
+        #   M(A) M(A) M(A) .. W1(B) W2(B) W3(B) W1(B) .. M(C) M(C)
+        # i.e. the main thread M does some task (A), then workers W{1,2,3} do some task
+        # (B) in parallel, then the main thread continues with another task (C),
+        # without overlap between A, B and C. For simplicity, we assume that there is
+        # only one parallel region B in a given parent range, but this restriction
+        # could be relaxed if needed.
+        child_df = compile_df.loc[
+            compile_df["RangeStack"].str.startswith(f"{launcher_row.RangeStack}:"), :
```
gspschmid, line 51: Since you're also using this idiom below might be worth factoring into a helper?

gspschmid (approved):
Superficial review on my end, but looks sensible (an example profile or tests would make it easier to judge the intended semantics).

.github/container/jax_nsys/python/jax_nsys/jax_nsys/data_loaders.py
olupton: I think we're allowed to expect that the data frame is sorted by launch time, which should mean it is also sorted by execution start time (under the hypothesis being asserted). I think it's conceivable that the assertion may fail in future, as I think in theory CUDA graphs could allow compute-compute overlap - but that failure is desirable.

.github/container/jax_nsys/python/jax_nsys/jax_nsys/analysis.py
olupton: Done

gspschmid (approved):

PR #823, opened 2024-05-14
Update base container to CUDA 12.4.1 with cuDNN 9.1

DwarKapex:


.github/container/Dockerfile.base
```diff
@@ -146,7 +146,6 @@ RUN install-nsight.sh
 ###############################################################################
 
 ADD install-cudnn.sh /usr/local/bin
-RUN install-cudnn.sh
```
terrykong, line 149: Is cudnn now included? Do we drop `install-cudnn.sh` moving forward?

DwarKapex: Yes, the cuDNN 9.1.0 is included in the CUDA 12.4.1 image (as you can see, the CUDA base images name `nvidia/cuda:12.4.1-cudnn-devel-ubuntu22.04` contains `cudnn`) and it's frozen, meaning we have less degree of freedom here and always work with certain CUDA and cuDNN.
I think, it is reasonable to leave `install-cudnn.sh` for couple reasons:
1. in case we will want to update the cuDNN for some experiments
2. give a customer a tool to update the cuDNN version for their needs.

terrykong (approved):

PR #827, opened 2024-05-15
fixed XLA flag for simplify fp conversion

kocchop:
removed the deprecated `--xla_gpu_simplify_all_fp_conversions`
replaced with `--xla_allow_excess_precision`

### DIFF
--- a/.github/container/test-maxtext.sh
+++ b/.github/container/test-maxtext.sh
@@ -187,7 +187,6 @@ export CUDA_DEVICE_MAX_CONNECTIONS=1
 
 export BASE_XLA_FLAGS=${BASE_XLA_FLAGS:---xla_gpu_enable_latency_hiding_scheduler=true 
                 --xla_gpu_enable_triton_gemm=false
-                --xla_allow_excess_precision
                 --xla_gpu_graph_level=0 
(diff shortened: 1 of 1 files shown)


.github/container/test-maxtext.sh
```diff
@@ -187,7 +187,7 @@ export CUDA_DEVICE_MAX_CONNECTIONS=1
 
 export BASE_XLA_FLAGS=${BASE_XLA_FLAGS:---xla_gpu_enable_latency_hiding_scheduler=true 
                 --xla_gpu_enable_triton_gemm=false
-                --xla_gpu_simplify_all_fp_conversions 
+                --xla_allow_excess_precision
```
olupton, line 190: This seems to be the default, why do we need to set it?

olupton (changes requested):

.github/container/test-maxtext.sh
olupton, line 190:
```suggestion
```

gpupuck: agree

kocchop: Thanks, will remove it

olupton (approved):

PR #830, opened 2024-05-16
Add FMHA PAXML test 

hmonishN:
Adding the JAX PAXML FMHA E2E system test to check for fmha lowering support. Following are the steps implemented in the test:

FMHA lowering flag is enabled, enabled the dumping of hlo to track fmha forward and backward instructions.

Similar to PR: https://github.com/NVIDIA/JAX-Toolbox/pull/442

### DIFF
--- a/.github/container/test-pax.sh
+++ b/.github/container/test-pax.sh
@@ -23,6 +23,8 @@ usage() {
     echo "  -s, --steps                Number of steps to run, defaults to 500."
     echo "  --multiprocess             Enable the multiprocess GPU mode."
     echo "  -o, --output NAME          Name for the output folder, a temporary folder will be created if none specified."
+    echo "  --save-hlo {0, 1}          1 to save the dumped hlo, 0 to remove the hlo dumped folder"
+    echo "  --enable-fmha {0, 1}       1 to enable fmha testing, 0 to run test without fmha; default is 0"
     echo "  --data-parallel            Data parallelism to use. Defaults to 1."
     echo "  --fsdp                     Fully-sharded data parallelism to use. Defaults to 1."
     echo "  --tensor-parallel          Tensor parallelism to use. Defaults to 1."
@@ -32,7 +34,7 @@ usage() {
     exit $1
 }
 
-args=$(getopt -o a:b:s:o:n:h --long additional-args:,batch-per-gpu:,dtype:,enable-te,enable-dropout,enable-fused-attn,model-type:,evaluate,steps:,help,multiprocess,output:,data-parallel:,fsdp:,tensor-parallel:,pipeline-parallel:,nodes: -- "$@")
+args=$(getopt -o a:b:s:o:n:h --long additional-args:,batch-per-gpu:,dtype:,enable-te,enable-dropout,enable-fused-attn,model-type:,enable-fmha:,evaluate,steps:,help,multiprocess,output:,save-hlo:,data-parallel:,fsdp:,tensor-parallel:,pipeline-parallel:,nodes: -- "$@")
 if [[ $? -ne 0 ]]; then
     exit $1
 fi
@@ -55,6 +57,8 @@ NVTE_FUSED_ATTN=0
 DROPOUT=0
 EVALUATE=0
 ADDITIONAL_ARGS=""
+ENABLE_FMHA=${ENABLE_FMHA:-0}
+SAVE_HLO=${SAVE_HLO:-1}
 
 eval set -- "$args"
 while [ : ]; do
@@ -74,6 +78,10 @@ while [ : ]; do
         --enable-te)
             ENABLE_TE=1
             shift 1
+            ;;
+        --enable-fmha)
+            ENABLE_FMHA="$2"
+            shift 2
             ;;
         --enable-dropout)
             DROPOUT='0.1'
@@ -101,6 +109,10 @@ while [ : ]; do
             ;;
         -o | --output)
             OUTPUT=$2
+            shift 2
+            ;;
+        --save-hlo)
+            SAVE_HLO="$2"
             shift 2
             ;;
         --data-parallel)
@@ -136,6 +148,21 @@ while [ : ]; do
     esac
 done
 
+# Set hlo dump folder after output folder is set.
+HLO_DIR=${OUTPUT}/hlo
+export BASE_XLA_FLAGS="${BASE_XLA_FLAGS:---xla_dump_hlo_as_text --xla_dump_to=${HLO_DIR}}"
+export XLA_FLAGS="${BASE_XLA_FLAGS} ${XLA_FLAGS:-}"
+echo "HLO will be dumped in ${HLO_DIR} dir."
+
+## Setting the env variables for FMHA
+if [[ "$ENABLE_FMHA" -eq "1" ]]; then  
+    echo "Setting XLA FMHA Flags";
+    export BASE_XLA_FLAGS_FMHA="${BASE_XLA_FLAGS_FMHA:---xla_gpu_fused_attention_use_cudnn_rng=true --xla_gpu_enable_cudnn_fmha=true}"
+    export XLA_FLAGS="${BASE_XLA_FLAGS_FMHA} ${XLA_FLAGS:-}"
+fi
+
+echo "XLA FLAGS: $XLA_FLAGS"
+
 # # Set derived variables
 
 GPUS_PER_NODE=$(nvidia-smi -L | grep -c '^GPU')
@@ -149,8 +176,10 @@ print_var NGPUS
 print_var OUTPUT
 print_var MULTIPROCESS
 print_var ENABLE_TE
+print_var ENABLE_FMHA
 print_var NVTE_FUSED_ATTN
 print_var EVALUATE
+print_var SAVE_HLO
 print_var DROPOUT
 print_var DP
 print_var FSDP
@@ -421,3 +450,26 @@ fi
 
 set +x
 echo "Output at ${OUTPUT}"
+
+if [[ "$ENABLE_FMHA" -eq "1" ]]; then 
+    ## Check if fmha instructions are present in the HLO dumped file or not.
+    fmha_regex="fmha[-bmm]?[-scale]?[-bias]?[-mask]?[-softmax]?[-dropout]?[-bmm]?[-backward]?*"
+    result=$(grep -irlnE "$fmha_regex" "${HLO_DIR}/"*.txt)
+
+    if [[ $SAVE_HLO -eq 0 ]]; then
+        rm -rf $HLO_DIR
+        echo "Removed dumped HLO directory!"
+    fi
+
+    if [ -z "$result" ]; then
+        echo "E: No FMHA instructions were found in the hlo files!"
+	exit 1
+    else
+        echo -e "Found FMHA instructions in the following HLO files: \n $result"
+    fi
+else
+    if [[ $SAVE_HLO -eq 0 ]]; then
+        rm -rf $HLO_DIR
+ 	echo "Removed dumped HLO directory!"
+    fi
+fi
--- a/.github/container/test-pax.sh
+++ b/.github/container/test-pax.sh
@@ -448,10 +448,10 @@ else
     $([[ $MULTIPROCESS != 0 ]] && echo --multiprocess_gpu)
 fi
 
-set +x
-echo "Output at ${OUTPUT}"
+echo "Checking for FMHA instructions in HLO!"
 
 if [[ "$ENABLE_FMHA" -eq "1" ]]; then 
+    echo "Inside if Statement!"
     ## Check if fmha instructions are present in the HLO dumped file or not.
     fmha_regex="fmha[-bmm]?[-scale]?[-bias]?[-mask]?[-softmax]?[-dropout]?[-bmm]?[-backward]?*"
     result=$(grep -irlnE "$fmha_regex" "${HLO_DIR}/"*.txt)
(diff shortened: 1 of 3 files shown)


.github/container/test-pax.sh
```diff
@@ -422,5 +451,31 @@ else
...
 
+echo "Checking for FMHA instructions in HLO!"
+
+if [[ "$ENABLE_FMHA" -eq "1" ]]; then 
+    echo "Inside if Statement!"
+    ## Check if fmha instructions are present in the HLO dumped file or not.
+    fmha_regex="fmha[-bmm]?[-scale]?[-bias]?[-mask]?[-softmax]?[-dropout]?[-bmm]?[-backward]?*"
+    result=$(grep -irlnE "$fmha_regex" "${HLO_DIR}/"*.txt)
+
+    if [[ $SAVE_HLO -eq 0 ]]; then
```
DwarKapex, line 462: I would expect to skip saving if `$SAVE_HLO == 0`, rather than delete it afterwards.

.github/container/test-pax.sh
```diff
@@ -422,5 +451,31 @@ else
...
+    fi
+
+    if [ -z "$result" ]; then
+        echo "E: No FMHA instructions were found in the hlo files!"
+	exit 1
+    else
+        echo -e "Found FMHA instructions in the following HLO files: \n $result"
+    fi
+else
+    if [[ $SAVE_HLO -eq 0 ]]; then
```
DwarKapex, line 474: You have this statement in both if-else-fi branches, so you can move it out of if-else statement.

.github/container/test-pax.sh
```diff
@@ -422,5 +451,31 @@ else
     $([[ $MULTIPROCESS != 0 ]] && echo --multiprocess_gpu)
 fi
 
+echo "Checking for FMHA instructions in HLO!"
+
+if [[ "$ENABLE_FMHA" -eq "1" ]]; then 
+    echo "Inside if Statement!"
```
DwarKapex, line 457: Remove?

.github/container/test-pax.sh
```diff
@@ -136,6 +148,21 @@ while [ : ]; do
     esac
 done
 
+# Set hlo dump folder after output folder is set.
+HLO_DIR=${OUTPUT}/hlo
+export BASE_XLA_FLAGS="${BASE_XLA_FLAGS:---xla_dump_hlo_as_text --xla_dump_to=${HLO_DIR}}"
```
DwarKapex, line 153: Could you please explain logic here: is `BASE_XLA_FLAGS` is set, than you always skip setting `HLO_DIR`?

.github/container/test-pax.sh
```diff
@@ -136,6 +148,21 @@ while [ : ]; do
...
+export XLA_FLAGS="${BASE_XLA_FLAGS} ${XLA_FLAGS:-}"
+echo "HLO will be dumped in ${HLO_DIR} dir."
+
+## Setting the env variables for FMHA
+if [[ "$ENABLE_FMHA" -eq "1" ]]; then  
+    echo "Setting XLA FMHA Flags";
+    export BASE_XLA_FLAGS_FMHA="${BASE_XLA_FLAGS_FMHA:---xla_gpu_fused_attention_use_cudnn_rng=true --xla_gpu_enable_cudnn_fmha=true}"
```
DwarKapex, line 160: Save here as above

.github/workflows/_test_upstream_pax.yaml
```diff
@@ -30,12 +30,22 @@ on:
...
 
-  single-process-multi-device:
+  pax-single-process-multi-device:
     strategy:
       matrix:
-        PARALLEL_CONFIG:
-        - [1, 8, 1, 1]
-        - [1, 1, 2, 4]
+        include:
+          - TEST_NAME: 8DP1FSDP1TP1PP
```
DwarKapex, line 37: Do you change the test name?

.github/container/test-pax.sh
hmonishN: for fmha the hlo is dumped in output folder by default to search for fmha instructions. The idea is to delete the hlo folder if user doesn't want to have hlo folder inside the output folder which is saved as artifact.

.github/container/test-pax.sh
```diff
@@ -32,7 +34,7 @@ usage() {
     exit $1
 }
 
-args=$(getopt -o a:b:s:o:n:h --long additional-args:,batch-per-gpu:,dtype:,enable-te,enable-dropout,enable-fused-attn,model-type:,evaluate,steps:,help,multiprocess,output:,data-parallel:,fsdp:,tensor-parallel:,pipeline-parallel:,nodes: -- "$@")
+args=$(getopt -o a:b:s:o:n:h --long additional-args:,batch-per-gpu:,dtype:,enable-te,enable-dropout,enable-fused-attn,model-type:,enable-fmha:,evaluate,steps:,help,multiprocess,output:,save-hlo:,data-parallel:,fsdp:,tensor-parallel:,pipeline-parallel:,nodes: -- "$@")
```
ashors1, line 37: We already have an `enable-fused-attn` flag which enables TE FMHA. Do you think we should merge `enable-fused-attn` and `enable-fmha` into a single flag?

.github/container/test-pax.sh
hmonishN: dumping the hlo is enabled by default in BASE_XLA_FLAGS, and BASE_XLA_FLAGS are appended to XLA_FLAGS env var.  if user wants to test fmha then BASE_XLA_FLAGS_FMHA is added and appended to XLA_FLAGS. The idea is to preserve the env var XLA_FLAGS before execution of this script.

.github/workflows/_test_upstream_pax.yaml
hmonishN: Yes, I have matrixified the args passed for this sub job as multi-node job and the JOB_NAME is defined as part of the matrix.

.github/container/test-pax.sh
hmonishN: I don't see that flag being used, is it a dummy arg? I can remove enable-fused-attn since enable-fmha is already added in t5x test ([here](https://github.com/NVIDIA/JAX-Toolbox/blob/hmonish/add_fmha_paxml_test/.github/container/test-t5x.sh#L26)) and will be helpful in keeping the naming convention simpler.

ashors1: It's used here: https://github.com/NVIDIA/JAX-Toolbox/blob/f44cdef645698a4f0d28ad34b7d100f48ee7da11/.github/container/test-pax.sh#L90-L93 Since these flags essentially do the same thing, but one is intended to be used with TE and one is used without TE, I think it would make sense to merge the two and just use this flag in conjunction with `enable-te` to determine whether to set `NVTE_FUSED_ATTN` or `--xla_gpu_enable_cudnn_fmha`

hmonishN: Oh, I see but how to handle the scenario (if it's a feasible one) where user wants to enable TE with NVTE_FUSED_ATTN but not --xla_gpu_enable_cudnn_fmha ?

nvjax-svc-0: I was thinking that if `--enable-te` and `--enable-fmha` are both set, we would set `NVTE_FUSED_ATTN=1`. If just `--enable-fmha` is set, we would instead use `--xla_gpu_enable_cudnn_fmha`. When using TE, I think `--xla_gpu_enable_cudnn_fmha` should be a no-op. Will confirm with the TE team.

ashors1: I was thinking that if `--enable-te` and `--enable-fmha` are both set, we would set `NVTE_FUSED_ATTN=1`. If just `--enable-fmha` is set, we would instead use `--xla_gpu_enable_cudnn_fmha`. When using TE, I think `--xla_gpu_enable_cudnn_fmha` should be a no-op. Will confirm with the TE team.

.github/container/test-pax.sh
hmonishN: Incorporated.

.github/container/test-pax.sh
hmonishN: Incorporated.

.github/container/test-pax.sh
ashors1: Confirmed that `--xla_gpu_enable_cudnn_fmha` is a no-op when TE is enabled

hmonishN: sounds good, lemme merge the flags then.

.github/container/test-pax.sh
```diff
@@ -75,14 +78,15 @@ while [ : ]; do
             ENABLE_TE=1
             shift 1
             ;;
+        --enable-fmha)
+            ENABLE_FMHA="$2"
+	    NVTE_FUSED_ATTN=1
```
ashors1, line 83: I think this should be `NVTE_FUSED_ATTN="$2"`

hmonishN: Updated

.github/container/test-pax.sh
DwarKapex: OK, let me clarify my question:
line 150 literally means:
```
if [[ "$BASE_XLA_FLAGS" == "" ]]; then
      BASE_XLA_FLAGS = "--xla_dump_hlo_as_text --xla_dump_to=${HLO_DIR}}"
fi
```
Meaning, that if `BASE_XLA_FLAGS` is already set (by any previous scripts, or globally in the system, etc), ${HLO_DIR} will not have any effect at all.

Is that expected behaviour?

DwarKapex: And why do you `export` it? You use it only locally.

DwarKapex:
LGTM after clarifying about extra XLA_FLAGS

.github/container/test-pax.sh
hmonishN: The mechanism was added as per the review comment of same PR for t5x: https://github.com/NVIDIA/JAX-Toolbox/pull/442#discussion_r1468912293
refer to the discussion for details of the implementation.
The implementation BASE_XLA_FLAGS="${BASE_XLA_FLAGS:---xla_dump_hlo_as_text --xla_dump_to=${HLO_DIR}}" means update the BASE_XLA_FLAGS with previous definition if any and append xla dump hlo flags to the env vars. This also gives us the flexibility of "zero out" the env var in this script without modifying code in this script by just doing BASE_XLA_FLAGS=""

hmonishN:
Can we merge this PR if it looks good?

.github/container/test-pax.sh
```diff
@@ -55,6 +57,8 @@ NVTE_FUSED_ATTN=1
 DROPOUT=0
 EVALUATE=0
 ADDITIONAL_ARGS=""
+ENABLE_FMHA=${ENABLE_FMHA:-0}
```
ashors1, line 60: Sorry, one more comment. Can we enable FMHA by default? We recently enabled FMHA by default in our nightly images, so it would be good to keep the same behavior here

hmonishN: Enabling it here by default will mean that for all the pax tests the fmha will be enabled, do we want that?

ashors1: Yes, that's the current behavior. Only this 5b test runs with FMHA disabled: https://github.com/NVIDIA/JAX-Toolbox/blob/bcb72054e9b3e707a31d59e4a68fcc2d7fa8a96f/.github/workflows/_test_pax_rosetta.yaml#L255

hmonishN: Enabled by default now, updated args for 5b model in _test_pax_rosetta.yaml

.github/workflows/_test_upstream_pax.yaml
```diff
@@ -210,6 +221,14 @@ jobs:
             BATCH_SIZE: 4
             EVALUATE: true
             ADDITIONAL_ARGS: "--model-type LLaMA70BProxy --evaluate"
+          - TEST_NAME: 2DP1FSDP1TP4PP_fmha
+            PARALLEL_CONFIG: [4, 2, 1, 1]
```
ashors1, line 225: Is this redundant now that fmha is enabled by default?

hmonishN: Yeah, I'll use the existing test configs (1-2 tests ) to add additional arg to save hlo (used to debug fmha instructions). Don't want to enable hlo dumping by default. (Terry mentioned it takes a lot of time in nightly to download artifacts. )

ashors1:
Changes look good to me. The presubmit CI did not complete, so I reran. I'll approve once CI completes and we verify that results are as expected

.github/container/test-pax.sh
```diff
@@ -17,12 +17,13 @@ usage() {
...
     echo "  --enable-te                If set, will run with env var ENABLE_TE=1." 
     echo "  --enable-dropout           If set, will set DROPOUT_PROB to 0.1."
-    echo "  --disable-fused-attn       Whether disable TE fused attention."
     echo "  --model-type               One of 126M, 5B, LLaMA70BProxy. Defaults to 126M"
     echo "  --evaluate                 Whether to test evaluation rather than training."
     echo "  -s, --steps                Number of steps to run, defaults to 500."
     echo "  --multiprocess             Enable the multiprocess GPU mode."
     echo "  -o, --output NAME          Name for the output folder, a temporary folder will be created if none specified."
+    echo "  --save-hlo {0, 1}          1 to save the dumped hlo, 0 to remove the hlo dumped folder"
+    echo "  --enable-fmha {0, 1}       1 to enable fmha testing, 0 to run test without fmha; default is 0"
```
ashors1, line 26: nit: default doesn't match below

PR #831, opened 2024-05-17
CUDA_MODULE_LOADING: use default lazy loading

olupton:
Prefer to set fewer magic variables. Eager loading cuDNN/cuBLAS during XLA compilation can also be noticably slow.

This was added in https://github.com/NVIDIA/JAX-Toolbox/pull/329


DwarKapex:
Does it affect only compilation time? How much do we save on that?
WDYT to simply revert #329?

olupton:
Opened https://github.com/NVIDIA/JAX-Toolbox/pull/842 with a straight revert.

PR #841, opened 2024-05-20
Install bazelisk as /usr/local/bin/bazel

DwarKapex:
Add `bazelisk` to manage what version of `bazel` to use on the fly

### DIFF
--- a/.github/container/Dockerfile.jax
+++ b/.github/container/Dockerfile.jax
@@ -38,7 +38,7 @@ EOF
 ADD build-jax.sh local_cuda_arch test-jax.sh /usr/local/bin/
 # Install bazelisk
 RUN ARCH="$(dpkg --print-architecture)" && \
-    wget -O /usr/bin/bazel https://github.com/bazelbuild/bazelisk/releases/latest/download/bazelisk-linux-${ARCH} && \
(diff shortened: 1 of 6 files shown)


.github/container/Dockerfile.jax
```diff
@@ -36,6 +36,10 @@ RUN --mount=type=ssh \
 EOF
 
 ADD build-jax.sh local_cuda_arch test-jax.sh /usr/local/bin/
+# Install bazelisk
+RUN ARCH="$(dpkg --print-architecture)" && \
+    wget -O /usr/bin/bazel https://github.com/bazelbuild/bazelisk/releases/latest/download/bazelisk-linux-${ARCH} && \
+    chmod +x /usr/bin/bazel
```
olupton, line 42: I think `/usr/local/bin/bazel` would be more consistent with the rest of the container.

olupton (changes requested):
The Triton container dockerfile also needs to learn how to use `bazel`.

.github/container/build-jax.sh
```diff
@@ -21,7 +21,7 @@ supported_compute_capabilities() {
 }
 
 clean() {
-    $(find -type f -executable -iname "bazel*") clean --expunge || true
+    /usr/bin/bazel clean --expunge || true
```
olupton, line 24: Maybe just `bazel` is cleaner?

olupton (approved):

PR #842, opened 2024-05-21
Revert "Adds CUDA_MODULE_LOADING=EAGER to core jax container env vars"

olupton:
Prefer to set fewer magic variables.
Note that these values were anyway not used inside e.g. the JAX unit test environment, so this was a source of inconsistency.
Eager loading cuDNN/cuBLAS during XLA compilation can also be noticeably slow.
See documentation here: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#cuda-environment-variables
Replaces https://github.com/NVIDIA/JAX-Toolbox/pull/831. Reverts NVIDIA/JAX-Toolbox#329.


ashors1 (approved):
I reran the CI because the previous run timed out. But assuming the CI passes, LGTM

PR #845, opened 2024-05-22
Work around upstream hermetic Python changes

olupton:
- Always use editable installations of jax and jaxlib, which avoids having to deal with wheel filenames that have dates in them. Using an editable installation requires that we disable hashes.
- Do not tackle the problem that this leaves the JAX source repo in a dirty state. People will have to be careful not to merge the modified requirements files.


DwarKapex (dismissed):
LGTM

.github/container/build-jax.sh
```diff
@@ -275,14 +266,21 @@ time python build/build.py \
...
     $BUILD_PARAM
 
-popd
+# Make sure that JAX depends on the local jaxlib installation
+# https://jax.readthedocs.io/en/latest/developer.html#specifying-dependencies-on-local-wheels
+line="jaxlib @ file://${BUILD_PATH_JAXLIB}"
+if ! grep -xF "${line}" "${SRC_PATH_JAX}/build/requirements.in"; then
+    pushd "${SRC_PATH_JAX}"
+    echo "${line}" >> build/requirements.in
+    bazel run //build:requirements_dev.update --repo_env=HERMETIC_PYTHON_VERSION="3.10"
```
gpupuck, line 278: Better to get the `3.10` version dynamically?

gspschmid (approved):

.github/container/build-jax.sh
```diff
@@ -275,14 +266,22 @@ time python build/build.py \
...
 
-popd
+# Make sure that JAX depends on the local jaxlib installation
+# https://jax.readthedocs.io/en/latest/developer.html#specifying-dependencies-on-local-wheels
+line="jaxlib @ file://${BUILD_PATH_JAXLIB}"
+if ! grep -xF "${line}" "${SRC_PATH_JAX}/build/requirements.in"; then
+    pushd "${SRC_PATH_JAX}"
+    echo "${line}" >> build/requirements.in
+    PYTHON_VERSION=$(python -c 'import sys; print("{}.{}".format(*sys.version_info[:2]))')
+    bazel run //build:requirements_dev.update --repo_env=HERMETIC_PYTHON_VERSION="${PYTHON_VERSION}"
```
gpupuck, line 279: One thing I wasn't so sure is that, is this setting `HERMETIC_PYTHON_VERSION` really necessary? I thought by default it's gonna use the system's python version. Could you try that? Other than that, everything else looks good!

gpupuck (approved):

.github/container/build-jax.sh
olupton: https://jax.readthedocs.io/en/latest/developer.html#specifying-python-version says it's only set automatically via `build.py`, which doesn't have a way of triggering ` requirements_dev.update`. The `_dev` version is needed to avoid putting hashes in the requirements file, which blocks using an editable installation.

PR #856, opened 2024-05-24
chore: Add `orbax-checkpoint` to manifest

ko3n1g:
Since they changed to vcs install

### DIFF
--- a/.github/container/manifest.yaml
+++ b/.github/container/manifest.yaml
@@ -175,4 +175,7 @@ panopticapi:
   tracking_ref: mute
   latest_verified_commit: a698a12deb21e4cf0f99ef0581b2c30c466bf355
   mode: git-clone
-  
+orbax-checkpoint:
+  url: https://github.com/google/orbax/#subdirectory=checkpoint
+  tracking_ref: main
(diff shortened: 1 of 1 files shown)


.github/container/manifest.yaml
```diff
@@ -175,4 +175,7 @@ panopticapi:
   tracking_ref: mute
   latest_verified_commit: a698a12deb21e4cf0f99ef0581b2c30c466bf355
   mode: git-clone
-  
\ No newline at end of file
+orbax-checkpoint:
+  url: https://github.com/google/orbax/#subdirectory=checkpoint
+  tracking_ref: main
+  latest_verified_commit: 16c2d409e365576284dbaf190ac002b24c1f927f
```
terrykong, line 181: Just to be consistent with the other vcs, you could tag it like so
```suggestion
  latest_verified_commit: 16c2d409e365576284dbaf190ac002b24c1f927f
  mode: pip-vcs
```

terrykong (changes requested):
Let's make sure the CI finishes to validate this works. I'm actually not sure this will work with the `#subdirectory=` syntax.

terrykong (approved):

gpupuck:
Nice catch!

PR #862, opened 2024-05-28
CI: Prevent metrics step from failing when tensorboard dir is not found

ashors1:
If a CI job fails before the tensorboard dir is created, it causes the entire metrics step to fail. Instead, we'd like for the metrics for only that job to fail and for the metrics to be checked for the remaining jobs.


terrykong (approved):

ashors1:
This CI run shows that the metrics step passes even though the multinode tests fail: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/9277815075

PR #863, opened 2024-05-29
nsys-jax: argparse+docs, fix post-processing

olupton:
This improves the command-line error messages and `--help`. The change to futures handling fixes some non-deterministic behaviour that sometimes stopped source files from being included in the output archive.

### DIFF
--- a/.github/container/nsys-jax
+++ b/.github/container/nsys-jax
@@ -16,27 +16,6 @@ import tempfile
 import time
 import zipfile
 
-# The valid invocations of this wrapper are:
-#   nsys-jax exe [exe args ...]
-#   nsys-jax [nsys args ...] -- exe [exe args ...]
-# where if `exe args` contains `--` then the second form must be used.
-# The wrapper has a few extra arguments of its own, which are all prefixed with
-# --nsys-jax- and must be passed as "nsys args" in the form above.
-nsys_flags_and_cmd = sys.argv[1:]
-try:
-    limit = nsys_flags_and_cmd.index("--")
-    nsys_flags = nsys_flags_and_cmd[:limit]
-    application = nsys_flags_and_cmd[limit + 1 :]
-except ValueError:
-    # No --, everything is the application
-    nsys_flags = []
-    application = nsys_flags_and_cmd
-
-if shutil.which(application[0]) is None:
-    raise Exception(
-        f"{application[0]} not found by shutil.which. Did you remember to place -- between extra arguments for nsys and your application?"
-    )
-
 
 # Expand %q{ENV_VAR} if the variable is defined.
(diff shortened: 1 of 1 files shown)


.github/container/jax_nsys/python/jax_nsys/jax_nsys/analysis.py
```diff
@@ -1,10 +1,135 @@
...
+            # https://github.com/NVIDIA/nccl/blob/ab2b89c4c339bd7f816fbc114a4b05d386b66290/src/collectives.cc#L122
+            assert msg_size_bits % collective_size == 0
+            msg_size_bits //= collective_size
+        assert msg_size_bits % 8 == 0
+        total_msg_size += msg_size_bits // 8
+
+    # Calculate the correction factor from algorithm bandwidth to bus bandwidth, see:
+    # https://github.com/NVIDIA/nccl-tests/blob/master/doc/PERFORMANCE.md#bus-bandwidth
+    collective = inst.opcode.removesuffix("-start")
+    bw_correction, bus_correction = {
```
gspschmid, line 82: Not a fan of computing all cases to use a single one :-D I'd usually probably factor this into a helper. Fwiw, from Python 3.10 you get `match`, so this could become:
```
def collective_correction(kind: str, size: int) -> tuple[float, float]:
  match kind:
    case "all-gather": return (size, (size - 1) / size)
    case ...
    case _: assert False, f"unknown collective kind {kind}"
```

.github/container/jax_nsys/python/jax_nsys/jax_nsys/analysis.py
```diff
@@ -1,10 +1,135 @@
...
+    # S32 is a 32-bit type and so on.
+    # FIXME: does not handle FP8 yet
+    for prefix in ["BF", "C", "F", "S", "U"]:
+        if enum_name.startswith(prefix):
+            return int(enum_name[len(prefix) :])
+    raise Exception(f"Could not deduce size of {enum_name}")
+
+
+@functools.lru_cache
+def get_message_size(program_id: int, instruction: str) -> int:
```
gspschmid, line 33: Returns a dataframe, not an int, no?

gspschmid (approved):
lgtm, minor suggestions

.github/container/jax_nsys/python/jax_nsys/jax_nsys/analysis.py
olupton: Right 🤦  -- all the more reason to actually set up some CI to lint this.

olupton:
Thanks, will merge this now so it ships in the 05-30 containers and fix those minor issues in a follow-up

PR #885, opened 2024-06-06
protoc: install a version older than google.protobuf

olupton:
See https://protobuf.dev/support/cross-version-runtime-guarantee: the generator cannot be newer than the runtime.

The versioning scheme is complicated, with the two components following different schemes, but the current snapshot of latest versions with 5.27.0 for `protobuf` on pypi.org (the runtime) and 27.1 for `protoc` at https://github.com/protocolbuffers/protobuf/releases/tag/v27.1 are not compatible.

This uses the pypi.org and github.com release dates to take the newest version with the correct ordering.

### DIFF
--- a/.github/container/jax_nsys/nsys-jax-ensure-protobuf
+++ b/.github/container/jax_nsys/nsys-jax-ensure-protobuf
@@ -46,12 +46,9 @@ runtime_date = min(dates)
 # Get the newest protoc version that isn't newer than that
 r = s.get("https://api.github.com/repos/protocolbuffers/protobuf/releases")
 r.raise_for_status()
-releases = r.json()
-for release in releases:
-  rdate = datetime.datetime.strptime(release["published_at"], '%Y-%m-%dT%H:%M:%SZ')
-  if rdate > runtime_date:
-    continue
-  break
+def is_old_enough(release):
+  return datetime.datetime.strptime(release["published_at"], '%Y-%m-%dT%H:%M:%SZ') <= runtime_date
+release = next(filter(is_old_enough, r.json()))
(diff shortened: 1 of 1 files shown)


.github/container/jax_nsys/nsys-jax-ensure-protobuf
```diff
@@ -27,19 +27,38 @@ echo "INFO: installing into virtual environment ${VIRTUALENV}"
...
+# Get the newest protoc version that isn't newer than that
+r = s.get("https://api.github.com/repos/protocolbuffers/protobuf/releases")
 r.raise_for_status()
+releases = r.json()
+for release in releases:
+  rdate = datetime.datetime.strptime(release["published_at"], '%Y-%m-%dT%H:%M:%SZ')
+  if rdate > runtime_date:
+    continue
+  break
+print(f"Installing {release['name']}")
```
gspschmid, line 55: Nit: I think many Python folks might disagree, but I find something functional like
```python
def is_old_enough(release):
  return datetime.datetime.strptime(release["published_at"], '%Y-%m-%dT%H:%M:%SZ') > runtime_date
release = next(filter(is_old_enough, r.json()))
```
easier to read

gspschmid (approved):

.github/container/jax_nsys/nsys-jax-ensure-protobuf
olupton: Thanks, done in da8435c9a48c65fb3b19f39f48f75378fa618644 (with a fix: > should be <=)

gspschmid (approved):

PR #899, opened 2024-06-13
Fix Triton build again

olupton:
AMD headers are included unconditionally; the simplest approach is to make sure they exist.


DwarKapex:
LGTM.
Did you file a bug to upstream?

olupton:
> Did you file a bug to upstream?

No, I don't think there's anyone that would want to fix this for us, so I'm not sure who I could file a bug with.

The problem is that (openxla/triton, bazel) typically works, and is used in XLA, and (openai/triton, cmake+wheel) typically works, but here we are trying to build (openxla/triton, cmake+wheel), and this sometimes breaks.

PR #902, opened 2024-06-14
nsys-jax: example of inspecting heap traces in compile-time metadata

olupton:


nouiz:
Maybe mention in the documentation that now there is a section about compiletime and one about heap memory usage?

PR #910, opened 2024-06-18
Install google.protobuf and protoc in containers

olupton:
Widely varying versions of `google.protobuf` (the Python runtime) are installed in the containers depending on which other packages are installed. The approach here is to let `pip-finalize.sh` install whatever `google.protobuf` version it wants, and then try to install a compatible `protoc` (the compiler / `.proto` -> `.py` translator).

This should allow some `virtualenv` hackery to be removed, and also avoid hitting GitHub API rate limits.


gspschmid (approved):
LGTM

PR #911, opened 2024-06-19
Update DALI pipeline for ViT

awolant:
This PR changes the DALI pipeline for ViT. It is no longer necessary to combine two pipelines together and this PR applies this to ViT pipeline.


terrykong:
@ashors1 to review and also run internal CI on this change

ashors1 (approved):

PR #921, opened 2024-06-25
Add maxtext for arm64 dockerfile.

nouiz:
Only needed special handling for tensorflow-text is needed.
Othewise, the 2 files are the same.


gpupuck:
@DwarKapex @olupton Please help review.

olupton (changes requested):
The CI config needs to be updated to actually build the maxtext container on arm64

olupton (approved):
I changed the few lines needed to enable the arm64 container build

PR #924, opened 2024-06-26
nsys-jax-combine: add report-merging script

olupton:
Also reorganise the data loading code to yield a more structured data frame format that's better suited to multi-profile analysis. This is documented inline: see the rendered notebook for a more readable version than the diff: https://gist.github.com/nvjax/9971bf22504f85e8fdff7559ae97efe7#file-analysis-ipynb


.github/container/jax_nsys/Analysis.ipynb
```diff
@@ -57,9 +57,99 @@
...
+    "The columns are as follows:\n",
+    "- `Name`: the name of the XLA module; this should always be the same for a given `ProgramId`\n",
+    "- `ProjStartNs`: the timestamp of the start of the module execution on the GPU, in nanoseconds\n",
+    "- `ProjDurNs`: the duration of the module execution on the GPU, in nanoseconds\n",
+    "- `OrigStartNs`: the timestamp of the start of the module launch **on the host**, in nanoseconds. *i.e.* `ProjStartNs-OrigStartNs` is something like the launch latency of the first kernel\n",
+    "- `OrigDurNs`: the duration of the module launch **on the host**, in nanoseconds\n",
+    "\n",
+    "The other profile data frame for GPU execution is `thunk_df`, which has a single row for each XLA thunk.\n",
+    "Loosely, each XLA module contains a series of thunks, and each thunk launches a GPU kernel.\n",
+    "In reality, thunks can be nested and may launch multiple kernels, but this data frame still provides the most granular distribution available of GPU execution time across the XLA module:"
```
gspschmid, line 106: Does the data ever contain both a thunk that has children _and_ one or multiple of its children? How would I only operate on the "leaf" thunks?

gspschmid (approved):
Looks great!

.github/container/jax_nsys/Analysis.ipynb
olupton: Answers: "no" and "trivially; those are the only ones present in the frame"

This does mean that you'd need to do some leg work to recover the intermediate thunks, of which the most obvious example is a `while` op:
Screenshot 2024-07-01 at 15 29 15
there will be no `while.5.0` rows in the thunk frame.

PR #975, opened 2024-08-02
Remove grok

DwarKapex:


terrykong (changes requested):
It looks like this PR also removed pax patch files and their reference in the manifest for some reason. Could you confirm if that was intentional?

DwarKapex:
> It looks like this PR also removed pax patch files and their reference in the manifest for some reason. Could you confirm if that was intentional?

I tried to remove everything, that has any connection to Grok. For instance, [paxml/PR-93.patch](https://github.com/NVIDIA/JAX-Toolbox/pull/975/files#diff-3b070c4c1860231e93952bd8730ebf2e0b3aa521473fc74838bf8cf23ec3c5b2) updates the Grok classes. I'm under impression, that this is to support Grok.
Also [praxis/PR-74.patch](https://github.com/NVIDIA/JAX-Toolbox/pull/975/files#diff-b6ca750e56dd219c7508818315fa787af2e79eddccf905c3a5c9b187ea1520ca) does some patches for Grok, but I'm not entirely sure I need to remove the whole patch or to do it partially.
WDYT?

.github/container/manifest.yaml
```diff
@@ -39,7 +39,6 @@ paxml:
   mode: git-clone
   patches:
     pull/46/head: file://patches/paxml/PR-46.patch # adds Transformer Engine support
-    pull/93/head: file://patches/paxml/PR-93.patch # exposes TE DPA argument in Grok to make it configurable from the command line
```
terrykong, line 42: You can revert this

.github/container/manifest.yaml
```diff
@@ -49,7 +48,6 @@ praxis:
   patches:
     pull/27/head: file://patches/praxis/PR-27.patch # This PR allows XLA:GPU to detect the MHA pattern more easily to call fused kernels from cublas.
     pull/36/head: file://patches/praxis/PR-36.patch # adds Transformer Engine support
-    pull/74/head: file://patches/praxis/PR-74.patch # experimental support for using TE FMHA in GQA
```
terrykong, line 52: you can revert this

.github/container/patches/paxml/PR-93.patch
terrykong, line 1: you can revert deleting this

.github/container/patches/praxis/PR-74.patch
terrykong, line 1: you can revert deleting this

terrykong (changes requested):
You can revert/leave the pax stuff. The patches are orthogonal to the grok container since that's a different codebase entirely.

olupton (approved):

PR #979, opened 2024-08-05
test-maxtext.sh: set default mem fraction to 0.9

sergachev:


nouiz (approved):

kocchop:
I was wondering if we can set the default to 0.90 and then we can remove the mem-fraction argument from examples altogether... the CI would run fine

sergachev:
done

PR #982, opened 2024-08-06
fix remat policy and fused attn variable in test-maxtext.sh

kocchop:
1. Fixed the remat policy typo `minimal_flash` 
2. Added few addition args to tune the perf
3. Fixed the enable fused attn variable

### DIFF
--- a/.github/container/test-maxtext.sh
+++ b/.github/container/test-maxtext.sh
@@ -33,10 +33,10 @@ usage() {
     echo "  -h, --help                 Print usage. Some examples:  
                                        1. test-maxtext.sh -b 2 --model-name=gpt3-52k
                                        2. test-maxtext.sh -b 2 --model-name=gemma-2b --dtype=fp8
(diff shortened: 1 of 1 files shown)


nouiz (approved):

.github/container/test-maxtext.sh
```diff
@@ -163,6 +164,10 @@ if [ $MODEL != "gpt3-52k" ]; then # gpt3-52k only works with dot_product
     ADDITIONAL_ARGS+=" attention=${ATTN_TYPE}"
 fi
 
+if [ ${ATTN_TYPE} == "cudnn_flash_te" ]; then
```
nouiz, line 167: Assert that it is a value that is supported?

kocchop: No, TE requires us to set the `NVTE_FUSED_ATTN=1` in order to use the fused attention api. This check is to set the environment variable. For other attn types, we don't need this

PR #988, opened 2024-08-09
nsys-jax: roll back to 2024.4.1 and handle single-gpu profiles of single module executions better

olupton:
- Nsight Systems 2024.5.1 is not currently supported by `nsys-jax`, roll it back while fixes are being prepared.
- Handle profiles that do not contain any communication (i.e. single GPU runs) bettter.
- Handle profiles that do not satisfy the warmup-removal heuristics more elegantly.


gspschmid (approved):
LGTM

PR #991, opened 2024-08-09
Old jaxlib install path

DwarKapex:


.github/container/Dockerfile.jax
```diff
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1-labs
 ARG BASE_IMAGE=ghcr.io/nvidia/jax:base
-ARG BUILD_PATH_JAXLIB=/opt/jax/dist
+ARG BUILD_PATH_JAXLIB=/opt/jaxlib
```
nouiz, line 3: nitpik jaxlibs as there is multiple one? to help make the difference with the one named jaxlib?

olupton: Done in e7837679f75c006d8c360cc3ce35ebdad75ec272 so this can get into today's nightlies.

olupton (approved):

PR #997, opened 2024-08-14
Dockerfile.jax: install JAX for TE build

olupton:
https://github.com/NVIDIA/TransformerEngine/pull/946 made TE's build process expect to be able to `import jax`.

### DIFF
--- a/.github/container/Dockerfile.jax
+++ b/.github/container/Dockerfile.jax
@@ -48,6 +48,7 @@ RUN ARCH="$(dpkg --print-architecture)" && \
 # TODO: move this patch into the manifest
 ADD xla-arm64-neon.patch /opt
 RUN build-jax.sh \
+    --bazel-cache ${BAZEL_CACHE} \
     --build-path-jaxlib ${BUILD_PATH_JAXLIB} \
     --src-path-jax ${SRC_PATH_JAX} \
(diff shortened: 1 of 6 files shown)


.github/container/Dockerfile.jax
```diff
@@ -54,14 +56,24 @@ RUN build-jax.sh \
     --xla-arm64-patch /opt/xla-arm64-neon.patch \ 
     --clean
 
+## Transformer engine: check out source and build wheel
+RUN <<"EOF" bash -ex -o pipefail
+pip install ninja && rm -rf ~/.cache/pip
+# TransformerEngine now needs JAX at build time
+pip install -e ${SRC_PATH_JAX}
```
olupton, line 63: This line can actually be deleted because `build-jax.sh` does install JAX inside this `builder` stage.

gpupuck:
It looks good, thanks for figuring out the symlinks solution!

The only suggestion is that could we have a separate script to create the link farm for all the necessary libraries? (Even in `build-jax.sh` could be okay) Because developers may have their own way to install a different cuDNN to test from the developer website. Having an individual script (or simply in `build-jax.sh`) could make the dev/test process easier.

DwarKapex (approved):

gpupuck (approved):
We've decided to take care of the link farm creation in a separate change.

PR #999, opened 2024-08-16
Try hermetic CUDA builds (matching system installed CUDA and cuDNN version)

gpupuck:
This is only a temporary WAR to unblock the JAX and DLFW image builds. The resulting image sizes do not change, compared to the ones we built using system CUDA libs.

### DIFF
--- a/.github/container/build-jax.sh
+++ b/.github/container/build-jax.sh
@@ -166,8 +166,12 @@ export TF_CUDNN_PATHS=/usr/lib/$(uname -p)-linux-gnu
 export TF_CUDA_VERSION=$(ls /usr/local/cuda/lib64/libcudart.so.*.*.* | cut -d . -f 3-4)
 export TF_CUDA_MAJOR_VERSION=$(ls /usr/local/cuda/lib64/libcudart.so.*.*.* | cut -d . -f 3)
 export TF_CUBLAS_VERSION=$(ls /usr/local/cuda/lib64/libcublas.so.*.*.* | cut -d . -f 3)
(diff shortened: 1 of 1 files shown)


nouiz (approved):

.github/container/build-jax.sh
```diff
@@ -262,10 +266,8 @@ time python "${SRC_PATH_JAX}/build/build.py" \
     --enable_cuda \
     --build_gpu_plugin \
     --gpu_plugin_cuda_version=$TF_CUDA_MAJOR_VERSION \
-    --cuda_path=$TF_CUDA_PATHS \
-    --cudnn_path=$TF_CUDNN_PATHS \
-    --cuda_version=$TF_CUDA_VERSION \
-    --cudnn_version=$TF_CUDNN_VERSION \
+    --bazel_options=--repo_env=HERMETIC_CUDA_VERSION=$CUDA_VERSION \
```
nouiz, line 269:
```suggestion
    --bazel_options=--repo_env=HERMETIC_CUDA_VERSION=$TF_CUDA_VERSION \
```

nouiz (changes requested):

.github/container/build-jax.sh
gpupuck: Sadly this won't work because `TF_CUDA_VERSION` will be `12.5`, the `CUDA_VERSION` from container env is `12.5.0` and `HERMETIC_CUDA_VERSION` requires to be a full version.

nouiz (approved):

PR #1013, opened 2024-08-22
Add CI argument for user-defined CUDA base image

yhtang:
Due to a recent need to create CUDA 12.4 images for a workstation that does not have forward compat...

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,6 +18,11 @@ on:
         description: Publish dated images and update the 'latest' tag?
         default: false
         required: false
+      CUDA_IMAGE:
+        type: string
+        description: CUDA image to use as base, e.g. nvidia/cuda:X.Y.Z-devel-ubuntu22.04
+        default: 'latest'
+        required: false
       BUMP_MANIFEST:
         type: boolean
         description: Bump git repos in manifest.yaml to head of tree?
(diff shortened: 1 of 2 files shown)


terrykong (dismissed):

.github/workflows/ci.yaml
```diff
@@ -28,6 +28,11 @@ on:
         description: "(used if BUMP_MANIFEST=true) If true: attempt to PR/merge manifest branch"
         default: false
         required: false
+      CUDA_IMAGE:
+        type: string
+        description: CUDA image to use as base, e.g. nvidia/cuda:X.Y.Z-devel-ubuntu22.04
+        default: 'latest'
```
olupton, line 34: Either you need to add special treatment of the "latest" value, or put the actual default value here instead of in the Dockerfile (https://github.com/NVIDIA/JAX-Toolbox/blob/400d83af7df9f93a5271d7f0c8daab7a47278ae8/.github/container/Dockerfile.base#L2)

olupton (changes requested):

.github/workflows/ci.yaml
yhtang: The logic was already in place from a long time ago: https://github.com/NVIDIA/JAX-Toolbox/blob/123e033ae794a5f1dbac566885d0fc2441cfb2ca/.github/workflows/_build_base.yaml#L135
😁 

It was also the conclusion from a previous discussion that we should not repeat the default value multiple times in the JAX-Toolbox code base. Placing it in the Dockerfile is the only option so that:
1. we have a single occurrence of this default value, while still make sure
2. the dockerfile is self-contained.

yhtang: Wait. I just realized that the logic is somehow broken. Working on a fix.

olupton (approved):

.github/workflows/ci.yaml
DwarKapex: Fixed it? =)

.github/workflows/ci.yaml
```diff
@@ -114,6 +120,17 @@ jobs:
             exit 1
           fi
       
+      - name: Determine CUDA image to use
+        id: cuda-image
+        shell: bash -x -e {0}
+        run: |
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
```
DwarKapex, line 127: Not quite understand the purpose of this if-statement. The default value is `latest` anyways. Or am I missing something?

PR #1021, opened 2024-08-29
Ensure jaxlibs folder is readable to all users in container

mgoldfarb-nvidia:
Containers that run with users other than root could not import jax due to restrictive permissions on the installed jaxlibs.


olupton:
The CI won't work with PRs from forks, but maybe https://github.com/NVIDIA/JAX-Toolbox/pull/1023 also covers this?

PR #1032, opened 2024-09-05
fix tensorboard events dir path

kocchop:
Fixed the tensorboard dir path after a recent change in MaxText software:
https://github.com/google/maxtext/pull/863


terrykong (approved):
Approved; assuming CI passes

PR #1045, opened 2024-09-17
Add an option to test-pax.sh to enable XLA cuDNN flash attention

Cjkkkk:
Provide an option to run XLA cuDNN flash attention as an alternative to TE cuDNN flash attention.


Cjkkkk:
Some trailing whitespace are removed as well. Let me know if that is intended.

nouiz (approved):

PR #1054, opened 2024-09-20
Increase SLURM job timeout

DwarKapex:
There are many unit tests that are run on OCI SLURM cluster, so it's natural, that some jobs are timed out duty to amount of scheduled jobs. Need to increase time out to be able to run all jobs.


nouiz:
Is this the timeout of one job own duration?
Or the wait in the queue?

terrykong (approved):

olupton:
> Is this the timeout of one job own duration? Or the wait in the queue?

It looks to me like it's increasing the test execution duration, not the maximum queued time. That doesn't make much sense to me.

PR #1058, opened 2024-09-23
Add JetStream to MaxText container

DwarKapex:


.github/container/Dockerfile.maxtext.amd64
kocchop, line 1: please rebase it on top https://github.com/NVIDIA/JAX-Toolbox/pull/1070

.github/workflows/_ci.yaml
kocchop, line 1: LGTM for now, will provide a jetStream inference test with MaxText soon

kocchop (commented):
LGTM, please rebase it on top of https://github.com/NVIDIA/JAX-Toolbox/pull/1070

kocchop:
In addition to the `pip install -r requirements.txt` we also need to add `pip install .` from the `JetStream` repo. The `google-jetstream` that is installed as part of maxtext requirements is not the latest. The above step would uninstall it and install the most recent compiled one from the repo.

.github/container/Dockerfile.maxtext.arm64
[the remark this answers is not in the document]
yhtang: rebased on top of the current main, which should contain #1070.

yhtang:
> In addition to the `pip install -r requirements.txt` we also need to add `pip install .` from the `JetStream` repo. The `google-jetstream` that is installed as part of maxtext requirements is not the latest. The above step would uninstall it and install the most recent compiled one from the repo.

Do you mean to add something like
```
echo "-e ${SRC_PATH_JETSTREAM}" >> /opt/pip-tools.d/requirements-jetstream.in
```
?

yhtang (changes requested):
See comment above

.github/container/Dockerfile.maxtext
```diff
@@ -26,6 +30,12 @@ for pattern in \
     sed -i "${pattern}" ${SRC_PATH_MAXTEXT}/requirements.txt;
 done
 echo "tensorflow-metadata>=1.15.0" >> ${SRC_PATH_MAXTEXT}/requirements.txt
+# remove outdated version of JetStream in favor of the upstream JetStream
+sed -i "/google-jetstream/d" ${SRC_PATH_MAXTEXT}/requirements.txt
+
+# Add the upstream JetStream
+git-clone.sh ${URLREF_JETSTREAM} ${SRC_PATH_JETSTREAM}
+echo "-r ${SRC_PATH_JETSTREAM}/requirements.txt" >> /opt/pip-tools.d/requirements-jetstream.in
```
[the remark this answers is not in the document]
olupton: Agreed, this should be more like
```
echo "-e file://${SRC_PATH_JETSTREAM}" > /opt/pip-tools.d/requirements-jetstream.in
```
but I think we do already do something similar for MaxText, I'm not sure why: https://github.com/NVIDIA/JAX-Toolbox/blob/20d3137c973225d0f14d5358f3be2407bf9ccbce/.github/container/Dockerfile.maxtext#L17C12-L17C28

But note that the build is failing, `pip` will need some hand-holding.

.github/workflows/_ci.yaml
```diff
@@ -689,3 +690,35 @@ jobs:
...
+    needs: build-maxtext
+    if: inputs.ARCHITECTURE == 'amd64' # no arm64 gpu runners
+    uses: ./.github/workflows/_test_unit.yaml
+    with:
+      TEST_NAME: jetstream
+      EXECUTE: |
+        docker run --shm-size=1g --gpus all ${{ needs.build-maxtext.outputs.DOCKER_TAG_FINAL }} \
+        bash <<"EOF" |& tee test-jetstream.log
+          cd /opt/jetstream
+          pip install -r requirements.txt
```
olupton, line 704: Why is this needed?

.github/workflows/_ci.yaml
```diff
@@ -689,3 +690,35 @@ jobs:
...
+          export CUDA_VISIBLE_DEVICES=0
```
olupton, line 705: Can we test multi-GPU functionality?

olupton (changes requested):

.github/workflows/_ci.yaml
DwarKapex: removed

.github/workflows/_ci.yaml
DwarKapex: I did not find JetStream unit-test examples on GitHub for multi-GPU. Even more: afaik, it does not support GPUs so far. But it will start GPU support in the nearest future.

PR #1065, opened 2024-09-26
Add pathwaysutils for MaxText to manifest file

DwarKapex:
The latest MaxText uses `pathwayutils`, which is added as a dependency. Need to add it to our manifest.yaml file to resolve reference issue during final installation.


kocchop (approved):
LGTM, this change should fix maxtext build failure

PR #1073, opened 2024-10-01
nsys-jax post-processing: treat host-device copies as 1-device collectives

olupton:
This adds logic to treat `dynamic[-update]-slice` operations that have a source/destination operand in the host memory space as being communication operations, labelling them as single-device "collectives".

The goal is to improve support for analysing profiles of execution including offloading to host memory.

Also fix using nsys 2024.6 by applying the same patch as 2024.5 that adds the thread ID.

### DIFF
--- a/.github/container/jax_nsys/python/jax_nsys/jax_nsys/analysis.py
+++ b/.github/container/jax_nsys/python/jax_nsys/jax_nsys/analysis.py
@@ -6,7 +6,7 @@ import pandas as pd  # type: ignore
 import pathlib
 from typing import Any
 
-from .protobuf import HloProto, xla_module_metadata
+from .protobuf import HloInstruction, HloProto, _host_memory_space, xla_module_metadata
 from .utils import make_child_mask, ProfilerData
 
 pd.options.mode.copy_on_write = True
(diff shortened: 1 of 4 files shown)


.github/container/jax_nsys/python/jax_nsys/jax_nsys/analysis.py
```diff
@@ -193,13 +198,51 @@ def _get_message_size(
...
+            transfer_size = _byte_size(comm_inst)
+        dest_on_host = _host_memory_space(comm_inst)
+        src_on_host = _host_memory_space(src_inst.proto())
+        assert src_on_host != dest_on_host, (
+            'dynamic[-update]-slice is only considered is only "communication" if it '
+            "represents a host-device transfer"
+        )
+        return (
+            transfer_size,
+            "device-to-host" if dest_on_host else "host-to-device",
```
zhenying-liu, line 241: Can we exclude dynamic-update-slice and dynamic-slice if they don't have host memory space color 5 on neither source nor destination?

olupton: There's an assert just above:
https://github.com/NVIDIA/JAX-Toolbox/blob/06ba7b3ec99abda2bbaff02da5d3959646dab945/.github/container/jax_nsys/python/jax_nsys/jax_nsys/analysis.py#L235-L238
that exactly one of source/destination is host memory (5), and it should never fail because of the logic here:
https://github.com/NVIDIA/JAX-Toolbox/blob/06ba7b3ec99abda2bbaff02da5d3959646dab945/.github/container/jax_nsys/python/jax_nsys/jax_nsys/protobuf.py#L82-L89
that only labels the op as being communication if the same condition is met.

zhenying-liu (approved):

PR #1075, opened 2024-10-02
Add missing port for ssh connection for upstream T5x test

DwarKapex:
Add missing ssh port for upstream T5x test


terrykong (approved):
LGTM - assuming tests pass

yhtang (changes requested):
Let's encode the port as a secret. Otherwise, it is visible to the outside world.

yhtang:
I just made this proposal to address the SSH port security issue for good: https://github.com/NVIDIA/JAX-Toolbox/issues/1093
@DwarKapex

PR #1076, opened 2024-10-02
README: make badges render consistently

olupton:
It seems that the `<picture>` tags used previously hit a flaky bit of GitHub rendering code and do not always display.
Making the badges be explicit hyperlinks to reasonable targets seems to avoid this flaky path and display the badges consistently.
In future, we might like to upload small `.md` summaries to the Gist and link those instead of the `.json` files.


sbhavani:
thanks for fixing the badges after the README update!

sbhavani (approved):

PR #1079, opened 2024-10-02
Do not add leading "//tests:" for given test targets

gpupuck:
The original logic is if the given test targets do not start with `//tests:`, then add this prefix for them.

I'm removing this logic as:
1. We don't actually see any use cases in the workflow
2. It will block us running per-directory suite targets such as `//tests/mosaic:gpu_tests`


andportnoy (approved):
Thanks, explicit is better than implicit.

PR #1010, opened 2024-08-21
remove deprecated XLA flag

kocchop:
1. `xla_gpu_enable_triton_gemm` is still needed. 
2. Removed some other deprecated XLA flags: `xla_gpu_enable_triton_softmax_fusion` 
3. Also removed some XLA flags that are now turned on by default. `xla_enable_async_all_gather` etc.

PR #1088, opened 2024-10-08
chore: remove deprecated flags from model xla flag configs

terrykong:
Applying the changes from https://github.com/NVIDIA/JAX-Toolbox/pull/1010/files

According to https://github.com/openxla/xla/blame/main/xla/debug_options_flags.cc, it looks like removing this is safe as these were the defaults as of today


nouiz (approved):
LGTM.

DwarKapex (approved):

PR #1097, opened 2024-10-13
ci[maxtext]: adds rosetta directory to maxtext containers

terrykong:
- makes xla flag env files available to users


DwarKapex (approved):

DwarKapex:
@terrykong You copy the whole rosetta folder. Is it intentional?

terrykong:
@DwarKapex It was intentional. I did a check and that dir is 2.2M. Would you rather I just copy in the maxtext directory?

PR #1106, opened 2024-10-16
disable goodput recorder in test-maxtest.sh

kocchop:
The goodput recorder was turned off by default and we do not use it. However, it has been turned on by default from MaxText side 2 days ago which is breaking the CI tests. We just need to explicitly disable these whenever we're running.

### DIFF
--- a/.github/container/test-maxtext.sh
+++ b/.github/container/test-maxtext.sh
@@ -245,24 +245,58 @@ RUN_NAME="logdir" ## the RUN_NAME cannot be changed
 if [ -z "$DECODER_BLOCK" ]; then
 
     # this part could be used to test different model ootb
-    RUN_SETTINGS="MaxText/train.py MaxText/configs/base.yml run_name=${RUN_NAME} model_name=${MODEL}\
-        steps=$STEPS per_device_batch_size=${BATCH_PER_GPU} remat_policy=${REMAT_POLICY} enable_checkpointing=false\
-        base_output_directory=$OUTPUT dataset_path=local dataset_type=synthetic hardware=$HARDWARE\
-        enable_goodput_recording=false monitor_goodput=false\
-        dcn_fsdp_parallelism=$dcn_FSDP ici_fsdp_parallelism=$ici_FSDP\
-        ici_data_parallelism=$ici_DP dcn_data_parallelism=$dcn_DP\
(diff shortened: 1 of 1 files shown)


gpupuck (dismissed):

.github/container/test-maxtext.sh
```diff
@@ -248,6 +248,7 @@ if [ -z "$DECODER_BLOCK" ]; then
     RUN_SETTINGS="MaxText/train.py MaxText/configs/base.yml run_name=${RUN_NAME} model_name=${MODEL}\
         steps=$STEPS per_device_batch_size=${BATCH_PER_GPU} remat_policy=${REMAT_POLICY} enable_checkpointing=false\
         base_output_directory=$OUTPUT dataset_path=local dataset_type=synthetic hardware=$HARDWARE\
+        enable_goodput_recording=false monitor_goodput=false\
```
terrykong, line 251: nit: I think it would be better to expand RUN_SETTINGS so that it's one key=value per line for readability and maintenance

terrykong (commented):
Other than my nit, LGTM

.github/container/test-maxtext.sh
kocchop: @terrykong Done!

gpupuck (approved):

terrykong (approved):

terrykong:
Thanks for the quick fix! Approved modulo tests passing

.github/container/test-maxtext.sh
```diff
@@ -245,22 +245,58 @@ RUN_NAME="logdir" ## the RUN_NAME cannot be changed
 if [ -z "$DECODER_BLOCK" ]; then
```
nouiz, line 245: @kocchop 
instead of ```if [ -z "$DECODER_BLOCK" ]; then ...```
It could be something like:
```if [ ! -z "$DECODER_BLOCK" ]; then DECODER_ARG="decoder_block=${DECODER_BLOCK}"; fi
RUN_SETTINGS="MaxText/train.py \
        MaxText/configs/base.yml \
        run_name=${RUN_NAME} \ 
        ${DECODER_ARG} \ 
       ....
```


This remove the duplicate code and make clear this is the only different in both condition.
Right now, it isn't clear if there is other diff then this one.

PR #1113, opened 2024-10-21
Add extra new line before .local_cuda.bazelrc

gpupuck:
Sometimes when there is no new line at the end of `.bazelrc` file, it will end up with `try-import %workspace%/.bazelrc.usertry-import %workspace%/.local_cuda.bazelrc`.

For example this commit did not contain a new line at the end: https://github.com/jax-ml/jax/commit/a2bc8c2e07912fac6d847423a7fbe42bdd33f42b


nouiz (approved):
LGTM, validate CI results.

PR #1114, opened 2024-10-22
Fix missing spaces in rsync commands

yhtang:
Only the one in _test_slurm_pyxis.yaml is critical; others are cosmetic.


yhtang:
I may merge this without review because it fixes an issue in #1103, which was merged before I could address it... Otherwise, all A100 tests will fail on the result collection step due to a misformed rsync command.

olupton (approved):

PR #1116, opened 2024-10-24
Update EFA/AWS-OFI-NCCL installation recipe

olupton:
This is not run automatically at present. EFA-enabled containers can be prepared by adding a new `RUN install-efa.sh` layer on top of the published JAX-Toolbox containers.


sbhavani (approved):
LGTM

PR #1122, opened 2024-10-28
base: CUDA 12.6.2 and Ubuntu 24.04

olupton:


gpupuck:
What's the motivation to create a venv?

olupton:
> What's the motivation to create a venv?

With the Python we pick up in Ubuntu 24.04, it will not allow `pip install` to system-wide directories without `--break-system-packages` or an explicit prefix.
If we use an explicit prefix (e.g. install `pip`-versioned software to `/usr/local` with Ubuntu-managed software in `/usr`), `pip uninstall` still complains (and `build-jax.sh` does that).

It's just a proposal, but explicitly separating the `pip`-managed software in `/opt/venv` doesn't seem unreasonable.

hmonishN:
You can also try PIP_BREAK_SYSTEM_PACKAGES=1 env var to fix this. We did the same and also DLFW teams as well.

PR #1125, opened 2024-10-29
jax-toolbox-triage: minor usability/doc improvements

olupton:
- Print the stdout/stderr of the first execution of the test case, which is supposed to fail, at INFO level along with a message encouraging the user to check that it is the correct failure.
- Print the path to the DEBUG log file at INFO level and, therefore, to the console.
- Expand the documentation.


.github/triage/jax_toolbox_triage/args.py
```diff
@@ -76,6 +84,15 @@ def parse_args():
             test case fails on this date.""",
         type=lambda s: datetime.date.fromisoformat(s),
     )
+    container_search_args.add_argument(
+        "--start-container",
```
nouiz, line 88: Rename to passing_container and failing_container to make it more clear which one must pass/fail?

docs/triage-tool.md
```diff
@@ -20,6 +20,10 @@ The tool follows a three-step process:
      failing, and a reference commit of XLA (JAX) that can be used to reproduce the
      regression.
 
+The third step can also be used on its own, via the `--start-container` and
+`--end-container` options, which allows it to be used between private container tags,
+without the dependency on the `ghcr.io/nvidia/jax` registry.
```
nouiz, line 25: Maybe specify that we suppose those container are based on our JAX-Toolbox container and  so has XLA under /opt/xla, JAX under /opt/jax and has the build-jax.sh script? If I forgot other expectations, just add them.

docs/triage-tool.md
```diff
@@ -46,11 +50,15 @@ needed to execute the test case.
 
 ## Usage
 
-To use the tool, there are two compulsory arguments:
-   * `--container`: which of the `ghcr.io/nvidia/jax:CONTAINER-YYYY-MM-DD` container
-     families to execute the test command in. Example: `jax` for a JAX unit test
-     failure, `maxtext` for a MaxText model execution failure
+To use the tool, there are two compulsory inputs:
    * A test command to triage.
+   * A specification of which containers to triage in:
```
nouiz, line 55: It isn't clear to me that it is a choice between both option bellow. Can this sentence clarify that?

.github/triage/jax_toolbox_triage/args.py
olupton: I was trying to keep it aligned with `--start-date` and `--end-date`, but I guess you're right it's clearer. Changed.

docs/triage-tool.md
olupton: Done. I don't immediately know of any others.

nouiz (approved):

PR #1133, opened 2024-10-31
Connect AWS/P5 (H100) instances to CI

olupton:
Make use of _test_slurm_pyxis.yaml a little less verbose, and add a new GPU_TYPE input to it. Also change the default to non-exclusive allocation, with exclusive allocation available on request.

### DIFF
--- a/.github/workflows/_test_slurm_pyxis.yaml
+++ b/.github/workflows/_test_slurm_pyxis.yaml
@@ -2,28 +2,10 @@ name: ~test multi-node jobs via SLURM+Pyxis
 
 on:
   workflow_call:
-    secrets:
-      SSH_PRIVATE_KEY:
-        required: true
-        description: SSH private key for accessing the SLURM login node
-      SLURM_LOGIN_USER:
-        required: true
-        description: Username for the SLURM login node
-      CONTAINER_REGISTRY_TOKEN:
-        required: true
-        description: Token for accessing the container registry
     inputs:
       NAME:
         type: string
         description: Name of the test case and output artifact tarball
-        required: true
-      SLURM_LOGIN_HOSTNAME:
-        type: string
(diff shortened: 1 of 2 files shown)


mjsML (changes requested):
This PR is making a massive assumption w.r.t scheduling. IIUC we are tearing down the SLURM cluster in AWS?

.github/workflows/_test_slurm_pyxis.yaml
```diff
@@ -2,29 +2,11 @@ name: ~test multi-node jobs via SLURM+Pyxis
 
 on:
   workflow_call:
-    secrets:
```
yhtang, line 5: Why would you remove it? It will make the workflow less adaptable to more clusters.

.github/workflows/_test_slurm_pyxis.yaml
```diff
@@ -78,15 +68,21 @@ on:
         value: ${{ jobs.run-test.outputs.SLURM_EXITCODE }}
 
 jobs:
-
   run-test:
     name: ${{ inputs.NAME }}
     runs-on: jumpbox
+    env:
+      OUTPUT_BASEDIR: /nfs/cluster
+      SLURM_CLUSTER: ${{ inputs.GPU_TYPE == 'A100' && secrets.CLUSTER_LOGIN_USER || secrets.AWS_LOGIN_USER }}@${{ inputs.GPU_TYPE == 'A100' && vars.HOSTNAME_SLURM_LOGIN || vars.HOSTNAME_AWS_SLURM }}
```
yhtang, line 76: This logic, if necessary, shall be better placed in the calling workflow to maximize reusability of this workflow (which is essentially a function to be called by the master workflow).

yhtang (changes requested):

PR #1151, opened 2024-11-12
CI: example job using EKS jumphost

olupton:


.github/workflows/_ci.yaml
```diff
@@ -399,6 +399,180 @@ jobs:
...
+                command:
+                - sh
+                - -c
+                - |
+                  install-efa.sh
+                  nsys-jax \
+                    --output=/opt/output/rank%q{JOB_COMPLETION_INDEX}.zip \
+                    -- \
+                    jax-nccl-test \
+                    --coordinator-address \
```
yhtang, line 490: Why not use a [JobSet](https://github.com/kubernetes-sigs/jobset)? I've added [automatic bootstrapping support for JAX in K8s jobset](https://github.com/jax-ml/jax/pull/22302). This can save the ClusterIP and bootstrapping boilerplates.

.github/workflows/_ci.yaml
```diff
@@ -399,6 +399,180 @@ jobs:
...
+              - name: ${{ github.run_id }}-${{ github.run_attempt }}-token
+              volumes:
+              - name: output
+                emptyDir: {}
+              - name: shmem
+                emptyDir:
+                  medium: Memory
+                  sizeLimit: 8Gi
+    steps:
+    - name: Login to GitHub Container Registry
```
yhtang, line 534: Is this still needed, as the container is pulled by the K8s job instead of the CI job itself?

.github/workflows/_ci.yaml
```diff
@@ -399,6 +399,180 @@ jobs:
         *-execution-combine.log
     secrets: inherit
 
+  test-nsys-jax-eks:
+    needs: build-jax
+    if: inputs.ARCHITECTURE == 'amd64' # arm64 runners n/a
+    runs-on: eks
+    env:
+      POST_PROCESS_JOB_DESCRIPTION: |
```
yhtang, line 407: How about saving the YAML specs as actual files under `.github/workflows`? You could also reuse something from https://github.com/jax-ml/jax/pull/24197/files although it is still not merged yet...

yhtang (changes requested):

.github/workflows/_ci.yaml
yhtang: For example of how to run a jobset, see https://github.com/jax-ml/jax/pull/24197/files

.github/workflows/_ci.yaml
olupton: We could do - the main counter-argument is that you then need some other solution for value substitution, whereas the status quo just uses GitHub Actions's substitution directly.

.github/workflows/_ci.yaml
olupton: I believe it's needed to generate the secret that is saved to Kubernetes in the next step.

.github/workflows/_ci.yaml
olupton: Given that we are not currently exposing the code that configures the Kubenetes cluster itself, I was reluctant to go in a direction where the public code (CI config) does not work without non-trivial private logic (JobSet installation, service account configuration, ...)

Also, the bootstrapping doesn't work out of the box because it relies on Python code that is not shipped in our containers. Pursuing that felt like a tangent to the core of this PR.

.github/workflows/_ci.yaml
yhtang: Yes, either `envsubst` or `yq` should work. I find this template-substitution pattern to provide better readability and reusability in the long run.

.github/workflows/_ci.yaml
yhtang: > non-trivial private logic (JobSet installation, service account configuration, ...)

The service account logic is pretty straightforward --- it's not gonna rank in the top N most hairy K8s configs in the world. Plus, we already provided a config that works OOTB!

It's one line to [install jobset](https://github.com/kubernetes-sigs/jobset?tab=readme-ov-file#installation):
```
kubectl apply --server-side -f https://github.com/kubernetes-sigs/jobset/releases/download/v0.7.1/manifests.yaml
```

That said, given that Jobs are 'built-in', maybe we should consider making `jax.distributed.initialize` automatic under it too.

.github/workflows/_ci.yaml
yhtang: I see. Makes sense.

.github/workflows/_ci.yaml
yhtang: > not currently exposing the code that configures the Kubenetes cluster

I wouldn't view the read-only service account and the installation of Jobset as the _sensitive_ information of a cluster. It's the equivalence of one `usermod -aG ...` and one `apt install` for a SLURM cluster.

olupton: I'm not saying we can't expose it because it's sensitive (it's not). But it's not currently exposed, and it won't be in this PR, so it can't be referred to. It would need to be checked what else is needed re: the AWS/EKS-specific parts of how the roles are managed, too.

> Plus, we already provided a config that works OOTB!

You mean the minikube CI PR to JAX? That's not really discoverable?

The error message still includes a placeholder, BTW: https://github.com/yhtang/jax/blob/2f67710e8ce9cd6c872f85e57417aa9bba7aa270/jax/_src/clusters/k8s_cluster.py#L70

.github/workflows/_ci.yaml
olupton: I reworked it with `yq`

yhtang (approved):
LGTM, except for the headless service is not deleted when the job finishes. You should be able to easily fix it though.

.github/workflows/_ci.yaml
yhtang: > You mean the minikube CI PR to JAX? That's not really discoverable?

I meant my [k8s CI PR](https://github.com/jax-ml/jax/pull/24197/files#diff-3d163a134c02c96a805a5b71ff18a506bc164680cfe152f3806525a64fdeb28f) to upstream.

yhtang: > The error message still includes a placeholder, BTW

You are right... the plan is to upstream another docs-only PR once the CI PR merges.

yhtang (approved):

PR #1153, opened 2024-11-13
Allow multi-GPU tests and enable MPS

olupton:
- CI ran in 15m50s with this change (more tests), vs. 18m before

### DIFF
--- a/.github/container/test-jax.sh
+++ b/.github/container/test-jax.sh
@@ -113,8 +113,6 @@ for t in $*; do
 done
 
 TEST_TAG_FILTER_ARRAY=()
-TEST_TAG_FILTER_ARRAY+=('-multiaccelerator')
-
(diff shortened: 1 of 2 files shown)


.github/container/test-jax.sh
```diff
@@ -113,8 +113,6 @@ for t in $*; do
 done
 
 TEST_TAG_FILTER_ARRAY=()
-TEST_TAG_FILTER_ARRAY+=('-multiaccelerator')
```
nouiz, line 116: Doesn't that remove tests?

olupton: I believe that before this PR we were *not* running the `multiaccelerator` tests (because of the minus), and with this PR we *are* running them.

nouiz (approved):

yhtang (approved):

PR #1159, opened 2024-11-18
Install [dev] dependecies instead of [test] dependencies for gemma tests

DwarKapex:
Gemma poetry installation file was changed recently (moved all [test] stuff under [dev]), so need to reflect the changes in `test-gemma` job


olupton (approved):
Note that not all the tests are passing.
Approving because running with failures is better than not running.

PR #1165, opened 2024-11-21
nsys-jax: re-work to be more pip-install-able

olupton:
The overarching goal of this PR is to get closer to a world where the `nsys-jax` tooling is straightforwardly `pip install`-able. While the diff looks scary, it's mostly re-organisation.

Substantive changes:
- `nsys-jax` no longer bundles Python code in the output archives, the `install.sh` script provided for users to run on local machines becomes, loosely, `install 'pip nsys-jax[jupyter] @ git+https://github.com/NVIDIA/JAX-Toolbox.git@COMMIT#subdirectory=.github/container/nsys_jax'`, where `COMMIT` corresponds to the `nsys-jax` command that produced the archive. For the `ghcr.io/nvidia/jax` containers, this is the commit of JAX-Toolbox that triggered the container build.

Changes included:
- Introduce `/opt/pip-tools-post-install.d`, which `pip-finalize.sh` will execute the contents of *after* installing the `pip`-managed world
  - Migrate `install-protoc` to use this, so `pip-finalize.sh` can forget about that detail.
  - Install https://github.com/brendangregg/FlameGraph/blob/master/flamegraph.pl via this.
  - Patch the `nvtx_gpu_proj_trace` Python code in Nsight Systems 2024.5 and 2024.6 via this.
- Move `nsys-jax` installation (specifically for the containers) into `install-nsys-jax.sh` and thereby clean up `install-nsight.sh`. The new script has to be told the git commit hash of JAX-Toolbox that is being built, because `nsys-jax` bakes this into an installation script in its output `.zip` archives to ensure the local environment matches the profile-collection environment.
- The CLI tools like `nsys-jax`, `nsys-jax-combine` and `install-protoc` are now handled via `[project.scripts]` in `pyproject.toml` instead of being standalone Python scripts. This is "more standard", and also makes it easier to share code between `nsys-jax` and `nsys-jax-combine`.
- The Python library is renamed from `jax_nsys` to `nsys_jax` for consistency.
- It's now possible to set the default data loading path via the `NSYS_JAX_DEFAULT_PREFIX` environment variable; previously the default was the current working directory, but that can be inconvenient to steer in Jupyter environments.


.github/container/nsys_jax/nsys_jax/scripts/nsys_jax.py
```diff
@@ -0,0 +1,799 @@
...
+  # Pick up the current profile data by default
+  export NSYS_JAX_DEFAULT_PREFIX="${{PWD}}"
+  # https://setuptools.pypa.io/en/latest/userguide/datafiles.html#accessing-data-files-at-runtime
+  NOTEBOOK=$("${{BIN}}/python" -c 'from importlib.resources import files; print(files("nsys_jax") / "analyses" / "Analysis.ipynb")')
+  echo "Launching: cd ${{SCRIPT_DIR}} && ${{BIN}}/jupyter-lab ${{NOTEBOOK}}"
+  cd "${{SCRIPT_DIR}}" && "${{BIN}}/jupyter-lab" "${{NOTEBOOK}}"
+else
+  echo "Skipping launch of Jupyter Lab due to NSYS_JAX_INSTALL_SKIP_LAUNCH"
+fi
+"""
```
olupton, line 85: This replaces the old `install.sh`, otherwise the content of this file is mostly the old `nsys-jax`.

.github/container/nsys_jax/nsys_jax/scripts/nsys_jax.py
```diff
@@ -0,0 +1,799 @@
...
+
+
+def create_install_script(output_queue):
+    """
+    Write an install.sh to the output archive that installs nsys-jax at the same
+    version/commit that the current execution is using.
+    """
+    jax_toolbox_sha = jax_toolbox_sha_with_prefix[1:]
+    install_script = install_script_template.format(jax_toolbox_commit=jax_toolbox_sha)
+    output_queue.put(("install.sh", install_script.encode(), COMPRESS_DEFLATE))
```
olupton, line 95: This drives the install script generation. The `1:` slice is because the hash is prefixed with `g` (for `git`).

.github/container/nsys_jax/nsys_jax/scripts/nsys_jax_combine.py
```diff
@@ -0,0 +1,155 @@
...
+                            # input.stem use a standardised format showing the device
+                            # numbers that were profiled in reportN.zip.
+                            dst_info = copy.copy(member)
+                            dst_info.filename = filename + "/" + input.stem
+                            write(dst_info)
+        if len(args.analysis):
+            assert mirror_dir is not None
+            # Execute post-processing recipes and add any outputs to `ofile`
+            for analysis in args.analysis:
+                result, output_prefix = execute_analysis_script(
```
olupton, line 144: FWIW this has been factored out and shared between `nsys-jax` and `nsys-jax-combine`

.github/container/nsys_jax/pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+
+[tool.setuptools_scm]
+root = "../../.." # .github/container/nsys_jax
+# written into the git checkout in case of an editable installation
+version_file = "nsys_jax/version.py"
+# __sha__ is not written by default
+version_file_template = """\
+__version__ = version = {version!r}
+__version_tuple__ = version_tuple = {version_tuple!r}
+__sha__ = {scm_version.node!r}
```
olupton, line 43: This is propagated so that the `install.sh` in the `nsys-jax[-combine]` output `.zip` archives can install the same `nsys-jax` commit as was used to generate the archive.

.github/workflows/nsys-jax.yaml
```diff
@@ -111,18 +108,15 @@ jobs:
...
-          pushd combined/
-          ./nsys_jax_venv/bin/python -m jupyterlab --version
+          NOTEBOOK=$(python -c 'from importlib.resources import files; print(files("nsys_jax") / "analyses" / "Analysis.ipynb")')
+          # Point to the extracted nsys-jax-combine output
+          export NSYS_JAX_DEFAULT_PREFIX="${PWD}/combined"
           # Run with ipython for the sake of getting a clear error message
-          ./nsys_jax_venv/bin/ipython Analysis.ipynb
+          ipython "${NOTEBOOK}"
+
+  # TODO: add a test that generates + runs install.sh
```
olupton, line 119: This is still TODO, but could be deferred to a follow-up.

.github/container/nsys_jax/nsys_jax/scripts/install_protoc.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+    parser.add_argument(
+        "prefix", help="Output prefix under which to install protoc", type=str
+    )
+    args = parser.parse_args()
+
+    s = requests.Session()
+    s.mount("https://", requests.adapters.HTTPAdapter(max_retries=5))
+
+    # protobuf versioning is complicated, see protocolbuffers/protobuf#11123 for more
```
gspschmid, line 23: Didn't this logic already use to live somewhere else? I don't recall where though.

.github/container/nsys_jax/nsys_jax/scripts/nsys_jax.py
gspschmid: > The 1: slice is because the hash is prefixed with g (for git).

Maybe say that in a comment?

.github/container/nsys_jax/nsys_jax/scripts/utils.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    script_file = importlib.resources.files("nsys_jax").joinpath(
+        "analyses", script + ".py"
+    )
+    if script_file.is_file():
+        return script_file
+    assert os.path.exists(
+        script
+    ), f"{script} does not exist and is not the name of a built-in analysis script"
+    return contextlib.nullcontext(pathlib.Path(script))
```
gspschmid, line 46: Why make this a contextmanager? As it is, this could just be a function returning the path.

docs/nsys-jax.md
```diff
@@ -0,0 +1,123 @@
...
+$ pip install git+https://github.com/NVIDIA/JAX-Toolbox.git#subdirectory=.github/container/nsys_jax
+```
+or, for an editable install from a specific branch/tag/commit:
+```console
+$ pip install --src /checkout-dir -e 'git+https://github.com/NVIDIA/JAX-Toolbox.git@main#subdirectory=.github/container/nsys_jax&egg=nsys-jax'
+```
+You may want to include this in a global `pip-compile`-based dependency resolution (as is done in the containers built
+from this repository), rather than running too many ad-hoc `pip install` commands.
+
+This will install all of the components mentioned so far, but does not currently include all (implicit) dependencies:
```
gspschmid, line 44: > but does not currently include all (implicit) dependencies

->

> but does not currently include the following implicit dependencies

?

gspschmid (approved):
Looks great, a few minor comments.

.github/container/nsys_jax/nsys_jax/scripts/install_protoc.py
olupton: It moved from here: https://github.com/NVIDIA/JAX-Toolbox/pull/1165/files#diff-634c5bb0e37da5e8d8aaebb04448e80c62f14a16b902a3e9931a7054038bcdb0L21-L32
This file is mainly a rename of `install-protoc`.

.github/container/nsys_jax/nsys_jax/scripts/utils.py
olupton: It's because the other branch (`return script_file`) above returns a context manager from `importlib.resources`. The extra indirection would allow `importlib.resources` to extract the relevant `.py` file if the module was installed as a `.zip` file, for example.

gspschmid (approved):

PR #1166, opened 2024-11-22
Bump based CUDA image to ubuntu24.04

DwarKapex:
Ubuntu24.04 uses `python-3.12` as a main interpreter. Unfortunately, not all python packages, we use here, has py-3.12 wheel for amd64/arm64, so need to build the following packages from source:
1. TF-Text
2. Lingvo

Also `python-3.12` added a system-wide protection layer ([PEP 668](https://peps.python.org/pep-0668/)) when install packages using `pip`:
```
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
```
There are at least 2 possible solutions:
1. install everything into `venv` (the initial solution was proposed by @olupton).
2. System-wide installation by forcing pip installation with env flag `PIP_BREAK_SYSTEM_PACKAGES=1`

This branch contains both solutions, but collective mind and experience of PyTorch team suggests to finalize the second solution (system-wide installation)


.github/container/Dockerfile.base
```diff
@@ -127,7 +129,11 @@ git apply </opt/pip/pip-vcs-equivalency.patch
 git add -u
 git commit -m 'Adds JAX_TOOLBOX_VCS_EQUIVALENCY as a trigger to treat all github VCS installs for a package as equivalent. The spec of the last encountered version will be used'
 EOF
-RUN pip install --upgrade --no-cache-dir -e /opt/pip pip-tools && rm -rf ~/.cache/*
+# Create a system-wide venv for the pip-installed world inside the containers
+RUN python -m venv --prompt jax /opt/venv && /opt/venv/bin/pip install --ignore-installed --no-cache-dir -e /opt/pip pip-tools
+# Make sure `python` refers to the venv version
+ENV PATH=/opt/venv/bin:${PATH}
```
yhtang, line 135: Is this sufficient instead of `source /opt/venv/bin/activate`?

.github/container/Dockerfile.jax
```diff
@@ -36,7 +36,7 @@ RUN --mount=type=ssh \
     --mount=type=secret,id=SSH_KNOWN_HOSTS,target=/root/.ssh/known_hosts \
 <<"EOF" bash -ex
     git-clone.sh ${URLREF_JAX} ${SRC_PATH_JAX}
-    sed 's/^numpy.*/numpy<2.0.0/' ${SRC_PATH_JAX}/build/requirements.in
+    sed -i 's/^numpy.*/numpy<2.0.0/' ${SRC_PATH_JAX}/build/requirements.in
```
yhtang, line 39: Can we add a comment above this to document the reason and also prompt future actions to remove it once TF supports NumPy 2?

.github/container/Dockerfile.maxtext
```diff
@@ -2,7 +2,7 @@
 
 ARG BASE_IMAGE=ghcr.io/nvidia/jax-mealkit:jax
 ARG URLREF_MAXTEXT=https://github.com/google/maxtext.git#main
-ARG URLREF_TFTEXT=https://github.com/tensorflow/text.git#v2.13.0
+ARG URLREF_TFTEXT=https://github.com/tensorflow/text.git#master
```
yhtang, line 5: Is this tested? It does not seem to be related to the bump to Ubuntu 24.04.

.github/container/Dockerfile.pax
```diff
@@ -3,7 +3,7 @@
 ARG BASE_IMAGE=ghcr.io/nvidia/jax-mealkit:jax
 ARG URLREF_PAXML=https://github.com/google/paxml.git#main
 ARG URLREF_PRAXIS=https://github.com/google/praxis.git#main
-ARG URLREF_TFTEXT=https://github.com/tensorflow/text.git#v2.13.0
+ARG URLREF_TFTEXT=https://github.com/tensorflow/text.git#master
```
yhtang, line 6: Same question as above.

.github/workflows/_build.yaml
```diff
@@ -31,7 +31,7 @@ on:
         required: true
       DOCKERFILE:
         type: string
-        description: "Dockerfile to use, e.g. .github/container/Dockerfile.t5x.amd64"
```
yhtang, line 34: Why renaming all the amd64 Dockerfiles? Is there a discussion that I missed? :grin:

yhtang (changes requested):

.github/container/Dockerfile.base
DwarKapex: Yes, it. The search mechanism starts checking the PATH one at a time till first match. So, even if there is another python in the PATH, it picks up the first one:
```
$> docker run -it --rm ghcr.io/nvidia/jax-toolbox-internal:11967699553-base-amd64 bash -c "which python && which pip"
/opt/venv/bin/python
/opt/venv/bin/pip
```

.github/container/Dockerfile.jax
DwarKapex: Looks like we don't it any more. In TF requirements for python3.12 there is `numpy==2.1.1`

.github/container/Dockerfile.maxtext
DwarKapex: Unfortunately, related. Google merge changes to support wheel build for python3.12 via this [PR](https://github.com/tensorflow/text/pull/1323). For now we have to use master  because the last release is 2.18.0 published 1 month ago and the fixing PR was merged 2 weeks ago. Waiting 2.18+ release to ping it here.

.github/container/Dockerfile.pax
DwarKapex: Same answer as above =)

.github/workflows/_build.yaml
DwarKapex: The difference between amd64 and arm64 is that we build some wheels (like lingo, tf-text) for arm64, but not for amd64. With this PR we build these packages for both archs, so no need to have 2 different Dockerfiles. I remove arm64 version of Dockerfile and extension `amd64`.

DwarKapex:
@olupton desperately need your input here =)

.github/container/Dockerfile.maxtext
nouiz: Note, this make use track master TF-Text instead of a pinned version.
We could pin a commit as an alternative.

.github/workflows/_build.yaml
yhtang: I would view manual building of lingvo/tftext wheel as a temporary remedy that (unfortunately) complicates the Dockerfile. This workaround shouldn't stay forever. We should push for proper wheel builds by upstream and at the meantime keep the simpler Dockerfiles so that they will become the default again hopefully soon.

.github/container/Dockerfile.base
yhtang: `/opt/venv/bin/activate` performs additional actions in addition to appending to `PATH`. Are we sure it is OK to omit these in our container?

.github/container/Dockerfile.maxtext
yhtang: > Note, this make use track master TF-Text instead of a pinned version. We could pin a commit as an alternative.

Our current 'standard' approach is to tracker the master in the Dockerfile, and then use `manifest.yaml` and the [`EXTRA_BUILD_ARGS`](https://github.com/NVIDIA/JAX-Toolbox/blob/b0e67537bca3955520f5503a0d869178eaf5d6ae/.github/workflows/_ci.yaml#L158) mechanism for CI reproducibility.

yhtang: @DwarKapex could you please add the plumbing for lingvo/tftext version tracking following the example of https://github.com/NVIDIA/JAX-Toolbox/blob/b0e67537bca3955520f5503a0d869178eaf5d6ae/.github/workflows/_ci.yaml#L158 ?

yhtang:
Some builds are still broken :stuck_out_tongue_winking_eye:

.github/container/Dockerfile.base
yhtang: > There was some discussion that using [#1122 (comment)](https://github.com/NVIDIA/JAX-Toolbox/pull/1122#issuecomment-2442330737) instead could be lower risk.

IMHO `PIP_BREAK_SYSTEM_PACKAGES=1` is just fine because we are inside a container.

DwarKapex: The only problem arises is like this:
```
Found existing installation: PyGObject 3.48.2
ERROR: Cannot uninstall 'PyGObject'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.
```
So IMHO, venv is better solution to avoid this kind of issues.

.github/workflows/_build.yaml
DwarKapex: Agree

.github/container/Dockerfile.base
DwarKapex: The "uninstallable" packages come from line 73 if this file (see above)

yhtang (changes requested):
Looks mostly good!

Could you please resurrect the `.amd64` Dockerfiles, even though we are not using them by default at this time? The concern here is that our lingvo/tftext builds add too many boilerplates to the Dockerfile, and are not easily removable should one day we want to move back to use the upstream wheels. We need a reference here to remind us how simple things could be 😁

gpupuck:
I don't understand everything you've changed, but I'm good with the changes made for `Dockerfile.base` and `Dockerfile.jax`.

.github/container/Dockerfile.base
```diff
@@ -129,7 +143,11 @@ git apply </opt/pip/pip-vcs-equivalency.patch
 git add -u
 git commit -m 'Adds JAX_TOOLBOX_VCS_EQUIVALENCY as a trigger to treat all github VCS installs for a package as equivalent. The spec of the last encountered version will be used'
 EOF
-RUN pip install --upgrade --no-cache-dir -e /opt/pip pip-tools && rm -rf ~/.cache/*
+
+# install all python packages system-wide.
+ENV PIP_BREAK_SYSTEM_PACKAGES=1
+RUN pip install --upgrade --ignore-installed --no-cache-dir -e /opt/pip pip-tools && rm -rf ~/.cache/*
```
nouiz, line 149: Can you add a comment to tell why --ignore-installed is needed?

PR #1188, opened 2024-12-06
Run NCCL tests on EKS using MPI operator

olupton:
See https://github.com/NVIDIA/JAX-Toolbox/actions/runs/12200800340/job/34037853757 for a previous sandbox run.

### DIFF
--- a/.github/eks-workflow-files/mpi-nccl-test.yml
+++ b/.github/eks-workflow-files/mpi-nccl-test.yml
@@ -1,0 +1,75 @@
+apiVersion: kubeflow.org/v2beta1
+kind: MPIJob
+metadata:
+  name: PLACEHOLDER
+spec:
+  # Without this then the first few attempts to run the launcher will result in errors
+  # due to failed DNS resolution of the worker names. It works eventually, given a big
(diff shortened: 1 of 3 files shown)


.github/eks-workflow-files/mpi-nccl-test.yml
```diff
@@ -0,0 +1,75 @@
...
+      # to provide useful diagnostics
+      restartPolicy: Never
+      template:
+        spec:
+          containers:
+            - image: PLACEHOLDER
+              imagePullPolicy: IfNotPresent
+              name: PLACEHOLDER
+              command:
+                - mpirun
```
yhtang, line 30: Does YAML permit the `[...]` syntax? This long list of mpirun arguments looks very funny :joy:

yhtang (approved):

.github/eks-workflow-files/mpi-nccl-test.yml
olupton: It's supported, but I don't think there's any middle ground between one very long line in `[...]` format or the current approach that `yq` wouldn't mangle (leading to a hard-to-read diff). I agree it looks a bit odd though.

PR #1206, opened 2024-12-19
No need to build tensorflow-text from source: py3.12 x86/arm64 wheels are available on pypi

DwarKapex:
Recently, Google has released tensorflow-text 2.18.1 that is available on python-3.12 and both archs: x86 and arm64. So, we don't need to build it from source any more

### DIFF
--- a/.github/container/Dockerfile.maxtext
+++ b/.github/container/Dockerfile.maxtext
@@ -8,7 +8,6 @@ ARG SRC_PATH_MAXTEXT=/opt/maxtext
 ## Download source and add auxiliary scripts
 ###############################################################################
 
-ARG BASE_IMAGE
 FROM ${BASE_IMAGE} as mealkit
 ARG URLREF_MAXTEXT
 ARG SRC_PATH_MAXTEXT
--- a/.github/container/Dockerfile.maxtext
+++ b/.github/container/Dockerfile.maxtext
@@ -43,6 +43,3 @@ FROM mealkit as final
 RUN pip-finalize.sh
 
 WORKDIR ${SRC_PATH_MAXTEXT}
-
-# When tftext wheels are published on pypi.org, revert this
-# Dockerfile to 5c4b687b918e6569bca43758c346ad8e67460154
--- a/.github/container/Dockerfile.maxtext
+++ b/.github/container/Dockerfile.maxtext
@@ -2,57 +2,16 @@
 
 ARG BASE_IMAGE=ghcr.io/nvidia/jax-mealkit:jax
 ARG URLREF_MAXTEXT=https://github.com/google/maxtext.git#main
-ARG URLREF_TFTEXT=https://github.com/tensorflow/text.git#master
 ARG SRC_PATH_MAXTEXT=/opt/maxtext
-ARG SRC_PATH_TFTEXT=/opt/tensorflow-text
-
-###############################################################################
-## build tensorflow-text and lingvo, which do not have working arm64 pip wheels
-###############################################################################
-
-ARG BASE_IMAGE
-FROM ${BASE_IMAGE} as wheel-builder
-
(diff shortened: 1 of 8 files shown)


.github/container/Dockerfile.maxtext
```diff
@@ -2,57 +2,16 @@
...
-echo "write_to_bazelrc \"build --linkopt='-Wl,--undefined-version'\"" >> oss_scripts/configure.sh
-
-./oss_scripts/run_build.sh
-EOF
 
 ###############################################################################
 ## Download source and add auxiliary scripts
 ###############################################################################
 
+ARG BASE_IMAGE
```
nouiz, line 11: duplicate of line 3.

.github/container/Dockerfile.maxtext
```diff
@@ -86,5 +45,5 @@ RUN pip-finalize.sh
 
 WORKDIR ${SRC_PATH_MAXTEXT}
 
-# When tftext and lingvo wheels are published on pypi.org, revert this
+# When tftext wheels are published on pypi.org, revert this
 # Dockerfile to 5c4b687b918e6569bca43758c346ad8e67460154
```
nouiz, line 49: Remove both lines and do that revert?

.github/container/Dockerfile.pax
```diff
@@ -128,16 +94,12 @@ ARG URLREF_PAXML
...
 COPY --from=lingvo-builder /opt/manifest.d/git-clone.yaml /opt/manifest.d/git-clone.yaml
 COPY --from=lingvo-builder /tmp/lingvo/dist/lingvo*-linux*.whl /opt/
 RUN echo "lingvo @ file://$(ls /opt/lingvo*.whl)" >> /opt/pip-tools.d/requirements-paxml.in
 
-COPY --from=tftext-builder ${SRC_PATH_TFTEXT}/tensorflow_text*.whl /opt/
-RUN echo "tensorflow-text @ file://$(ls /opt/tensorflow_text*.whl)" >> /opt/pip-tools.d/requirements-paxml.in
-
 # paxml + praxis
 RUN <<"EOF" bash -ex
 echo "tensorflow_datasets==4.9.2" >> /opt/pip-tools.d/requirements-paxml.in
```
nouiz, line 105: Updates the last 2 lines of this file?

.github/container/Dockerfile.maxtext
DwarKapex: I cannot do that now: the commit mentioned here assumes that both `tftext` and `lingvo` are available on python-3.12, but it's not true for `lingvo`, that's why we still need to build `lingvo` for both x86_64 and arm64.

.github/container/Dockerfile.pax
DwarKapex: Same as above.

.github/container/Dockerfile.maxtext
olupton: I don't think lingvo is relevant to maxtext (and we are commenting on the maxtext Dockerfile). I would just delete these two lines.

.github/container/Dockerfile.pax
olupton: You could update the comment to only cite lingvo.

rosetta/Dockerfile.gemma
```diff
@@ -93,7 +55,7 @@ optax
 protobuf
 tfds-nightly
 tensorflow
-tensorflow-text @ file://$(ls /opt/tensorflow_text*.whl)
```
olupton, line 96: Do we still need to explicitly list `tensorflow-text` as a dependency now the wheels for it are published? (not a big deal)

olupton (approved):
Generally LGTM, only nits.

olupton (approved):

PR #1236, opened 2025-01-09
Only set CUDA_DEVICE_MAX_CONNECTIONS=1 for Hopper/cc9.0 runs

olupton:


gpupuck (approved):

nouiz:
Note, we should also merge this in the main branch.

olupton:
> Note, we should also merge this in the main branch.

I https://github.com/NVIDIA/JAX-Toolbox/pull/1249 opened for this - please review.

PR #1237, opened 2025-01-09
 CI: run MaxText tests on AWS with NGC release candidate images

olupton:


gpupuck (dismissed):
Nice!

.github/eks-workflow-files/maxtext-job.yaml
```diff
@@ -0,0 +1,120 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: PLACEHOLDER
```
DwarKapex, line 4: What is `PLACEHOLDER` here? Will be replaced for real job name?

.github/eks-workflow-files/maxtext-job.yaml
```diff
@@ -0,0 +1,120 @@
...
+              export SERVICE_NAME=$0
+              export JOB_NAME=$1
+              cat >each-process.sh <<'EOL'
+              export JAX_COORDINATOR_IP=${JOB_NAME}-0.${SERVICE_NAME}
+              export JAX_COORDINATOR_PORT=3389
+              export NNODES=16 # actually #processes == #GPUs
+              export NODE_RANK=$((JOB_COMPLETION_INDEX*8 + LOCAL_RANK))
+              export JAX_LOCAL_DEVICE_IDS=$LOCAL_RANK
+              export NCCL_DEBUG=INFO
+              export NCCL_DEBUG_FILE=/opt/output/nccl.$NODE_RANK.log
```
DwarKapex, line 48: Does it make sense to add JobID to log file?

.github/eks-workflow-files/maxtext-job.yaml
```diff
@@ -0,0 +1,120 @@
...
+                --steps=20 \
+                --fsdp=16 \
+                -a "scan_layers=false \
+                    max_target_length=4096 \
+                    use_iota_embed=true \
+                    logits_dot_in_fp32=false \
+                    profiler=nsys \
+                    skip_first_n_steps_for_profiler=3 \
+                    profiler_steps=8" \
+                |& tee /opt/output/output.$NODE_RANK.log >"${console}"
```
DwarKapex, line 70: `/opt/..` -> $NCCL_DEBBUG_FILE?
Does it mean, that NCCL log and test-maxtext logs will be added into single file?

.github/eks-workflow-files/maxtext-job.yaml
```diff
@@ -0,0 +1,120 @@
...
+              name: shmem
+            - mountPath: /opt/output
+              name: output
+        - name: upload
+          image: amazon/aws-cli
+          command:
+            - bash
+            - -c
+            - |
+              JOB_NAME="$0"
```
DwarKapex, line 101: Where does `JOB_NAME` come from?

.github/eks-workflow-files/maxtext-job.yaml
```diff
@@ -0,0 +1,120 @@
...
+              while [[ ! -f /opt/output/.done ]]; do
+                sleep 1
```
DwarKapex, line 103: 1 sec? Maybe, 5-10-20 secs?

.github/eks-workflow-files/maxtext-job.yaml
olupton: No, there are two files `output.N.log` and `nccl.N.log`
The 0th process on each node/pod (i.e. ranks 0 and 8 when running 16 processes on 2 nodes) prints to stdout too, which is what streams to GitHub Actions logs, while the other 7 processes per node/pod are only saved to disk

.github/eks-workflow-files/maxtext-job.yaml
olupton: The `PLACEHOLDER` just below. This bash command is `bash -c "JOB_NAME=$0 ..." PLACEHOLDER`. This is filled in with `yq` like the other `PLACEHOLDER`s.

.github/eks-workflow-files/maxtext-job.yaml
olupton: 1 second, yes. This is just polling for existence of a local file, so I think 1 second is fine.

.github/eks-workflow-files/maxtext-job.yaml
olupton: That's part of the JOB_NAME in the destination directory of `aws s3 cp` below.

.github/eks-workflow-files/maxtext-job.yaml
```diff
@@ -0,0 +1,120 @@
...
+      containers:
+        - name: maxtext
+          image: PLACEHOLDER
+          ports:
+            - containerPort: 3389
+          command:
+            - bash
+            - -c
+            # The logging logic: stream stdout/stderr from the 0th process inside this pod,
+            # record all of the processes' stdout/stderr + the INFO-level NCCL logs to file
```
Steboss, line 37: Out of curiosity, do we store these logs somewhere? Or do we have a way to inspect them?

.github/workflows/_test_maxtext_k8s.yaml
```diff
@@ -0,0 +1,107 @@
...
+          git diff .github/eks-workflow-files/maxtext-job.yaml
+      - name: Submit Kubernetes job
+        run: kubectl apply -f .github/eks-workflow-files/maxtext-job.yaml
+      - name: Wait for Kubernetes job to start
+        run: |
+          # Launcher job is created eagerly, but suspended. Kueue un-suspends it when
+          # resources are available, but that is where there can be a long wait if the
+          # cluster is busy executing other jobs.
+          kubectl wait --for=create job/${JOB_NAME}
+          kubectl wait --for=jsonpath='{.spec.suspend}=false' job/${JOB_NAME} --timeout=3600s
```
Steboss, line 66: Did we have a test for these two commands? Just to be sure they work fine, and to if timeout=3600s is long enough?

.github/workflows/_test_maxtext_k8s.yaml
```diff
@@ -0,0 +1,107 @@
...
+      - name: Stream Kubernetes job output
+        run: |
+          # Streaming logs will fail if the container/pod is still pending
+          while [[ -n $(kubectl get pods --selector=batch.kubernetes.io/job-name=${JOB_NAME} --output=jsonpath='{.items[?(@.status.phase == "Pending")].metadata.name}') ]]; do
+            sleep 1
```
Steboss, line 71: Is 1 second too short or test wise is enough?

Steboss (approved):
Minor comments, but great work!

.github/eks-workflow-files/maxtext-job.yaml
olupton: Currently they are uploaded to an S3 bucket, where they live for 30 days. This happens in the `upload` container below. So: we could manually recover if needed for debugging, but the relatively-user-friendly path is to just look at the log output streamed to GitHub Actions.

.github/workflows/_test_maxtext_k8s.yaml
olupton: There's not currently any targeted testing - obviously this will get run regularly in CI, so it is implicitly integration-tested I suppose.

.github/workflows/_test_maxtext_k8s.yaml
olupton: I think it's fine as a starting point - we can tune later if we see problems.

PR #1238, opened 2025-01-09
Remove V100 from test environment

DwarKapex:
We drop support of VOLTA and don't need run tests for V100 anymore.


.github/workflows/_test_unit.yaml
```diff
@@ -34,23 +34,14 @@ jobs:
...
-      fail-fast: false
-      matrix:
-        GPU_ARCH: [V100, A100]
-        include:
-          - EXTRA_LABEL: "self-hosted"
-          # ensures A100 job lands on dedicated runner for this particular job
-          - GPU_ARCH: A100
-            EXTRA_LABEL: "${{ github.run_id }}"
-    name: ${{ inputs.TEST_NAME }}-${{ matrix.GPU_ARCH }}-unit-test
+    name: ${{ inputs.TEST_NAME }}-A100-unit-test
```
yhtang, line 37: Why not shorten the matrix list to `[A100]`? This way, it is easier to extend this list later to include newer GPUs such as H100/B100.

yhtang (changes requested):

yhtang (approved):

PR #1241, opened 2025-01-10
[nsys-jax] Add ratio of hidden communication time to total communication time

sfvaroglu:


.github/container/nsys_jax/nsys_jax/analyses/communication.py
```diff
@@ -95,6 +96,33 @@ def format_bandwidth(data, collective):
...
+    for collective, df in steady_state.communication.groupby(
+        ["Collective"]
+    ):
+        collective_types.add(collective)
+        summary_data[collective] = df["DurHiddenMsToDurMs"].mean()
+
+    collective_width = max(len("Collective"), max(len(f"{collective}") for collective in collective_types))
+    ratio_width = len("Mean HiddenToTotalMs")
+
+    print()
```
DwarKapex, line 110: I think python `prettyTable` (https://pypi.org/project/prettytable/) should work here and you don't need to do all these calculations manually

sfvaroglu: Done. Thanks!

.github/container/nsys_jax/nsys_jax/analysis.py
```diff
@@ -331,6 +331,9 @@ def calculate_collective_metrics(
     comm_df["BusBandwidthGBPerSec"] = (
         comm_df["AlgorithmBandwidthGBPerSec"] * comm_df["BusBandwidthCorrection"]
     )
+    comm_df["DurHiddenMsToDurMs"] = comm_df["ProjDurHiddenMs"] / (
+        comm_df["ProjDurMs"] + comm_df["ProjDurHiddenMs"]
+    )
```
olupton, line 336: I would be inclined to calculate this just before use, rather than in the data loader.

.github/container/nsys_jax/nsys_jax/analyses/communication.py
```diff
@@ -95,6 +96,29 @@ def format_bandwidth(data, collective):
...
+        collective_types.add(collective)
+        summary_data[collective] = df["DurHiddenMsToDurMs"].mean()
+
+    table = PrettyTable()
+    table.field_names = ["Collective", "Mean HiddenToTotalMs"]
+
+    for collective in collective_types:
+        mean_value = summary_data[collective]
+        table.add_row([collective[0], mean_value])
+    print(table)
```
olupton, line 111: The result is visible in the CI here: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/12753978814/job/35547441911#step:13:301
But admittedly that test is not expected to have overlap, so it's not very interesting.

It's also visible in the CI here: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/12753978740/job/35546782084#step:5:104, which is a bit more interesting because the test data (which are committed to the repo for that test, not generated on the fly) do contain some overlap.

olupton (approved):
LGTM for a first iteration.

In terms of plumbing these new metrics into a monitoring system, wouldn't it be better to also write them to a machine-readable file? Otherwise you would have to e.g. grep numbers back out of the `prettytable`.

I don't think there is currently test coverage for this, but the intent is that files written by the analysis recipes are saved to the output `.zip`.

olupton (approved):
Thanks!

PR #1248, opened 2025-01-14
Update the dockerfile base image to cuda-dl-base

Steboss:
Update the base docker image, so we can use cuda-dl-base from `nvcri`.


.github/container/Dockerfile.base
olupton, line 1: GitHub won't let me comment on individual lines lower down this file.

I think that, as well as the change you've made, we want to drop a significant number of the later lines in this `Dockerfile`, because they relate to components that are already shipped in `cuda-dl-base`.

e.g.
- I think the `tcpx` plugin code is outdated and unused (@yhtang, @chaserileyroberts) - similarly the autoconfig script for GCP
- `install-nsight.sh`, `install-cudnn.sh`, `install-nccl.sh`: all of these install CUDA-ish components that are distributed in the `cuda-dl-base` container
- `install-ofed.sh`, `install-efa.sh`: these install bits of the networking stack that come in the base container, running them on top of `cuda-dl-base` will actively break things
- `check-shm.sh`: the base container implements this

@DwarKapex can advise on other Dockerfile changes that are needed when changing the base.

olupton (changes requested):

.github/container/Dockerfile.base
Steboss: I'll start introducing these changes :)

Steboss:
So I run a check in the `cuda-dl-base` image, and I can see that: 
- for `nccl` we need to create symlink for `include` and `lib` directories, so they're mapped in `opt/nvidia/nccl`
- same for `cudNN` 
- we can safely remove `install-ofed.sh`
- and amazon efa support 

For the symlink, we'd just need this part of the `install-nccl.sh` script (and the counterpart in `install-cudnn.sh` script): 
```
arch=$(uname -m)-linux-gnu
for nccl_file in $(dpkg -L libnccl2 libnccl-dev | sort -u); do
  # Real files and symlinks are linked into $prefix
  if [[ -f "${nccl_file}" || -h "${nccl_file}" ]]; then
    # Replace /usr with $prefix and remove arch-specific lib directories
    nosysprefix="${nccl_file#"/usr/"}"
    noarchlib="${nosysprefix/#"lib/${arch}"/lib}"
    link_name="${prefix}/${noarchlib}"
    link_dir=$(dirname "${link_name}")
    mkdir -p "${link_dir}"
    ln -s "${nccl_file}" "${link_name}"
  else
    echo "Skipping ${nccl_file}"
  fi
done
```

@DwarKapex does it sound right to you?

Steboss:
@DwarKapex 
- Updated the based Dockerfile to have `cuda-dl-base` image
- to avoid having conflicts and re-install of `nccl` and `cudnn` i've modified `install-cudnn.sh` and `install-nccl.sh` 
- for both script I added a part where we're doing the symlink step to that resources are present in `/opt/nvidia/{package}`

.github/container/Dockerfile.base
```diff
@@ -153,72 +136,20 @@ ENV PIP_BREAK_SYSTEM_PACKAGES=1
...
-ENV LD_LIBRARY_PATH=/opt/amazon/efa/lib:${LD_LIBRARY_PATH}
-ENV PATH=/opt/amazon/efa/bin:${PATH}
-
-##############################################################################
-## NCCL sanity check utility
-##############################################################################
-
-ADD install-nccl-sanity-check.sh /usr/local/bin
-ADD nccl-sanity-check.cu /opt
-RUN install-nccl-sanity-check.sh
```
olupton, line 207: @gspschmid I assume you would like to keep this?

.github/container/Dockerfile.base
```diff
@@ -153,72 +136,20 @@ ENV PIP_BREAK_SYSTEM_PACKAGES=1
...
-ADD jax-nccl-test parallel-launch /usr/local/bin/
```
olupton, line 208: Please don't delete this line - these are relevant to `nsys-jax` and testing thereof.

.github/container/install-cudnn.sh
```diff
@@ -70,3 +37,26 @@ for cudnn_file in $(dpkg -L ${libcudnn_name} ${libcudnn_dev_name} | sort -u); do
...
+    # Check if the symlink already exists
+    if [[ -e "${link_name}" ]]; then
+      echo "Symlink ${link_name} already exists. Skipping."
+    else
+      ln -s "${adjusted_target}" "${link_name}"
+    fi
+  else
+    echo "Skipping symlink ${symlink} with target ${symlink_target}"
+  fi
+done
```
olupton, line 62: Would the link generation code above this be sufficient if `CUDNN_MAJOR_VERSION` was still set?

.github/container/install-nccl.sh
```diff
@@ -42,7 +11,15 @@ if [[ -d "${prefix}" ]]; then
   exit 1
 fi
 arch=$(uname -m)-linux-gnu
-for nccl_file in $(dpkg -L libnccl2 libnccl-dev | sort -u); do
```
olupton, line 45: Did this break?

Steboss: Actually it's the same. 
In this implementation we have:
```
nccl_packages=$(dpkg -l 'libnccl*' | awk '/^ii/ {print $2}')
```
Which is the same having in the for loop `libnccl2 libnccl-dev`

olupton:
The `nsys-jax` test failures are because the 24.12 cuda-dl-base includes Nsight Systems 2024.7, whereas we currently install 2024.6 because of some issues with 2024.7 (https://github.com/NVIDIA/JAX-Toolbox/pull/1176 is the - pending - attempt to move to 2024.7). A possible workaround would be to use the 24.11 cuda-dl-base for the moment.

.github/container/Dockerfile.base
```diff
@@ -153,50 +136,20 @@ ENV PIP_BREAK_SYSTEM_PACKAGES=1
...
-###############################################################################
-## Install the latest versions of Nsight Systems and Nsight Compute
-###############################################################################
-
-ADD install-nsight.sh /usr/local/bin
-RUN install-nsight.sh
-
-###############################################################################
-## Install cuDNN
+## Symlink for cuDNN
```
DwarKapex, line 139: **Do we want to be able to customize cuDNN and NCCL?**

In the past we had several issues with NCCL: the base-docker-image NCCL had a bug, that was fixed in the next NCCL version and developers had to upgrade it manually. Maybe let's keep this option here?
Introduce 2 docker vars like NCCL_VERSION, CUDNN_VERSION and leave them empty to signal the script, that we want to continue with default versions, or in case it's not empty, run update in the `install-<nccl/cudnn>.sh` scripts? Also add these vars as GitHub vars, so that we can test them thru webUI.

WDYT? Is this feasable as upcoming PR?

.github/container/Dockerfile.base
```diff
@@ -207,26 +160,14 @@ ADD nccl-sanity-check.cu /opt
 RUN install-nccl-sanity-check.sh
 ADD jax-nccl-test parallel-launch /usr/local/bin/
 
-###############################################################################
-## Add the systemcheck to the entrypoint.
-###############################################################################
-
-COPY check-shm.sh /opt/nvidia/entrypoint.d/
```
DwarKapex, line 214: It is an important message about using `shm` in docker runtime. I would leave it here.

.github/container/Dockerfile.base
```diff
@@ -207,26 +160,14 @@ ADD nccl-sanity-check.cu /opt
...
-
-###############################################################################
-## Add the GCP - TCPX check to the entrypoint.
-###############################################################################
-
-# TODO(chaserileyroberts): Reenable once fully tested on GCP.
-# COPY gcp-autoconfig.sh /opt/nvidia/entrypoint.d/
```
DwarKapex, line 221: Not sure about this one. Better ask Yu-Hang/Chase whether we need this or not.

.github/container/Dockerfile.base
```diff
@@ -207,26 +160,14 @@ ADD nccl-sanity-check.cu /opt
...
-
 ###############################################################################
 ## Install the nsys-jax JAX/XLA-aware profiling scripts, patch Nsight Systems
 ###############################################################################
 
 ADD install-nsys-jax.sh /usr/local/bin
 RUN install-nsys-jax.sh ${JAX_TOOLBOX_REF}
 
+
```
DwarKapex, line 170: Why? =)))

Steboss: Whopps, I must have entry a new line by mistake

.github/container/Dockerfile.base
Steboss: Re-added

.github/container/Dockerfile.base
Steboss: This is a good shout. 
I agree, also based on what Olli said before. I will introduce these changes in a second PR, after merging this

.github/container/Dockerfile.base
```diff
@@ -153,50 +136,20 @@ ENV PIP_BREAK_SYSTEM_PACKAGES=1
...
 
 ADD install-cudnn.sh /usr/local/bin
 RUN install-cudnn.sh
 
 ###############################################################################
-## Install NCCL
+## Symlink for NCCL
 ###############################################################################
 
+# same fro this
```
nouiz, line 149: fro -> for?

nouiz (approved):

.github/container/Dockerfile.base
olupton: I suggested removing this because the new base container already implements an equivalent message:
```
$ docker run -it nvcr.io/nvidia/cuda-dl-base:24.12-cuda12.6-devel-ubuntu24.04

==========
== CUDA ==
==========

NVIDIA Release  (build )
CUDA Version 12.6.3.004
Container image Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.

Various files include modifications (c) NVIDIA CORPORATION & AFFILIATES.  All rights reserved.

This container image and its contents are governed by the NVIDIA Deep Learning Container License.
By pulling and using the container, you accept the terms and conditions of this license:
https://developer.nvidia.com/ngc/nvidia-deep-learning-container-license

WARNING: The NVIDIA Driver was not detected.  GPU functionality will not be available.
   Use the NVIDIA Container Toolkit to start this container with GPU support; see
   https://docs.nvidia.com/datacenter/cloud-native/ .

NOTE: The SHMEM allocation limit is set to the default of 64MB.  This may be
   insufficient for CUDA.  NVIDIA recommends the use of the following flags:
   docker run --gpus all --ipc=host --ulimit memlock=-1 --ulimit stack=67108864 ...
   ```

File: .github/container/Dockerfile.base (hunk 7 above)
Comment (olupton) [reply .github/container/Dockerfile.base:221]: It is related to TCPX, but it got disabled almost immediately after it was added.
@yhtang to comment on the next steps given his recent work.

File: .github/container/Dockerfile.base [hunk 10]
```diff
@@ -153,50 +136,20 @@ ENV PIP_BREAK_SYSTEM_PACKAGES=1
...
-###############################################################################
-
-ENV TCPX_LIBRARY_PATH=/usr/local/tcpx/lib64
-COPY --from=tcpx-installer /var/lib/tcpx/lib64 ${TCPX_LIBRARY_PATH}
-
-###############################################################################
-## Install the latest versions of Nsight Systems and Nsight Compute
-###############################################################################
-
-ADD install-nsight.sh /usr/local/bin
```
olupton, line 166: I think you can also delete the script from the repository (including wherever else this applies)

.github/container/Dockerfile.base
olupton, line 149:
```suggestion
```

.github/container/Dockerfile.base
```diff
@@ -153,50 +136,20 @@ ENV PIP_BREAK_SYSTEM_PACKAGES=1
...
+# same fro this 
 ADD install-nccl.sh /usr/local/bin
 RUN install-nccl.sh
 
-###############################################################################
-## RoCE and InfiniteBand support
-###############################################################################
-
-ADD install-ofed.sh /usr/local/bin
-RUN install-ofed.sh
```
olupton, line 188: Delete this from the repository too?

.github/container/Dockerfile.base
```diff
@@ -153,50 +136,20 @@ ENV PIP_BREAK_SYSTEM_PACKAGES=1
...
-
-##############################################################################
-## Amazon EFA support (need to run it inside container separately)
-##############################################################################
-
-ADD --chmod=777 \
-  install-efa.sh \
```
olupton, line 195: Delete this?

.github/container/Dockerfile.base
```diff
@@ -153,50 +136,20 @@ ENV PIP_BREAK_SYSTEM_PACKAGES=1
...
-  test-aws-efa.sh \
```
olupton, line 196: Leave this one for the moment (just `ADD`, no `RUN`)

.github/container/install-cudnn.sh
```diff
@@ -70,3 +39,26 @@ for cudnn_file in $(dpkg -L ${libcudnn_name} ${libcudnn_dev_name} | sort -u); do
...
+#     # Check if the symlink already exists
+#     if [[ -e "${link_name}" ]]; then
+#       echo "Symlink ${link_name} already exists. Skipping."
+#     else
+#       ln -s "${adjusted_target}" "${link_name}"
+#     fi
+#   else
+#     echo "Skipping symlink ${symlink} with target ${symlink_target}"
+#   fi
+# done
```
olupton, line 64:
```suggestion
```

.github/container/Dockerfile.base
Steboss: I'll rename those, as we're using these scripts to link correctly all the libraries from `/opt/nvidia/*`

.github/container/Dockerfile.base
Steboss: Why do we want to keep those files in the container?

olupton: I was suggesting dropping the `install-efa.sh` (because EFA is installed in the base container) but keeping the test script (because I'm not immediately sure if it could still be useful).

Having looked at it now: https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/container/test-aws-efa.sh I think we could just drop it too. It's not testing functionality, just that the installation exists, but those components are unconditionally installed in the base container.

.github/container/Dockerfile.base
Steboss: It applies, as we need nsight to be linked in `/opt/nvidia/` when running the build of jax

Steboss: Sorry Olli, you're right
I swapped two names in my mind XD dropping this

.github/container/.DS_Store
olupton, line 1: Remove this one!

Steboss: damn, I didn't select the files properly, sorry

.github/container/Dockerfile.base
```diff
@@ -207,18 +158,6 @@ ADD nccl-sanity-check.cu /opt
 RUN install-nccl-sanity-check.sh
 ADD jax-nccl-test parallel-launch /usr/local/bin/
 
-###############################################################################
-## Add the systemcheck to the entrypoint.
-###############################################################################
-
-COPY check-shm.sh /opt/nvidia/entrypoint.d/
```
olupton, line 214: `check-shm.sh` could also be removed

olupton (dismissed):
LGTM, only one minor nit left.
@yhtang and/or @chaserileyroberts to review the GCP networking relevant parts.

olupton (dismissed):

.github/container/Dockerfile.base
yhtang: It only applies to the A3 nodes and a different configuration is needed for A3 mega. Given that Google's interconnect technology is still quickly evolving, I now think we should refrain from shipping this type of autoconfig scripts until the API stabilizes. For the time being, we can add a docs page and link to GCP's own configuration guide.

yhtang (changes requested):
What are the symlink-xyz scripts modeled after? How do other DLFW containers accomodate the dl core container?

Steboss:
@yhtang 
The symlink-xyz scripts are meant to create a symlink for `nvcc` and `cudnn`. These packages are already install in `cuda-dl-base` image, but they're not linked to `/opt/nvidia/` folder, as we were doing before.
This is all a jax/xla thing, that's why we might not need this in the other DLFW

.github/container/Dockerfile.base
```diff
@@ -1,27 +1,10 @@
 # syntax=docker/dockerfile:1-labs
-ARG BASE_IMAGE=nvidia/cuda:12.6.3-devel-ubuntu24.04
+ARG BASE_IMAGE=nvcr.io/nvidia/cuda-dl-base:24.11-cuda12.6-devel-ubuntu24.04
```
gpupuck, line 2: The latest is `24.12-cuda12.6-devel-ubuntu24.04`.

Steboss: thanks for the comment, I'll update the image now :)

Steboss:
@olupton @yhtang 
i think it may be wise to add an additional step in the CI, that check for the very latest `cuda-dl-base` image, so we can avoid updating this manually, and we'll have an automatic system that does it for us.

.github/container/Dockerfile.base
olupton: This was intentional, see above: https://github.com/NVIDIA/JAX-Toolbox/pull/1248#issuecomment-2594896851

yhtang:
> @olupton @yhtang 
> i think it may be wise to add an additional step in the CI, that check for the very latest `cuda-dl-base` image, so we can avoid updating this manually, and we'll have an automatic system that does it for us.

Bumping the CUDA base image is usually not a light job, as it may break things. Hence why we always update it via a PR. The DL base image is only updated once a month so IMHO we can live with it.

yhtang (dismissed):

olupton (approved):

PR #1278, opened 2025-01-30
Simplify the way we're generating badges

Steboss:
This PR contains a way to have a standard way to generate badges from tests


Steboss:
TODO: 
- pass whenever possible the number of passed/failed/error tests as variable 
- compare artefacts outputs with historical data

PR #1284, opened 2025-02-03
Run NCCL tests on the JAX-specific base container

olupton:
The prior setup pre-dated https://github.com/NVIDIA/JAX-Toolbox/pull/1248, now things can be simpler.


olupton:
Note that the green CI is because the main pipeline is not running, because it exceeds the reusable workflow limit. @Steboss agreed to take a look.

.github/actions/build-container/action.yml
```diff
@@ -0,0 +1,212 @@
+name: Build container
```
Steboss, line 1: This action can substitute `_build.yaml` reusable workflow

.github/workflows/_ci.yaml
```diff
@@ -52,109 +53,144 @@ jobs:
...
+          CONTAINER_NAME: ${{ matrix.container_name }}
+          DOCKERFILE: ${{ matrix.dockerfile }}
+          RUNNER_SIZE: ${{ matrix.runner_size }}
+          ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
+          ssh-known-hosts: ${{ vars.SSH_KNOWN_HOSTS }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          bazel-remote-cache-url: ${{ vars.BAZEL_REMOTE_CACHE_URL }}
+          EXTRA_BUILD_ARGS: ${{ matrix.extra_build_args }}
+      # make sure each container has the right tag
+      - name: Export container tags
```
Steboss: We need to export tags, so these can be readable in `outputs`. If we're not exporting these data here, `outputs` will be overwritten everytime a container gets built

.github/actions/build-container/action.yml
olupton: Can we delete `_build.yaml` then?

.github/workflows/_ci.yaml
```diff
@@ -52,109 +53,144 @@ jobs:
...
-      ARCHITECTURE: ${{ inputs.ARCHITECTURE }}
-      ARTIFACT_NAME: artifact-equinox-build
-      BADGE_FILENAME: badge-equinox-build
-      BUILD_DATE: ${{ inputs.BUILD_DATE }}
-      BASE_IMAGE: ${{ needs.build-jax.outputs.DOCKER_TAG_MEALKIT }}
-      CONTAINER_NAME: equinox
-      DOCKERFILE: .github/container/Dockerfile.equinox
-      EXTRA_BUILD_ARGS: |
-        URLREF_EQUINOX=${{ fromJson(inputs.SOURCE_URLREFS).EQUINOX }}
+      CONTAINER: ${{ needs.build-base.outputs.DOCKER_TAG }}
```
olupton, line 61: This doesn't need to sit on top of `build-base`, it should work with `${{ inputs.CUDA_IMAGE }}`.

.github/workflows/_ci.yaml
```diff
@@ -52,109 +53,144 @@ jobs:
...
-      BUILD_DATE: ${{ inputs.BUILD_DATE }}
-      BASE_IMAGE: ${{ needs.build-jax.outputs.DOCKER_TAG_MEALKIT }}
-      CONTAINER_NAME: upstream-t5x
-      DOCKERFILE: .github/container/Dockerfile.t5x
-      EXTRA_BUILD_ARGS: |
-        URLREF_T5X=${{ fromJson(inputs.SOURCE_URLREFS).T5X }}
-        URLREF_AIRIO=${{ fromJson(inputs.SOURCE_URLREFS).AIRIO }}
-    secrets: inherit
+    strategy:
+        matrix:
```
olupton: I wonder if this part is worth it, IIUC it forces all tests to wait for all builds to complete, and there seems to be extra boilerplate too w.r.t just writing out the different `build-XXX` jobs?

Steboss: I agree :)

.github/workflows/_test_nccl.yaml
```diff
@@ -17,26 +14,13 @@ permissions:
...
-      ARTIFACT_NAME: artifact-mpi-operator-compatible-base-build
-      BADGE_FILENAME: badge-mpi-operator-compatible-base-build
-      BUILD_DATE: 0000-00-00 # not important; this image is never published
-      BASE_IMAGE: ${{ inputs.CONTAINER }}
-      CONTAINER_NAME: mpi-operator-compatible-base
-      DOCKERFILE: .github/container/Dockerfile.mpi-operator-compatible-base
-      RUNNER_SIZE: small
-    secrets: inherit
   nccl-test:
-    needs: build-mpi-operator-compatible-base
```
olupton: I think this is still needed.

Steboss: We can move this to the main `_ci.yaml`

olupton: That would stop manually launching the NCCL workflow from working.

Steboss: ah I see, so let's keep `_build` for nccl

olupton: Why can't you use the same action here?

Steboss: Ok gotcha!

Steboss:
@olupton ready for review :)

.github/workflows/_test_nccl.yaml
```diff
@@ -51,7 +62,7 @@ jobs:
         id: var
         shell: bash
         run: |
-          JOB_NAME="nccl-test-${{ github.run_id }}-${{ github.run_attempt }}-${TEST_NAME//_/-}"
+          JOB_NAME="nccl-tets-${{ github.run_id }}-${{ github.run_attempt }}-${TEST_NAME//_/-}"
```
olupton: 🧐

olupton (commented):
Thanks

.github/workflows/_test_nccl.yaml
Steboss: well spotted

Steboss (approved):

PR #1302, opened 2025-02-14
Ray tutorial

keshavb96:
This document presents a detailed tutorial on how Ray can be used together with JAX to achieve fault tolerant training.


docs/resiliency/example/run_example.md
```diff
@@ -0,0 +1,25 @@
...
+```shell
+nvidia-docker build -t ray_resiliency_example -f Dockerfile .
+```
+
+## Running the example
+
+ After sshing into a node with at least 2 GPUs run the following command to get into the previously built container.
+
+```shell
+nvidia-docker run --name resilient_jax --network=host --security-opt seccomp=unconfined --cap-add SYS_PTRACE -it --shm-size=50g --ulimit memlock=-1 resilient_jax
```
gspschmid, line 18: Doesn't make the image name you assigned above, so need to replace `resilient_jax` by `ray_resiliency_example`.

Also, maybe launch `./launch_ray_job.sh` directly? As in `docker run ... ray_resiliency_example ./launch_ray_job.sh`

docs/resiliency/example/run_example.md
```diff
@@ -0,0 +1,25 @@
+## Illustrative example
+
+The example is a single node toy example that demonstrates how to run a simple training loop with a toy model, checkpoint along the way and recover upon encountering simulated failures and hangs. It is meant to be easy to run without needing access to a large compute cluster. 
+
+## Building the image
+
+Start by building the docker image using the Dockerfile in this directory.
+
+```shell
+nvidia-docker build -t ray_resiliency_example -f Dockerfile .
```
gspschmid, line 10: What's the difference between `nvidia-docker run ...` and `docker run --gpus=all ...`?

gspschmid (commented):
Tried this out and needed to make a few changes to get the container ready (see inline comments).

~Once that was set up I ran into the following error on a viking node:~
```console
$ docker build -t ray_resiliency_example -f Dockerfile .
...
$ docker run --gpus=all --name resilient_jax --network=host --security-opt seccomp=unconfined --cap-add SYS_PTRACE -it --shm-size=50g --ulimit memlock=-1 ray_resiliency_example
root@viking-prod-283:/ray_resiliency_example# ./launch_ray_job.sh
...
redis.exceptions.ConnectionError: Error 111 connecting to 10.78.2.240:6380. Connection refused.
```
~Full log here: https://gist.github.com/gspschmid/ff1d8e7873a5010d880cc8350bf314f1~

Nvm, `launch_ray_job.sh` had the line to launch redis commented out, it seems to work after uncommenting that!

docs/resiliency/example/launch_ray_job.sh
```diff
@@ -0,0 +1,33 @@
...
+IP=$(hostname -I | awk '{print $1}')
+PORT=6379
+REDIS_PORT=6380
+IP_HEAD=$IP:$PORT
+
+export NGPUS=$GPUS_PER_NODE # Since this is only for one node
+export REDIS_ADDR=$IP:$REDIS_PORT
+
+# Start the redis server
+# redis-server --bind "$IP" --port "$REDIS_PORT" --protected-mode no --daemonize yes
```
gspschmid, line 15: Need to uncomment this line to start redis.

docs/resiliency/example/ray_example.py
```diff
@@ -0,0 +1,319 @@
...
+        return
+    
+if __name__ == '__main__':
+    ray.init(address='auto', logging_level=0)
+    hang_time_threshold = 120
+    coordinator = RayClusterCoordinator(ModelTrainer, hang_time_threshold)
+    coordinator.initialize_workers()
+    coordinator.log("Initialized workers")
+    asyncio.run(coordinator.run(input_shape=(16, 512), num_steps=30))
+
```
gspschmid, line 313: Remove blank lines?

docs/resiliency/example/ray_example_driver.py
```diff
@@ -0,0 +1,32 @@
...
+        if status in {JobStatus.SUCCEEDED, JobStatus.STOPPED, JobStatus.FAILED}:
+            if status in {JobStatus.STOPPED, JobStatus.FAILED}:
+                logs = client.get_job_logs(job_id)
+                print(logs, flush=True)
+            break
+        time.sleep(5)
+        if status == JobStatus.RUNNING:
+            logs = client.get_job_logs(job_id)
+            print(logs[len(prev_logs):], flush=True)
+            prev_logs = logs
```
gspschmid, line 29: Probably don't want to keep an unbounded amount of logs around.
```python
print(logs[prev_logs_len:], flush=True)
prev_logs_len = len(logs)
```

Though I assume there's a more canonical solution to printing logs from all Ray workers to stdout than repeatedly calling `get_job_logs`?

docs/resiliency/example/launch_ray_job.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Start the ray worker node on the same physical node as the head node (only 1 since this is for a single physical node ray cluster)
+# Typically we want to have the head node and the worker nodes be on separate physical nodes
+# This is for simplicity
+ray start --address "$IP_HEAD" --include-log-monitor=False --resources="{\"worker_units\": $GPUS_PER_NODE}" --min-worker-port=10002 --max-worker-port=11000 --block &
+# Sleep for a second to let the node connect
+sleep 5
+
+python3 ray_example_driver.py
+
```
gspschmid, line 30: Remove blank lines?

docs/resiliency/example/ray_example.py
```diff
@@ -0,0 +1,319 @@
...
+                loss, params, optimizer_state = train_step(model, params, optimizer, optimizer_state, sharded_input)
+
+                # Wait for step to finish before checkpointing then save checkpoint
+                params, optimizer_state = jax.block_until_ready((params, optimizer_state))
+                ModelTrainer.checkpoint(ckpt_mgr, params, optimizer_state, step)
+
+            # Fail every 10 seconds
+            current_time = datetime.datetime.now()
+            time_since_failure_sim = (current_time - failure_timer_start).total_seconds()
+            if time_since_failure_sim >= 10:
```
gspschmid, line 293: Perhaps we should make only one worker fail and not all of them every 10 seconds?

docs/resiliency/example/ray_example.py
```diff
@@ -0,0 +1,319 @@
...
+
+        train_step = jax.jit(ModelTrainer.train_step, static_argnames=['model','optimizer'])
+        input_sharding = jax.sharding.NamedSharding(mesh, jax.sharding.PartitionSpec('data_parallel', None))
+        starting_step = (ckpt_mgr.latest_step() + 1) if ckpt_mgr.latest_step() is not None else 0
+
+        # Train loop
+        failure_timer_start = datetime.datetime.now()
+        for step in range(starting_step, num_steps):
+            time.sleep(1)
+            with self.EnableHeartbeat():
```
gspschmid, line 281: Why make this a context manager? Seems like we're just running a heartbeat at the end of every step?

docs/resiliency/example/Dockerfile
```diff
@@ -0,0 +1,7 @@
+# Base maxtext image
+FROM ghcr.io/nvidia/jax:maxtext
+
+RUN apt-get update && \
+    apt-get install -y redis && \
+    python3 -m pip install redis && \
+    python3 -m pip install "ray[default]"
```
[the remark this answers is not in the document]
keshavb96: Ah yes, I was mounting them into the container. I'll update the Dockerfile :)

docs/resiliency/example/ray_example_driver.py
keshavb96: So Ray dumps logs to /tmp/ray/session_latest where each actor writes to it's own log file, but it's a bit of a nightmare to navigate. This was just my placeholder to view logs as the program runs. It's entirely optional.

gspschmid: It sounds like by default Ray already redirects all the logs of workers to the driver -- do you understand why those are not printed in your case? Is it because you're launching through the job API? Would be good to clean this up and make it clear in the tutorial what the user is supposed to look at.

docs/ray_resilient_jax.md
```diff
@@ -0,0 +1,658 @@
...
+```
+Now the Ray cluster consists of 1 Ray head node and 2 Ray worker nodes, all allocated across the 2 physical nodes with IP addresses `IP_ADDR_1` and `IP_ADDR_2`.
+
+### Starting a Ray Cluster with a SLURM allocation
+
+While the example above is simple and illustrates how to quickly bringup a Ray cluster on two physical nodes, it is inefficient to manually run the scripts when scaling to many more nodes. Additionally, it is likely that physical resources are allocated to an application by a scheduler like SLURM or Kubernetes. In this guide we will assume that SLURM is the physical resource allocator. Below is a template of a SLURM batch script that sets up a ray cluster with the specification as above (1 Ray worker node per physical node / 8 GPUs per Ray worker node), but over `<NUM_NODES>` physical nodes.
+
+```console
+#!/bin/bash
+#SBATCH --nodes=<NUM_NODES>+1
```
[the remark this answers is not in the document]
keshavb96: Yes this is possible and is what the `Starting a Ray Cluster Manually` section and the toy example do. But in general, wouldn't we want the coordinator / control plane to be in its own node and not suffer from the same failure modes as the node that are involved in the computation?

docs/resiliency/ray_resilient_jax.md
```diff
@@ -13,6 +13,9 @@ python3 -m pip install "ray[default]" redis
 ```
 
 The need for redis will be explained a little later in this guide.
```
gspschmid, line 15: If we're not explaining why, then we might as well not mention it here.

docs/resiliency/ray_resilient_jax.md
```diff
@@ -13,6 +13,9 @@ python3 -m pip install "ray[default]" redis
...
+
+While the tutorial below, is comprehensive and detailed it is only meant to be a guide to be able to use Ray to implement fault tolerant workloads of your own. It is not presented in a way that one can just copy and paste its contents to produce a runnable example. For a full runnable example, please refer to the [simple toy example](./example/) that includes a simplified example that can be run on a single node (with at least 2 GPUs) in a fault tolerant manner.
```
gspschmid, line 17: Put the bottom line up front -- I would remove everything from "It assumes familiarity ..." until here (including the installation instructions, which are part of the example Dockerfile anyways). I would replace all of this by something like:

> For a full runnable example of resilient JAX training using Ray see *the-example-directory*, which includes a Docker image you can try out on a single node with a least two GPUs. The guide below gives an overview of the example code and how you might want to extend it to your own model.

docs/resiliency/example/ray_example.py
gspschmid: @keshavb96 Ideally we should deterministically fail in different variants, so the reader (and ourselves) are convinced that various kinds of failures are handled correctly. E.g. first failure is only segfault on rank0, then only segfault on some other rank. Then timeout on rank0, etc.

In the future we should also think about simulating failures not just at the end of step, but also in the middle of a step (e.g. what if a rank fails during `train_step` or during `.checkpoint`?).

docs/ray_resilient_jax.md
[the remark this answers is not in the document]
gspschmid: Why would the `N+1`-th node not be subject to the same failure modes the other `N` nodes are?

docs/ray_resilient_jax.md
```diff
@@ -0,0 +1,658 @@
...
+import redis
+        
+class RayClusterCoordinator:
+  def __init__(self, worker_cls, num_workers) -> None:
+    self.worker_cls = worker_cls
+    self.num_workers = num_workers
+  
+    self.workers = [worker_cls.options(num_gpus=1, 
+                                       num_cpus=16, 
+                                       resources={"worker_units": 1}).remote() for _ in range(self.num_workers)]
```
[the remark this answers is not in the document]
gspschmid: This still seems to be the case in some code snippets.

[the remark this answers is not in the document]
keshavb96: Fixed

docs/resiliency/example/ray_example.py
keshavb96: The initial intention was to leave room for something a little more complicated. Perhaps not a hearbeat every iteration, but a heartbeat every so often, or something more clever even. In this case, you're right, that's all it's doing. The idea was to just encapsulate the heartbeat functionality, no matter how complex it got in the context manager implementation, so that a user could always just say with self.EnableHeartbeat() no matter how complex the heartbeat mechanism was. Should I maybe mention this? Or just change it?

docs/resiliency/example/ray_example_driver.py
keshavb96: Resolving this here responding to it below

docs/resiliency/ray_resilient_jax.md
```diff
@@ -646,8 +535,106 @@ logs = client.get_job_logs(job_id)
 print(logs)
```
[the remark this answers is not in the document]
keshavb96: So workers forward logs to the driver by default. The issue is that when you launch a job on a Ray cluster that the logs can be found in /tmp/ray/session_latest. They don't get outputted to stdout and I haven't found any other way to "show" runtime progress without getting the logs and dumping them to stdout. You could use something like `tail -f` but that seems a little clunky to me. 

So it seems like the decision to make is whether we "re-route" the logs continuously to stdout or if it's ok to just tell them to look in /tmp/ray/session_latest. Thoughts?

docs/ray_resilient_jax.md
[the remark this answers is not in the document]
keshavb96: The idea was that the node that is designated as the control plane could be different in a number of ways. If they are all the same node in the underlying hardware sure, they would be subject to the same failure modes. But I was imagining that the head node would run on a node that didn't even have GPUs. Of course in this example it runs on just another compute node in the cluster.

gspschmid (approved):
One comment re triggering multiple non-rank-0 failures

PR #1315, opened 2025-02-24
Update symlnk-cudnn.sh

yhtang:
libcudnn is renamed to nvidia-cudnn in CUDA 12.9


DwarKapex (approved):

olupton:
I think this is stale, please ping internally if it's still needed for something.

PR #1318, opened 2025-02-26
jax-toolbox-triage: support srun --container-XXX as an alternative backend to Docker

olupton:
The main point is that this enables triaging on Slurm clusters whose nodes do not have `docker` installed. This should implicitly enable triaging multi-process tests too.

### DIFF
--- a/docs/triage-tool.md
+++ b/docs/triage-tool.md
@@ -54,7 +54,11 @@ virtualenv triage-venv
 ```
 
 The tool should be invoked on a machine with `docker` available and whatever GPUs are
-needed to execute the test case.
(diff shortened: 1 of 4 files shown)


docs/triage-tool.md
```diff
@@ -54,7 +54,11 @@ virtualenv triage-venv
 ```
 
 The tool should be invoked on a machine with `docker` available and whatever GPUs are
-needed to execute the test case.
+needed to execute the test case if the default runtime (`--container-runtime=docker`)
+is used.
+If `--container-runtime=pyxis` is used instead, the tool should be invoked on a machine
```
gpupuck, line 59: wow, nice!

gpupuck (approved):

PR #1320, opened 2025-02-27
Move to CUDA DL 25.02 base container

olupton:


.github/container/Dockerfile.base
```diff
@@ -42,6 +42,7 @@ apt_packages=(
   vim
   wget
   jq
+  zip
```
olupton, line 45: This was causing a unit test in JAX to fail.

.github/container/nsys_jax/nsys_jax/scripts/patch_nsys.py
```diff
@@ -899,6 +899,456 @@ def consolidate_ranges(range_df):
              "start": "Start",
 '''
 
+patch_content_2025_1_1 = r'''diff --git a/nsys_recipe/lib/args.py b/nsys_recipe/lib/args.py
```
olupton, line 902: This is just a "rebase" of the patch in https://github.com/NVIDIA/JAX-Toolbox/pull/1323; it shouldn't be needed next time.

DwarKapex (approved):

PR #1322, opened 2025-02-27
Add `TransformerEngine` test workflow

aybchan:

PR #1332, opened 2025-03-07
Improve error handling, s3 mounting, distributed tests for `axlearn`

Steboss:


README.md
```diff
@@ -269,7 +269,7 @@ We support and test the following JAX frameworks and model architectures. More d
       </td>
       <td>
         <a href="https://gist.github.com/nvjax/913c2af68649fe568e9711c2dabb23ae#file-badge-maxtext-test-json">
-          <img style="height:1em;" src="https://img.shields.io/endpoint?url=https%3A%2F%2Fgist.githubusercontent.com%2Fnvjax%2F913c2af68649fe568e9711c2dabb23ae%2Fraw%2Fbadge-axleran-test.json&logo=nvidia&label=A100%20distributed">
+          <img style="height:1em;" src="https://img.shields.io/endpoint?url=https%3A%2F%2Fgist.githubusercontent.com%2Fnvjax%2F913c2af68649fe568e9711c2dabb23ae%2Fraw%2Fbadge-axlearn-test.json&logo=nvidia&label=A100%20distributed">
```
olupton, line 272: H100, not A100?

.github/actions/submit-delete-k8s-job/action.yml
```diff
@@ -32,6 +33,37 @@ runs:
...
           kubectl logs --all-containers=true --all-pods=true --follow job/${{ inputs.job-name }}
-          
+
+          # Check whether the job succeeded or failed 
+          while readarray -d : -t status < <(kubectl get job/${{ inputs.job-name }} -o 'jsonpath={.status.failed}:{.status.succeeded}'); do
+            failures="${status[0]:-0}"
+            successes="${status[1]:-0}"
+            total=$((failures + successes))
+
+            if [[ $total -lt 2 ]]; then
```
olupton, line 43: I believe this needs to match the parallelism configured in the job, which is not immediately known in this workflow.

Either it should be an input, or it should be extracted programmatically here.

Steboss:
@olupton this PR introduces: 
- error handling in `actions/submit-delete-k8s-job` 
- borrowed from #1322 distributed tests  scripts (we can merge #1322 first, and then I can solve any conflict) 
- `fuji-train-perf.py` for performance run tests on bigger models 
- cleaned up `symlnk-cudnn.sh` and `symlnk-nccl.sh` 
- `axlearn-job.yml` deployment for distributed tests in `axlearn` 
- `.pre-commit-config.yaml` for pre-commit setup (if we want to use pre-commit, I can run it in another PR, so we can fix existing files and add some blobs in the README about it)

.github/actions/submit-delete-k8s-job/action.yml
```diff
@@ -33,5 +33,50 @@ runs:
...
+          # if parallelism is not set, use default value of 1 
+          echo "Parallelism ${parallelism}"
+          if [ -z "${parallelism}" ]; then
+            echo "No parallelism specified, defaulting to 1"
+            parallelism=1
+          fi
+
+          while IFS=: read -r failures successes; do
+            failures="${status%:*}"
+            successes="${status#*:}"
```
olupton, line 47: I'm puzzled by this, where does `status` come from?

I guess this was originally based on https://github.com/NVIDIA/JAX-Toolbox/blob/b0ec72a3e72600379453eb320afec9a45538e9c3/.github/workflows/_test_maxtext_k8s.yaml#L77 somehow?

.github/container/fuji-train-perf.py
```diff
@@ -0,0 +1,187 @@
...
+    f"======================\n"
+    )
+
+    # Build the model config
+    config_fn = c4_trainer.named_trainer_configs()[config_name]
+    trainer_config: SpmdTrainer.Config = config_for_function(config_fn).fn()
+    # Intra-node parallelism --> multinode, fsdp = num of gpus in single node
+    ici_mesh_shape = mesh_shape_from_axes(fsdp=fsdp_size)
+    # Inter-node parallelism --> multinode, dp_size = numb of nodes
+    dcn_mesh_shape = mesh_shape_from_axes(data=dp_size)
```
olupton, line 153: IIUC this is hardcoding that this script will only work if executed with `--fsdp <gpus_per_node> --dp <number_of_nodes>`

.github/container/symlnk-cudnn.sh
```diff
@@ -10,7 +10,7 @@ CUDNN_MAJOR_VERSION=9
 prefix=/opt/nvidia/cudnn
 if [[ -d "${prefix}" ]]; then
   echo "Skipping link farm creation"
-  exit 1
+  exit 0
```
olupton, line 13: Why was this needed?

.github/container/symlnk-nccl.sh
```diff
@@ -8,7 +8,7 @@ set -ex -o pipefail
 prefix=/opt/nvidia/nccl
 if [[ -d "${prefix}" ]]; then
   echo "Skipping link farm creation"
-  exit 1
+  exit 0
```
olupton, line 11: Why was this needed?

.github/eks-workflow-files/axlearn/axlearn-job.yml
```diff
@@ -15,46 +13,41 @@ spec:
...
-                  volumeMounts:
-                    - name: output
-                      mountPath: /opt/output
-                - name: upload
-                  image: amazon/aws-cli
+                      LOG_DIR="/opt/output/logs"
+                      mkdir -p ${LOG_DIR}
+
+                      # Start MPS daemon
+                      nvidia-cuda-mps-control -d &
```
olupton, line 25:
```suggestion
                      nvidia-cuda-mps-control -d
```
-d will make it detach on its own

.github/workflows/_ci.yaml
```diff
@@ -66,6 +66,23 @@ jobs:
...
+      BADGE_FILENAME: badge-triton-build
+      BUILD_DATE: ${{ inputs.BUILD_DATE }}
+      BASE_IMAGE: ${{ needs.build-jax.outputs.DOCKER_TAG_MEALKIT }}
+      CONTAINER_NAME: triton
+      DOCKERFILE: .github/container/Dockerfile.triton
+      RUNNER_SIZE: large
+      EXTRA_BUILD_ARGS: |
+        URLREF_JAX_TRITON=${{ fromJson(inputs.SOURCE_URLREFS).JAX_TRITON }}
+    secrets: inherit
+
```
olupton, line 85: rebase mix-up? this was deleted in main. also later in the file.

.github/container/fuji-train-perf.py
olupton, line 1: You should test this by implementing https://github.com/NVIDIA/JAX-Toolbox/blob/a0f5c502d430bd40c5e96f6ce37736b2f63cbe7d/.github/eks-workflow-files/axlearn/axlearn-fuji-model.yml using it (and thereby make it faster by running fewer steps).

.github/container/fuji-train-perf.py
Steboss: The comment is a bit misleading and I used it as a reminder for myself. Actually for any value of `fsdp` and `dp` the code would work

.github/container/symlnk-cudnn.sh
Steboss: This is an error  that @gpupuck spotted and comes from the [old code](https://github.com/NVIDIA/JAX-Toolbox/blob/505b7418e6375c750a50d1de7b4297f62f4d4eba/.github/container/install-cudnn.sh#L51), and we don't need to have an `exit 1` as it's not an error. Maybe we could avoid having `exit` here?

.github/container/symlnk-nccl.sh
Steboss: Same as above

.github/actions/submit-delete-k8s-job/action.yml
Steboss: you're right Olli, I started from there and I left some changes. 
The actual values of `failures` and `successes` are read from 
```
done <<EOF
          $(kubectl get job/"${{ inputs.job-name }}" -o 'jsonpath={.status.failed}:{.status.succeeded}')
          EOF
```
and passed in the while loop every time. I think I left some typos from previous modifications.

.github/container/symlnk-cudnn.sh
olupton: I am still curious what workflow is running this twice, do you know? @gpupuck ?

.github/eks-workflow-files/axlearn/axlearn-job.yml
```diff
@@ -15,46 +13,41 @@ spec:
...
-                  image: amazon/aws-cli
+                      LOG_DIR="/opt/output/logs"
+                      mkdir -p ${LOG_DIR}
+
+                      # Start MPS daemon
+                      nvidia-cuda-mps-control -d
+                      # Run tests
+                      pytest-xdist.sh 1 6 axlearn-unittests.jsonl test-axlearn.sh --directory "." --output "/opt/output/" --test-files "/opt/axlearn/axlearn/common/*_test.py" | tee -a ${LOG_DIR}/pytest_stdout.log
+
+                      wait
```
olupton, line 29: What's this for?

.github/workflows/_ci.yaml
```diff
@@ -682,9 +683,10 @@ jobs:
         path: |
           sitrep.json
           badge-axlearn-test.json
+          axlearn-unittests.jsonl
           axlearn-output/*
 
-  # the fuji test will run for 20 minutes only, as per 2025-02-24 
+  # the fuji test will run for 20 minutes only, as per 2025-02-24
   # is not possible to set the `max_steps` value
   # this will be done with a customer python code
```
olupton, line 691: This is done now, right?

.github/eks-workflow-files/axlearn/axlearn-job.yml
```diff
@@ -15,46 +13,41 @@ spec:
...
-                      # Upload logs to S3 bucket
-                      aws s3 cp /opt/output/ s3://jax-toolbox-eks-output/axlearn/${RUN_ID}/ --recursive --exclude "*" --include "*.log"
+                  resources:
+                    limits:
+                        nvidia.com/gpu: 8
                   volumeMounts:
                     - name: output
                       mountPath: /opt/output
+                    - name: s3-storage
+                      mountPath: /jax-toolbox-eks-output
```
olupton, line 45: Can we use `subPath` to avoid the manual/error-prone part of remembering to copy to unique paths under `/jax-toolbox-eks-output`? Also, that path could be a more succinct `/output` or whatever, right?

Steboss:
@olupton  
Re fuji-train-perf.py](https://github.com/NVIDIA/JAX-Toolbox/pull/1332/files/e95e0907671b277e5c1781a5c7da33f497b48d9c#diff-d172ca086cdf084c1f4dfb43e58a8bc9bc220a91e98e3ed1450a68d70afaedfc):
```
Sure, with the caveat that something will probably complain if the product of all the mesh dimensions doesn't match the global device count.

But my point was that this hardcodes the only permitted inter-node parallelism is DP, and the only permitted intra-node parallelism is FSDP, right?
```
Correct :)

.github/container/fuji-train-perf.py
Steboss: ah sorry, now I can see what you mean. 
Nope we can have also tensor and pipeline parallelism. I can add those options as part of the call

.github/eks-workflow-files/axlearn/axlearn-job.yml
Steboss: Beautiful shout, thanks Olli

.github/container/pytest-xdist.sh
olupton, line 1: Looks like an accidental change.

.github/container/fuji-train-perf.py
```diff
@@ -0,0 +1,266 @@
...
+    trainer_config.summary_writer.write_every_n_steps = write_summary_steps
+    # Call AXLearn Jax setup
+    launch.setup()
+    # Setup the config
+    trainer_config.set(
+        recorder=config_for_function(lambda: measurement.global_recorder)
+    )
+    # Launch training
+    with open(output_log_file, "w") as f:
+        with redirect_stdout(f), redirect_stderr(f):
```
olupton, line 250: I'm not too excited about this approach...

.github/container/fuji-train-perf.py
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+    times_arr = np.array(times, dtype=np.float32)
+    # Metrics
+    tokens_per_sec_gpu = (gbs * seq_len) / times_arr / world_size
+    seqs_per_sec_gpu = (gbs) / times_arr / world_size
+
+    return {
+        "tokens_per_sec_gpu_mean": float(tokens_per_sec_gpu.mean()),
+        "tokens_per_sec_gpu_std": float(tokens_per_sec_gpu.std()),
```
olupton, line 134: Is it easy to tweak the test config so this isn't just zero? In the CI logs it looks like there's just a single value, and it's not easy to tell if it's correctly avoiding including compilation/initialisation time. If not, and the 0th step is biased slow, taking the mean isn't going to be a good approach.

.github/container/fuji-train-perf.py
```diff
@@ -0,0 +1,266 @@
...
+        for line in f:
+            match = pattern.search(line)
+            if match:
+                times.append(float(match.group(1)))
+
+    if not times:
+        return {
+            "tokens_per_sec_gpu_mean": 0.0,
+            "tokens_per_sec_gpu_std": 0.0,
+            "seqs_per_sec_gpu_mean": 0.0,
```
olupton, line 121: Did you look into calculating an equivalent TFLOPS/s/gpu metric to what we report with MaxText?

olupton (approved):
I think we can get this merged, it's been open a while.

That said, I think the metric/perf extraction is going to need some more changes.

.github/container/fuji-train-perf.py
Steboss: yes, me too, but it's the only way to redirect `absl` stdout and stderr to the output file

.github/container/fuji-train-perf.py
Steboss: definitely easy 
I think we should run at least 301 steps to avoid having this. 
The first 100 steps include the 0th step (unfortunately we can't get the time for a single step, but we'll have an average over 100 steps by default), the other 200 steps will help improving the mean and the stdev. What do you think?

.github/container/fuji-train-perf.py
Steboss: I can check if we can get similar info

.github/container/fuji-train-perf.py
olupton: Is the 100 value (for how often it prints) hardcoded?

Did you try the tensorboard suggestion?

olupton (commented):
Generally LGTM, see separate comment about tensorflow

.github/container/Dockerfile.axlearn
```diff
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1-labs
 ARG BASE_IMAGE=ghcr.io/nvidia/jax-mealkit:jax
-ARG URLREF_AXLEARN=https://github.com/apple/axlearn.git#main
+ARG URLREF_AXLEARN=https://github.com/Steboss/axlearn.git#main
```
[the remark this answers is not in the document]
Steboss: Correc @olupton

.github/eks-workflow-files/axlearn/axlearn-fuji-model.yml
```diff
@@ -20,39 +18,29 @@ spec:
...
+                          --ici_fsdp=8 \
+                          --dcn_dp=1 \
+                          --gbs=16 \
+                          --ga=1 \
+                          --seq_len=4096 \
+                          --max_step=301 \
+                          --save_checkpoint_steps=100 \
+                          --write_summary_steps=100 \
+                          --output_log_file=${OUTPUT_LOG_FILE} \
+                          --world_size=8
```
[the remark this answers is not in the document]
Steboss: Correct

olupton (approved):

PR #1338, opened 2025-03-10
Slim down CI configurations, remove triton container

olupton:
- JAX-Triton compatibility with upstream Triton wheels is now said to be better, drop the `triton` JAX-Toolbox container
- Remove some MaxText and T5X test configurations; JAX-Toolbox CI is not intended to be exhaustive


.github/workflows/_test_t5x_rosetta.yaml
```diff
@@ -415,180 +415,180 @@ jobs:
...
-      matrix:
-        N_GPU: [8]
-      fail-fast: false
-    runs-on: jumpbox
-    env:
-      BADGE_FILENAME_PREFIX: badge-rosetta-t5x-vit-single-process-multi-device
-    steps:
-      - name: Print environment variables
-        run: env
+  # vit-single-process-multi-device:
```
Steboss, line 418: Do we want to keep these processes for the future, or we can delete them?

Steboss (approved):
Just a little comment for curiosity

.github/workflows/_test_t5x_rosetta.yaml
olupton: I think the whole file will probably go away soon, so I wasn't too worried about keeping it clean.

PR #1341, opened 2025-03-12
Add MACE example from Mario Geiger

DwarKapex:
All credits and many thanks for MACE examples to @mariogeiger


mariogeiger:
I have a new version using cuEquivariance 0.3.0 that involves custom cuda kernels. It is much faster but you might prefer this "pure jax" version for this project?

PR #1353, opened 2025-03-21
nsys-jax: improve alignment for multi-GPU runs

olupton:
This removes assumptions about the ordering of thunks being identical
across different GPUs, which are not correct in the presence of
overlapped communications and enough jitter.

The summary analysis script gains a wait time metric and prettier formatting.

A race condition in the CI config is fixed.


Steboss (approved):
Great job!

.github/workflows/_test_unit.yaml
```diff
@@ -104,6 +104,9 @@ jobs:
             fi
             echo "badge_message=${passed_tests}/${total_tests} passed" >> $GITHUB_OUTPUT
             echo "summary=${{ inputs.TEST_NAME }} unit test on ${{ matrix.GPU_ARCH }}: ${total_tests} total tests, ${errors} errors, ${passed_tests} passed, ${failed_tests} failed." >> $GITHUB_OUTPUT
+            if [[ ${passed_tests} < ${total_tests} ]]; then
```
Steboss, line 107: Just a comment
This will make the step `name: Summarize test statistics` to fail, so no badge creation. Are we ok with this?

olupton: The next two have `if: "!cancelled()"` so I thought it should be OK.

olupton: I can see a badge in https://github.com/NVIDIA/JAX-Toolbox/actions/runs/14054259487/job/39351322118?pr=1353 even though the job failed

PR #1382, opened 2025-04-07
Add nvcr.io authorization and ping tf-text to 2.18.1

DwarKapex:


.github/container/Dockerfile.maxtext
```diff
@@ -23,6 +23,8 @@ for pattern in \
     "s|protobuf==3.20.3|protobuf>=3.19.0|g" \
     "s|tensorflow-datasets|tensorflow-datasets>=4.8.0|g" \
     "s|sentencepiece==0.1.97|sentencepiece>=0.2|g" \
+    "s|tensorflow-text>=2.13.0|tensorflow-text==2.18.1|g" \
+    "s|tensorflow>=2.13.0|tensorflow==2.18.1|g" \
```
gpupuck, line 27: We don't actually install `tensorflow`, instead they are
```
tensorflow-text: 2.18.1
tensorflow_cpu: 2.18.1
```

olupton: https://github.com/NVIDIA/JAX-Toolbox/blob/d55bfa76bfeb93b91202f5abdc3f63a066388d70/.github/container/pip-finalize.sh#L49-L54 should mean `tensorflow` here is fine.

olupton (approved):
LGTM, though we'd need an internal trigger to be sure

PR #1401, opened 2025-04-17
Infer compute cap from CUDA_ARCH_LIST if possible

michael132:
Enhance CUDA architecture support in build-jax.sh

- Added logic to infer compute capabilities from the CUDA_ARCH_LIST variable if set.
- Updated the supported_compute_capabilities function to handle both CUDA_ARCH_LIST and predefined architectures (amd64, arm64).

### DIFF
--- a/.github/container/build-jax.sh
+++ b/.github/container/build-jax.sh
@@ -8,9 +8,24 @@ print_var() {
     echo "$1: ${!1}"
 }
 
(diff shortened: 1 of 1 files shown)


michael132:
Switched to a branch in the main repo to avoid CI issue
https://github.com/NVIDIA/JAX-Toolbox/pull/1406

PR #1409, opened 2025-04-22
Fix redundant install of JAX wheels

yhtang:
There is a `pip install` in build-jax.sh that installs the JAX wheels in the 'builder' stage. Since the stage is later discarded, this installation is useless and unnecessary during container build (but still useful for code development).


yhtang:
I almost forgot... Since the redundant installation happens in a builder stage, it doesn't bloat our mealkit/final images. Nonetheless it is still better to eliminate the unnecessary installation.

gpupuck:
Just for my understanding, this will leave the final installations during `pip-finalize.sh`, correct?

yhtang:
> Just for my understanding, this will leave the final installations during `pip-finalize.sh`, correct?

Yes, that is correct.

olupton (approved):

PR #1416, opened 2025-04-24
Ensure JAX installation is editable

yhtang:
Currently, the JAX Python source that we **installed** in the container is created by `build-jax.sh` (via _python build.py build --editable ... --wheels=**jax**,jaxlib,jax-cuda-plugin,jax-cuda-pjrt_)  and placed under `/opt/jaxlibs/jax`:

```
$ docker run --gpus all ghcr.io/nvidia/jax:jax python -c 'import jax; print(jax.__path__)'
['/opt/jaxlibs/jax/jax']
```

As a result, two copies of the JAX source exist in the container at `/opt/jax` and `/opt/jaxlibs/jax`, leading to potential confusions. Also, this defeats the 'editable' purpose of the build, as mounting a JAX working copy from outside the container to `/opt/jax` will not replace the installed JAX package.

Given that the editable wheel at `/opt/jaxlibs/jax` is practically identical to the original source (sans a number of build-related files), we should bypass creating this extra copy and directly specify `/opt/jax` as the installation target in the Dockerfile and build script. With this change, we will have
```
>>> import jax; print(jax.__path__)
['/opt/jax/jax']
```
which is the desired usage pattern.

### DIFF
--- a/.github/container/Dockerfile.jax
+++ b/.github/container/Dockerfile.jax
@@ -106,7 +106,7 @@ RUN <<"EOF" bash -ex
 for component in $(ls ${BUILD_PATH_JAXLIB}); do
     echo "-e file://${BUILD_PATH_JAXLIB}/${component}" >> /opt/pip-tools.d/requirements-jax.in;
 done
-echo "-e file://${BUILD_PATH_JAXLIB}/jax[k8s]" >> /opt/pip-tools.d/requirements-jax.in
+echo "-e file://${SRC_PATH_JAX}[k8s]" >> /opt/pip-tools.d/requirements-jax.in
 EOF
 
 ## Flax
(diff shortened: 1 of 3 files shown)


gpupuck:
You'll need to change `test-jax.sh` also

yhtang:
> test-jax.sh

done 😁

gpupuck (commented):
```
pip --disable-pip-version-check install .... -e ${BUILD_PATH_JAXLIB}/jax

    ## after installation (example)
    # jax                     0.5.4.dev20250325    /opt/jaxlibs/jax   <--------
    # jax-cuda12-pjrt         0.5.4.dev20250325    /opt/jaxlibs/jax_cuda12_pjrt
```
Install dir and resulting output needs to be updated too

yhtang:
> ```
> pip --disable-pip-version-check install .... -e ${BUILD_PATH_JAXLIB}/jax   <------
> ...
> ## after installation (example)
> # jax                     0.5.4.dev20250325    /opt/jaxlibs/jax   <--------
> # jax-cuda12-pjrt         0.5.4.dev20250325    /opt/jaxlibs/jax_cuda12_pjrt
> ```
> 
> Install dir and resulting output needs to be updated too

Fixed

gpupuck (approved):

PR #1436, opened 2025-05-06
build-jax.sh, test-jax.sh: cleanup and de-duplicate

olupton:
`build-jax.sh`:
- Logic for updating `requirements.in` is changed a little, now that all three wheels are included in the upstream version.
- Drop many `TF_*` environment variable settings.
- Build TransformerEngine for the same GPU architectures as JAX.
- Do not delete the `.jax_configure.bazelrc` file, so `test-jax.sh` does not have to recreate it. Drop the `.local_cuda.bazelrc` hack as a result.
- Delete old `--dry` and `--xla-arm64-patch` options.

`test-jax.sh`:
- Add new `--tests-per-gpu` and `--visible-gpus` options to steer how many devices are visible to each test shard.
- Do not call `build.py`, as `build-jax.sh` now leaves the configuration file intact.
- Propagate `JAX_ENABLE_X64` setting to tests.

CI: run `single-gpu` and `multi-gpu` test batteries.


.github/container/Dockerfile.jax
```diff
@@ -31,41 +31,46 @@ ARG BUILD_PATH_JAXLIB
...
 
-
 RUN --mount=type=ssh \
     --mount=type=secret,id=SSH_KNOWN_HOSTS,target=/root/.ssh/known_hosts \
 <<"EOF" bash -ex
     git-clone.sh ${URLREF_JAX} ${SRC_PATH_JAX}
     git-clone.sh ${URLREF_XLA} ${SRC_PATH_XLA}
 EOF
 
-ADD build-jax.sh local_cuda_arch test-jax.sh /usr/local/bin/
```
gpupuck, line 42: I think we still need `test-jax.sh` to be in `/usr/local/bin` for internal unit tests.

.github/container/build-jax.sh
```diff
@@ -248,117 +200,74 @@ print_var CLEAN
...
-    --local_xla_path=$SRC_PATH_XLA \
-    --output_path=${BUILD_PATH_JAXLIB} \
+    "--local_xla_path=${SRC_PATH_XLA}" \
+    "--output_path=${BUILD_PATH_JAXLIB}" \
     $BUILD_PARAM
-popd
 
 # Make sure that JAX depends on the local jaxlib installation
 # https://jax.readthedocs.io/en/latest/developer.html#specifying-dependencies-on-local-wheels
-local_jax_whl="jax @ file://${SRC_PATH_JAX}"
```
gpupuck, line 323: You dropped `jax @ file:///opt/jax` here, would that cause dependency mismatch during `pip-finalize.sh`?

.github/container/test-jax.sh
```diff
@@ -26,19 +28,32 @@ SRC_PATH_JAX="/opt/jax"
 BUILD_JAXLIB=0
```
gpupuck, line 28: Previously, we removed
```
jax_source_dir() {
    dirname `python -c "import jax; print(*jax.__path__)"`
}
```
Because `jax` was not editable anymore, now it is editable again. Should we add this logic back?

.github/container/test-jax.sh
```diff
@@ -89,112 +112,103 @@ while [ : ]; do
...
+        set_default VISIBLE_GPUS 1
+        FLAGS+=("//tests:backend_independent_tests")
+        ;;
+    multi-gpu)
+        set_default JOBS_PER_GPU 8
+        # TODO: unclear if JOBS_PER_GPU>1 will cause deadlocks for any tests in the suite
+        set_default VISIBLE_GPUS all
+        # --test_tag_filters=A,B is an OR of A and B, not an AND
+        FLAGS+=(
+            "--test_tag_filters=multiaccelerator"
```
gpupuck, line 160: Is `multiaccelerator` a subset of `jax_test_gpu`?

.github/container/build-jax.sh
olupton: This requirements file is not relevant to `pip-finalize.sh`'s dependency resolution, it's used to resolve which versions of wheel dependencies bazel uses when running JAX's tests.

I think it was a bug that it was included here in the first place.

.github/container/test-jax.sh
olupton: The idea is that `jax_test_gpu` is partitioned into `single-gpu` and `multi-gpu` using the `multiaccelerator` tag.

.github/container/test-jax.sh
olupton: I'm ambivalent, it seems a bit fragile to re-bake an implicit assumption that it is, in fact, editable.

gpupuck (dismissed):

PR #1437, opened 2025-05-07
Use latest PyPI pip

olupton:
- Drop the pin to 23.3.1 and patch introduced in #405


olupton:
@aybchan will take a look separately

PR #1439, opened 2025-05-07
Add JetStream to /opt/jetstream

yhtang:
`jetstream/tools` contains checkpoint converters that are not shipped in the wheel. This is a stop-gap solution and the longer term solution is to have upstream ship a CLI for the converters.


gpupuck (dismissed):
LGTM

gpupuck (approved):

PR #1443, opened 2025-05-09
Fix error in AXLearn tests, and remove unnecessary ones

Steboss:
This PR does the following: 
- remove `array_serialization_test.py` that's causing the test to hang with the following error: 
```bash 
[pod/axlearn-14925960362-xb57t/axlearn]   File "/opt/jax/jax/experimental/array_serialization/serialization.py", line 193, in __del__
[pod/axlearn-14925960362-xb57t/axlearn]     logger.warning('Please add `.wait_until_finished()` in the main thread '
[pod/axlearn-14925960362-xb57t/axlearn] Message: 'Please add `.wait_until_finished()` in the main thread before your program finishes because there is a possibility of losing errors raised if the this class is deleted before writing is completed.'
[pod/axlearn-14925960362-xb57t/axlearn] Arguments: ()
[pod/axlearn-14925960362-xb57t/axlearn] sssssssssssssss.ssssssssssss.ssssssssssssssssssssssFsFs.s...sssssssFs.Fs [ 97%]
```
causing the EKS job to run out of time - so we can't get the tests 
- remove tests that are redundant (namely, there are similar test already running) such as: 
```bash 
"/opt/axlearn/axlearn/common/deberta_test.py"
"/opt/axlearn/axlearn/common/distilbert_test.py"
"/opt/axlearn/axlearn/common/trainer_test.py"
"/opt/axlearn/axlearn/common/decoder_test.py"
"/opt/axlearn/axlearn/common/adapter_torch_test.py"
"/opt/axlearn/axlearn/common/attention_test.py"
"/opt/axlearn/axlearn/common/convolution_test.py"
```
- remove tests for models that we're not currently using: 
```bash 
"/opt/axlearn/axlearn/common/mixture_of_experts_test.py"
"/opt/axlearn/axlearn/common/t5_test.py"
"/opt/axlearn/axlearn/common/vision_transformer_test.py"
"/opt/axlearn/axlearn/common/input_reading_comprehension_test.py"
"/opt/axlearn/axlearn/common/input_t5_test.py"
```
- remove tests like `summary_writer_test.py` that is mostly using python library we're not employing here (e.g. `wandb`)
- add the installation of `pytest-xdist` and `pytest-reportlog` to avoid the following error: 
```bash 
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --report-log=/tmp/tmp.iL7rVQtXBq --dist=load --tx --tx 
```

Overall, this should allow us to reduce the testing time from 50 minutes to 30 minutes, covering the most important tests as well, that are dealing with general AXLearn infrastructure.


.github/container/test-axlearn.sh
```diff
@@ -117,30 +117,42 @@ if [ "${#expanded_test_files[@]}" -eq 0 ]; then
...
-
-if [ -f "$EXCLUDE_LIST_FILE" ]; then
-    echo "Reading exclusion list from '$EXCLUDE_LIST_FILE'"
-    mapfile -t EXCLUDE_PATTERNS < "$EXCLUDE_LIST_FILE"
-else
-    echo "Exclusion list file not found at '$EXCLUDE_LIST_FILE'"
-fi
+# exclude all those tests that are not necessary because of time or error
+EXCLUDE_PATTERNS=(
+  "/opt/axlearn/axlearn/common/array_serialization_test.py"
```
olupton, line 122: It's a bit of a shame to have the `/opt/axlearn` prefix here.

olupton (commented):
I see an error that seems like a test/infra bug, rather than a failing test: https://github.com/NVIDIA/JAX-Toolbox/actions/runs/14931992154/job/41952767038#step:6:41440

Some of the tests look like they are failing due to missing input data.

Also, the CI job is marked as successful despite the tests failing.

.github/eks-workflow-files/axlearn/axlearn-job.yml
```diff
@@ -24,8 +24,7 @@ spec:
                       # Start MPS daemon
                       nvidia-cuda-mps-control -d
                       # Run tests
-                      pytest-xdist.sh 1 6 ${LOG_DIR}/axlearn-unittests.jsonl test-axlearn.sh --directory "." --output ${LOG_DIR} --test-files "/opt/axlearn/axlearn/common/*_test.py" | tee -a ${LOG_DIR}/pytest_stdout.log
```
olupton, line 27: Curious what was wrong with pytest-xdist?

.github/container/test-axlearn.sh
```diff
@@ -117,53 +139,53 @@ if [ "${#expanded_test_files[@]}" -eq 0 ]; then
...
         final_test_files+=("$test_file")
-    fi 
+    fi
+done
+
+# RUN TESTS
+runs=(
+  "|not (gs_login or tpu or high_cpu or fp64 or for_8_devices)|base"
+  "JAX_ENABLE_X64=1|fp64|fp64"
+  "XLA_FLAGS='--xla_force_host_platform_device_count=8'|for_8_devices|8dev"
```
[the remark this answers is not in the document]
Steboss: Spot on, I kept this as I was working on it, it's not needed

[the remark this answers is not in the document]
Steboss: I mean, I was re-workign what AXLearn has offered as bash script for running tests

.github/eks-workflow-files/axlearn/axlearn-job.yml
Steboss: It looks like there was a slow down in the way I was using `pytest-xdist.sh` and the previous `test-axlearn.sh` script. In particular, I didn't realise I was doing the following: 
- The 48 workers I was spining up with `pytest-xdist.sh` weren't actively used. Indeed `test-axlearn.sh` was looping across all the tests, and every file was launching `pytest XX_test.py`. To the 48 workers were created again and again. 
- This means that, considering we have to run about 100 tests, we had 48 x 100 = 4800 pytest worker processes. Each worker had to import the whole code base and have time and space for memory allocations. Out of these, just 1 worker per test was actually used, as I was running a single file per `pytest` command. 
- This should explain why we were taking at least 1h30minute to run the whole test 
- The bug on `array_serialization_test.py` still persists, and that's due to JAX version vs AXLearn, causing the tests to hang 

In the new flow [that was inspired by this script on axlearn](https://github.com/apple/axlearn/blob/main/run_tests.sh), I am actually using pytest distributed within the `test-axlearn.sh` script itself, as AXLearn does. Files are called all at once, and subdivided based on their marker. Moreover, I didn't know, we could use `--dist worksteal -n auto` to allocate better the worker resources while running tests.

Steboss:
In this PR I added a workflow dispatch, so we can test single parts of the CI.
In particular, this may be a better trick to be used during CI, so that we can have `MODE=SOMETHING` to trigger only specific parts of the workflow, rather than testing everything

Steboss:
@olupton 
I can see that the axlearn tests now are working fine. We have some tests that are still failing - I can have a look at those. 
Working on the way we're monitoring and returning the k8s job status, as it's still giving green flag on the axlearn eks job.

Steboss:
@olupton 
It looks like we need the `XLA_FLAGS="--xla_force_host_platform_device_count=8"` for the `for_8_devices` tests, otherwise the XLA tests will fail as: 
```bash 
[pod/axlearn-15024599931-2vfmc/axlearn] ___________________ HostArrayTest.test_fixed_process_shape67 ___________________
[pod/axlearn-15024599931-2vfmc/axlearn] [gw93] linux -- Python 3.12.3 /usr/bin/python3
[pod/axlearn-15024599931-2vfmc/axlearn] 
[pod/axlearn-15024599931-2vfmc/axlearn] self = <axlearn.common.host_array_test.HostArrayTest testMethod=test_fixed_process_shape67>
[pod/axlearn-15024599931-2vfmc/axlearn] platform = 'cpu', mesh_shape = (-1, 2), process_shape = [1]
[pod/axlearn-15024599931-2vfmc/axlearn] partition = PartitionSpec('data', 'model')
[pod/axlearn-15024599931-2vfmc/axlearn] 
[pod/axlearn-15024599931-2vfmc/axlearn]     @parameterized.product(
[pod/axlearn-15024599931-2vfmc/axlearn]         platform=("cpu", "tpu"),
[pod/axlearn-15024599931-2vfmc/axlearn]         mesh_shape=[
[pod/axlearn-15024599931-2vfmc/axlearn]             (-1, 1),  # Fully partitioned along one dim.
[pod/axlearn-15024599931-2vfmc/axlearn]             (2, -1),  # Partitioned along multiple dims.
[pod/axlearn-15024599931-2vfmc/axlearn]             (-1, 2),  # Test the other way.
[pod/axlearn-15024599931-2vfmc/axlearn]             (1, -1),
[pod/axlearn-15024599931-2vfmc/axlearn]         ],
[pod/axlearn-15024599931-2vfmc/axlearn]         process_shape=[
[pod/axlearn-15024599931-2vfmc/axlearn]             # Each process produces single dim.
[pod/axlearn-15024599931-2vfmc/axlearn]             [1],  # Not divisible by number of devices (replicated).
[pod/axlearn-15024599931-2vfmc/axlearn]             [8],  # Divisible by number of devices.
[pod/axlearn-15024599931-2vfmc/axlearn]             [16],  # Multiple elements per device.
[pod/axlearn-15024599931-2vfmc/axlearn]             # Each process produces multiple dims.
[pod/axlearn-15024599931-2vfmc/axlearn]             [1, 1],  # Not divisible by number of devices (replicated).
[pod/axlearn-15024599931-2vfmc/axlearn]             [2, 1],  # Can be partitioned over dim=0, replicated on dim=1.
[pod/axlearn-15024599931-2vfmc/axlearn]             [16, 1],  # Multiple elements per device.
[pod/axlearn-15024599931-2vfmc/axlearn]             [2, 4],  # Can be fully partitioned.
[pod/axlearn-15024599931-2vfmc/axlearn]             [8, 8],  # Can be fully partitioned.
[pod/axlearn-15024599931-2vfmc/axlearn]         ],
[pod/axlearn-15024599931-2vfmc/axlearn]         partition=(
[pod/axlearn-15024599931-2vfmc/axlearn]             DataPartitionType.FULL,
[pod/axlearn-15024599931-2vfmc/axlearn]             DataPartitionType.REPLICATED,
[pod/axlearn-15024599931-2vfmc/axlearn]             PartitionSpec("data"),
[pod/axlearn-15024599931-2vfmc/axlearn]             PartitionSpec("data", "model"),
[pod/axlearn-15024599931-2vfmc/axlearn]         ),
[pod/axlearn-15024599931-2vfmc/axlearn]     )
[pod/axlearn-15024599931-2vfmc/axlearn]     # NOTE: while annotated with `for_8_devices`, this runs on other configurations.
[pod/axlearn-15024599931-2vfmc/axlearn]     @pytest.mark.for_8_devices
[pod/axlearn-15024599931-2vfmc/axlearn]     def test_fixed_process_shape(
[pod/axlearn-15024599931-2vfmc/axlearn]         self,
[pod/axlearn-15024599931-2vfmc/axlearn]         platform: str,
[pod/axlearn-15024599931-2vfmc/axlearn]         mesh_shape: tuple[int, int],
[pod/axlearn-15024599931-2vfmc/axlearn]         process_shape: Sequence[int],
[pod/axlearn-15024599931-2vfmc/axlearn]         partition: Union[DataPartitionType, PartitionSpec],
[pod/axlearn-15024599931-2vfmc/axlearn]     ):
[pod/axlearn-15024599931-2vfmc/axlearn]         """Tests roundtrip host-to-global and global-to-host with fixed process shape."""
[pod/axlearn-15024599931-2vfmc/axlearn]     
[pod/axlearn-15024599931-2vfmc/axlearn] >       mesh_shape = infer_mesh_shape(mesh_shape)
[pod/axlearn-15024599931-2vfmc/axlearn] 
[pod/axlearn-15024599931-2vfmc/axlearn] axlearn/common/host_array_test.py:124: 
[pod/axlearn-15024599931-2vfmc/axlearn] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[pod/axlearn-15024599931-2vfmc/axlearn] 
[pod/axlearn-15024599931-2vfmc/axlearn] mesh_shape = (-1, 2)
[pod/axlearn-15024599931-2vfmc/axlearn] 
[pod/axlearn-15024599931-2vfmc/axlearn]     def infer_mesh_shape(mesh_shape: MeshShape, *, num_devices: Optional[int] = None) -> MeshShape:
[pod/axlearn-15024599931-2vfmc/axlearn]         """Infer the value for -1 from len(jax.devices()) and other dims if there is -1 in mesh shape.
[pod/axlearn-15024599931-2vfmc/axlearn]     
[pod/axlearn-15024599931-2vfmc/axlearn]         Args:
[pod/axlearn-15024599931-2vfmc/axlearn]             mesh_shape: The original MeshShape, which might have -1 in one axis.
[pod/axlearn-15024599931-2vfmc/axlearn]             num_devices: The devices that will be used to construct the mesh.
[pod/axlearn-15024599931-2vfmc/axlearn]                 If None, defaults to len(jax.devices()).
[pod/axlearn-15024599931-2vfmc/axlearn]     
[pod/axlearn-15024599931-2vfmc/axlearn]         Returns
[pod/axlearn-15024599931-2vfmc/axlearn]             A new MeshShape with inferred value for -1.
[pod/axlearn-15024599931-2vfmc/axlearn]         """
[pod/axlearn-15024599931-2vfmc/axlearn]         if -1 not in mesh_shape:
[pod/axlearn-15024599931-2vfmc/axlearn]             return mesh_shape
[pod/axlearn-15024599931-2vfmc/axlearn]     
[pod/axlearn-15024599931-2vfmc/axlearn]         if mesh_shape.count(-1) > 1:
[pod/axlearn-15024599931-2vfmc/axlearn]             raise ValueError(f"Only one axis can be -1 in {mesh_shape=}.")
[pod/axlearn-15024599931-2vfmc/axlearn]     
[pod/axlearn-15024599931-2vfmc/axlearn]         # Handle the case with one -1.
[pod/axlearn-15024599931-2vfmc/axlearn]         prod = math.prod(mesh_shape, start=-1)
[pod/axlearn-15024599931-2vfmc/axlearn]         if num_devices is None:
[pod/axlearn-15024599931-2vfmc/axlearn]             num_devices = len(jax.devices())
[pod/axlearn-15024599931-2vfmc/axlearn]         if num_devices % prod != 0:
[pod/axlearn-15024599931-2vfmc/axlearn] >           raise ValueError(
[pod/axlearn-15024599931-2vfmc/axlearn]                 f"Unable to infer -1 in mesh shape {mesh_shape} as num_devices {num_devices} "
[pod/axlearn-15024599931-2vfmc/axlearn]                 f"is not a multiple of the product {prod} of mesh axes."
[pod/axlearn-15024599931-2vfmc/axlearn]             )
[pod/axlearn-15024599931-2vfmc/axlearn] E           ValueError: Unable to infer -1 in mesh shape (-1, 2) as num_devices 1 is not a multiple of the product 2 of mesh axes.
[pod/axlearn-15024599931-2vfmc/axlearn] 
[pod/axlearn-15024599931-2vfmc/axlearn] axlearn/common/utils.py:1834: ValueError
```
I ran a test, that results in: 
- with flag: 1 failed, 161 passed 
- without flag: 203 failed, 100 passed

Steboss:
This may be a new JAX version error 
```
AttributeError: module 'jax.experimental.array_serialization.serialization' has no attribute '_spec_has_metadata'
```

.github/container/test-axlearn.sh
```diff
@@ -117,53 +137,65 @@ if [ "${#expanded_test_files[@]}" -eq 0 ]; then
...
         final_test_files+=("$test_file")
-    fi 
+    fi
+done
+
+# RUN TESTS
+runs=(
+  "JAX_PLATFORMS='gpu'|not (gs_login or tpu or high_cpu or fp64 or for_8_devices)|base"
+  "JAX_PLATFORMS='gpu' JAX_ENABLE_X64=1|fp64|fp64"
+  "JAX_PLATFORMS='gpu'|for_8_devices|8dev"
```
olupton, line 169: Curious why this needs to be set explicitly? (`JAX_PLATFORMS='gpu'`)

Steboss: This was me still testing. No need to export it, as devices are read correctly. 
The `for_8_devices` just works for CPUs, so I'll take this out from the tests we want to run

.github/eks-workflow-files/axlearn/axlearn-job.yml
```diff
@@ -21,11 +21,14 @@ spec:
                       LOG_DIR="/output/${RUN_ID}"
                       mkdir -p ${LOG_DIR}
 
-                      # Start MPS daemon
-                      nvidia-cuda-mps-control -d
                       # Run tests
-                      pytest-xdist.sh 1 6 ${LOG_DIR}/axlearn-unittests.jsonl test-axlearn.sh --directory "." --output ${LOG_DIR} --test-files "/opt/axlearn/axlearn/common/*_test.py" | tee -a ${LOG_DIR}/pytest_stdout.log
+                      export CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7
```
olupton, line 25: Why is this needed?

.github/workflows/ci.yaml
```diff
@@ -40,6 +40,18 @@ on:
...
         default: ''
         required: false
+      MODE:
+        type: choice
+        description: |
+          This option is to run just specific part in the _ci step.
+          - full - everything will be run (default)
+          - jax - run all the tests related to jax system
+          - te - run all the tests related to TE
+          - rosetta - run build rosetta
```
olupton, line 50: I would rename your `rosetta` group to be `t5x`.

.github/workflows/ci.yaml
```diff
@@ -197,6 +209,7 @@ jobs:
       CUDA_IMAGE: ${{ needs.metadata.outputs.CUDA_IMAGE }}
       MANIFEST_ARTIFACT_NAME: ${{ needs.metadata.outputs.MANIFEST_ARTIFACT_NAME }}
       SOURCE_URLREFS: ${{ needs.bump-manifest.outputs.SOURCE_URLREFS }}
+      MODE: 'axlearn' # ${{ github.event_name == 'workflow_dispatch' && inputs.MODE || 'full' }}
```
olupton, line 212: I guess you need to make `MODE` an input to `ci.yaml` too?
I guess then it could be required as an input to `_ci.yaml`.

.github/container/test-axlearn.sh
```diff
@@ -117,53 +137,71 @@ if [ "${#expanded_test_files[@]}" -eq 0 ]; then
...
+    done
+    if [ "$exclude" = false ]; then
         final_test_files+=("$test_file")
-    fi 
+    fi
+done
+
+# RUN TESTS
+runs=(
+  "|not (gs_login or tpu or high_cpu or fp64 or for_8_devices)|base"
```
olupton, line 174: IIUC we are not running the `for_8_devices` tests at all anymore? Do you think they don't provide useful extra coverage?

.github/workflows/ci.yaml
Steboss: Yes correct, 
`MODE` is an input of `ci.yaml` and it's been added to `_ci.yaml` as input

.github/container/test-axlearn.sh
Steboss: thanks for spotting this, committed, restarted the CI

Steboss: I ran the final tests, and we're going to have also the tests `for_8_devices`, as I think they're relevant for our cases. However, I've kept them separated from the main test suite because: 
- these tests require exactly 8 devices to be present at run time, and they will be likely to fail if there's some shared resource
- it looks like the JAX state was shared across all the tests run. Thus, if a test was using JAX_PLATFORMS='cpu', this was passed to the subsequent test. This has caused all the errors to appear when running the `for_8_devices` test as part of the main loop.

.github/eks-workflow-files/axlearn/axlearn-job.yml
Steboss: Sorry about that, I run a last test on `for_8_devices`. this is not needed anymore

olupton (approved):
I think we can merge this to speed up the pipeline, but I left some more comments on error handling and robustness.

.github/workflows/_ci.yaml
```diff
@@ -565,16 +616,19 @@ jobs:
       id: log-s3
       run: |
         mkdir -p axlearn-output
-        aws s3 cp s3://jax-toolbox-eks-output/axlearn/${{ github.run_id }}/summary.txt axlearn-output/
-        aws s3 cp s3://jax-toolbox-eks-output/axlearn/${{ github.run_id }}/ axlearn-output/ --recursive --exclude "*" --include "*.log"
-        aws s3 cp s3://jax-toolbox-eks-output/axlearn/${{ github.run_id }}/axlearn-unittests.jsonl axlearn-output/
+        aws s3 cp s3://jax-toolbox-eks-output/axlearn/${{ github.run_id }}/logs/summary.txt axlearn_output/
+        aws s3 cp s3://jax-toolbox-eks-output/axlearn/${{ github.run_id }}/logs/ axlearn_output/ --recursive --exclude "*" --include "*.log"
+        aws s3 cp s3://jax-toolbox-eks-output/axlearn/${{ github.run_id }}/logs/ axlearn_output/ --recursive --exclude "*" --include "*.xml"
```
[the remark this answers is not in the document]
Steboss: ah what a stupid error, fixing this

.github/eks-workflow-files/axlearn/axlearn-job.yml
```diff
@@ -21,12 +21,8 @@ spec:
 
                       LOG_DIR="/output/${RUN_ID}"
                       mkdir -p ${LOG_DIR}
-
-                      # Start MPS daemon
-                      nvidia-cuda-mps-control -d
-                      # Run tests
-                      pytest-xdist.sh 1 6 ${LOG_DIR}/axlearn-unittests.jsonl test-axlearn.sh --directory "." --output ${LOG_DIR} --test-files "/opt/axlearn/axlearn/common/*_test.py" | tee -a ${LOG_DIR}/pytest_stdout.log
-
+                      # test on JAX, make sure 8 devices are visible
```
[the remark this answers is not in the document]
Steboss: Fixed :)

olupton (approved):
I think we can merge this as an improvement to the current situation.

The job is still marked green despite test errors/failures, although they do seem to be captured in the badge. This should be fixed in a follow-up.

PR #1430, opened 2025-05-02
CUDA: use 12.9.1-based 25.06 base container

olupton:
- [x] Update README with accurate date

PR #1481, opened 2025-05-30
Add GKE example

aybchan:
Add `GKE` `MaxText` train ([example run](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/15744603099/job/44379358307)) and `NCCL` test ([example run](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/15744603099/job/44378422712)) workflows with reusable composite action for managing `xpk` job lifecycle (launch, logs streaming, clean up, artifact upload).

Patches on `xpk` address the following identified issues:
- https://github.com/AI-Hypercomputer/xpk/issues/476
- https://github.com/AI-Hypercomputer/xpk/issues/488
- https://github.com/AI-Hypercomputer/xpk/issues/490
- https://github.com/AI-Hypercomputer/xpk/issues/491
- https://github.com/AI-Hypercomputer/xpk/issues/492

Cluster create with `xpk` ([example run](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/15591134618/job/43910254644#step:5:1)) - added as a separate [workflow](https://github.com/NVIDIA/JAX-Toolbox/pull/1481/files#diff-801fc28cafbf1e0fa0ea521355fa8a1c9e6c01dcb8b1083c47f66e2ead4d560a) for demonstration purposes (will not be operational in the CI)

### DIFF
--- a/.github/gke-workflow/nccl/scripts/test.sh
+++ b/.github/gke-workflow/nccl/scripts/test.sh
@@ -48,8 +48,6 @@ run_nccl() {
          -x NCCL_SHIMNET_GUEST_CONFIG_CHECKER_CONFIG_FILE=/usr/local/nvidia/lib64/a3plus_guest_config.textproto \
          -x NCCL_FASTRAK_PLUGIN_ACCEPT_TIMEOUT_MS=${NCCL_FASTRAK_PLUGIN_ACCEPT_TIMEOUT_MS} \
          -x NCCL_NVLS_ENABLE=${NCCL_NVLS_ENABLE} \
-         taskset \
-         -c 32-63 \
          ${NCCL_BENCHMARK} --minbytes ${NCCL_MINBYTES} \
                            --maxbytes ${NCCL_MAXBYTES} \
                            --stepfactor ${NCCL_STEPFACTOR} \
--- a/.github/gke-workflow/nccl/scripts/test.sh
+++ b/.github/gke-workflow/nccl/scripts/test.sh
@@ -1,92 +1,63 @@
 set -x
 
-: "${BENCHMARK:?Must set BENCHMARK}"
-: "${NHOSTS:?Must set NHOSTS}"
+export SCRIPT_DIR=/scripts
 
 ulimit -n 1048576
 
-DATA_MIN="${DATA_MIN:-8G}"
-DATA_MAX="${DATA_MAX:-8G}"
-GPU_PER_NODE="${GPU_PER_NODE:-8}"
-RUN_ITERS="${RUN_ITERS:-20}"
-WARMUP_ITERS="${WARMUP_ITERS:-5}"
+NCCL_LIB_DIR=${NCCL_LIB_DIR} . /usr/local/nvidia/lib64/nccl-env-profile.sh
 
-SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
-
-host_nic="eth0"
-gpu_nics="eth1,eth2,eth3,eth4,eth5,eth6,eth7,eth8"
-find_nics() {
-    local nics_discovered=0
-    local host_nic_discovered=""
-    local gpu_nics_discovered=""
-    for netdev in $(ls /sys/class/net | awk '{print $1}'); do
-        pci_bdf=$(cat "/sys/class/net/${netdev}/device/uevent" 2>/dev/null | grep "PCI_SLOT_NAME" | awk -F '=' '{print $2}' || echo "")
-        if [[ -z "${pci_bdf}" ]]; then
-            continue
-        fi
-        if [[ "${pci_bdf}" =~ 00\:[0-9a-f]{2}\.0 ]]; then
-            host_nic_discovered="${netdev}"
-        elif [[ "" =~ [0|8][6-7d-e]\:00\.0 ]]; then
-            if [[ -z "${gpu_nics_discovered}" ]]; then
-                gpu_nics_discovered="${netdev}"
-            else
-                gpu_nics_discovered="${gpu_nics_discovered},${netdev}"
-            fi
-        fi
-    done
-    if [[ -n "" ]]; then
-        host_nic="${host_nic_discovered}"
-    fi
-    if [[ -n "" ]]; then
-        gpu_nics="${gpu_nics_discovered}"
-    fi
-}
-
-find_nics
+: "${NCCL_BENCHMARK:?Must set NCCL_BENCHMARK}"
+NCCL_MINBYTES="${NCCL_MINBYTES:-8G}"
+NCCL_MAXBYTES="${NCCL_MAXBYTES:-16G}"
+NCCL_STEPFACTOR="${NCCL_STEPFACTOR:-2}"
+NCCL_ITERS="${NCCL_ITERS:-100}"
+NCCL_WARMUP_ITERS="${NCCL_WARMUP_ITERS:-0}"
 
 run_nccl() {
   mpirun --mca btl tcp,self \
          --mca btl_tcp_if_include eth0 \
          --allow-run-as-root \
-         -np $(( GPU_PER_NODE * "${NHOSTS}" )) \
-         --hostfile "${SCRIPT_DIR}/hostfiles${NHOSTS}/hostfile${GPU_PER_NODE}"     \
-         -x NCCL_DEBUG_FILE="/tmp/${BENCHMARK}"-%h-%p.log     \
-         -x NCCL_TOPO_DUMP_FILE="/tmp/${BENCHMARK}"_topo.txt     \
-         -x NCCL_GRAPH_DUMP_FILE="/tmp/${BENCHMARK}"_graph.txt     \
+         -np $(( GPUS_PER_NODE * "${NHOSTS}" )) \
+         --hostfile "${SCRIPT_DIR}/hostfiles${NHOSTS}/hostfile${GPUS_PER_NODE}"     \
          -x LD_LIBRARY_PATH \
          -x PATH     \
          -x NCCL_DEBUG=VERSION \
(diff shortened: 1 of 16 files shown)


.github/gke-workflow/nccl/scripts/test.sh
```diff
@@ -0,0 +1,92 @@
...
+        host_nic="${host_nic_discovered}"
+    fi
+    if [[ -n "" ]]; then
+        gpu_nics="${gpu_nics_discovered}"
+    fi
+}
+
+find_nics
+
+run_nccl() {
```
olupton, line 16: Where did this long list of settings etc. come from? I thought that these GCP-specific NCCL environment settings were provided by GCP in a shell script alongside their NCCL plugin?

.github/workflows/_create_gke_cluster_xpk.yml
olupton, line 1: This is unused, right? Just kept for reference?

.github/gke-workflow/xpk/blueprint.patch
```diff
@@ -0,0 +1,35 @@
...
+@@ -194,20 +193,18 @@ class BlueprintGenerator:
+     a3_megagpu_pool_0 = DeploymentModule(
+         id="a3_megagpu_pool_0",
+         source="modules/compute/gke-node-pool",
+-        use=["gke_cluster", gpu_subnets_name, "group_placement_0"],
++        use=["gke_cluster", gpu_subnets_name],
+         settings={
+             "name": f"{cluster_name}-a3-megagpu-pool-0",
+             "machine_type": system.gce_machine_type,
++            "guest_accelerator": [{"type":"nvidia-h100-mega-80gb", "count": 8, "gpu_driver_installation_config": {"gpu_driver_version": "LATEST"}}],
```
olupton: Is this still blindly taking the latest version from https://docs.nvidia.com/datacenter/tesla/drivers/releases.json ? That is now R575, which is a new feature branch that doesn't support forward compatibility. So I wonder if we will start to see problems next time the cluster is re-created.

.github/gke-workflow/xpk/docker_resources.patch
```diff
@@ -0,0 +1,109 @@
...
+   if system.accelerator_type == AcceleratorType['GPU']:
+     gpu_direct_name = 'fastrak'
+     if args.device_type == H100_DEVICE_TYPE:
+       gpu_direct_name = 'tcpx'
+       gpu_env_yaml += """
++                  - name: NCCL_LIB_DIR
++                    value: /opt/nvidia/nccl/lib
+                   - name: LD_LIBRARY_PATH
+-                    value: /usr/local/nvidia/lib64
++                    value: /opt/nvidia/nccl/lib:/usr/local/cuda-12.8/targets/x86_64-linux/lib:/usr/local/nvidia/lib64
```
olupton: `/opt/nvidia/nccl/lib` I understand is a workaround to make sure the container's NCCL wins out over GCP's one they mount into `/usr/local/nvidia/lib64`. I'm not sure why `/usr/local/cuda-12.8/targets/x86_64-linux/lib` is needed though? Clearly this will rot as the CUDA version inside the containers changes.

.github/gke-workflow/xpk/tcpxo_decorator.patch
```diff
@@ -0,0 +1,42 @@
...
+           {'name': 'sys', 'mountPath': '/hostsysfs'},
+           {'name': 'proc-sys', 'mountPath': '/hostprocsysfs'},
+       ],
+-      'env': [{'name': 'LD_LIBRARY_PATH', 'value': '/usr/local/nvidia/lib64'}],
++      'env': [
++          {'name': 'LD_LIBRARY_PATH', 'value': '/opt/nvidia/nccl/lib:/usr/local/cuda-12.8/targets/x86_64-local/lib:/usr/local/nvidia/lib64'},
++          {'name': 'NCCL_LIB_DIR', 'value': '/opt/nvidia/nccl/lib'},
++          {'name': 'NCCL_TUNER_PLUGIN', 'value': 'none'},
++          {'name': 'NCCL_NET_PLUGIN', 'value': '/opt/hpcx/nccl_rdma_sharp_plugin/lib/libnccl-net.so'}
++      ],
```
olupton: I'm surprised this is needed? `add_tcpxo_daemon_container` looks like this is the environment of the sidecar container that runs the TCPXO daemon, which will presumably not have the `/opt/hpcx` and `/opt/nvidia` paths.

.github/gke-workflow/xpk/tcpxo_decorator.patch
```diff
@@ -0,0 +1,42 @@
...
+   }
+   job_manifest['spec']['template']['spec']['containers'].append(
+       tcpxo_daemon_container
+@@ -175,12 +180,22 @@ def update_gpu_containers(job_manifest):
+     if 'nvidia.com/gpu' in container.get('resources', {}).get('limits', {}):
+       container.setdefault('env', [])
+       container['env'].append(
+-          {'name': 'LD_LIBRARY_PATH', 'value': '/usr/local/nvidia/lib64'}
++          {'name': 'LD_LIBRARY_PATH', 'value': '/opt/nvidia/nccl/lib:/usr/local/cuda-12.8/targets/x86_64-local/lib:/usr/local/nvidia/lib64'}
```
olupton: Same comment as above.

.github/workflows/_create_gke_cluster_xpk.yml
olupton, line 1: This isn't used, right?

.github/gke-workflow/nccl/scripts/test.sh
```diff
@@ -0,0 +1,92 @@
...
+    done
+    if [[ -n "" ]]; then
+        host_nic="${host_nic_discovered}"
+    fi
+    if [[ -n "" ]]; then
+        gpu_nics="${gpu_nics_discovered}"
+    fi
+}
+
+find_nics
```
olupton: Seems like a no-op (`-n ""`)? Also I am not sure if it is safe to assume that the launcher (where `mpirun` is called) is actually on one of the compute nodes - so introspecting here seems iffy.

Generally I do not think we want to be adopting this kind of logic ourselves, and we should prefer to take GCP's configuration of GCP-specific NCCL parameters.

.github/gke-workflow/nccl/scripts/test.sh
```diff
@@ -0,0 +1,92 @@
...
+         -x CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 \
+         -x NCCL_NET_GDR_LEVEL=PIX \
+         -x NCCL_FASTRAK_ENABLE_HOTPATH_LOGGING=0 \
+         -x NCCL_TUNER_PLUGIN=libnccl-tuner.so \
+         -x NCCL_TUNER_CONFIG_PATH=/usr/local/nvidia/lib64/a3plus_tuner_config.textproto \
+         -x NCCL_SHIMNET_GUEST_CONFIG_CHECKER_CONFIG_FILE=/usr/local/nvidia/lib64/a3plus_guest_config.textproto \
+         -x NCCL_FASTRAK_PLUGIN_ACCEPT_TIMEOUT_MS=600000 \
+         -x NCCL_NVLS_ENABLE=0 \
+         taskset \
+         -c 32-63 \
```
olupton: Seems odd - why are we pinning all 8 ranks per node to the same few cores?

.github/gke-workflow/nccl/scripts/test.sh
```diff
@@ -0,0 +1,92 @@
+set -x
+
+: "${BENCHMARK:?Must set BENCHMARK}"
+: "${NHOSTS:?Must set NHOSTS}"
+
+ulimit -n 1048576
+
+DATA_MIN="${DATA_MIN:-8G}"
+DATA_MAX="${DATA_MAX:-8G}"
```
olupton: Make the defaults match the EKS test?

olupton (commented):
Looks good. The MaxText job hasn't run yet in the current pipeline, so I haven't checked the logs there, but posting a few comments.

Generally I'm in favour of merging this and refining later - it seems low-risk of breaking anything else.

.github/gke-workflow/xpk/tcpxo_decorator.patch
aybchan: You're right, this is not necessary on the tcpxo sidecar - has been removed

.github/gke-workflow/nccl/scripts/test.sh
aybchan: Updated this script to source the NCCL environment variables from the Google-provided script

.github/gke-workflow/nccl/scripts/test.sh
aybchan: Updated this script to source the NCCL environment variables from the Google-provided script

.github/gke-workflow/xpk/tcpxo_decorator.patch
aybchan: The patch on update_gpu_containers is necessary as it updates the main (user-provided) container which has these libs. Without this it defaults to using the Google-provided ones

.github/workflows/_create_gke_cluster_xpk.yml
aybchan: Yes, will not be operational in the CI. I wanted to keep a copy of the recipe used to create the cluster for reference

.github/gke-workflow/xpk/blueprint.patch
aybchan: True, updated to use default driver

.github/gke-workflow/nccl/scripts/test.sh
aybchan: Removed

.github/gke-workflow/xpk/docker_resources.patch
olupton: FWIW #1430 just moved us to 12.9.1

PR #1487, opened 2025-06-02
triage tool: manage test case output files

olupton:
The main feature added here is that a special directory, `/triage-tool-output`, is mounted into the container when the test case is executed, and the host-side versions of this directory are tracked by the tool so it can identify those corresponding to the last-known-good and first-known-bad invocations of the test case.

This means that the user can immediately examine "what changed", without having to manually re-run those two configurations.

Other minor changes:
- Add a `Container` interface with `PyxisContainer` and `DockerContainer` implementations, for neater type annotations.
- Add a result cache, which allows the tool to avoid some redundant re-executions
- Write the same log information to `info.log` as to the console
- Only call `git fetch` if needed; if commit-level triage is using a "new"/failing container, all relevant git history should already be available.

### DIFF
--- a/.github/triage/jax_toolbox_triage/logic.py
+++ b/.github/triage/jax_toolbox_triage/logic.py
@@ -420,6 +420,12 @@ def commit_search(
     logger: instance to log output to
     skip_precondition_checks: if True, some tests that should pass/fail by
         construction are skipped
+
+    Returns a 3-tuple of (summary_dict, last_known_good, first_known_bad),
+    where the last element can be None if skip_precondition_checks=True. The
+    last two elements' .result fields will always be, respectively, True and
+    False, but the other fields can be used to obtain stdout+stderr and
+    output files from those test invocations.
     """
     return _commit_search(
         commits=commits,


docs/triage-tool.md
```diff
@@ -292,6 +292,21 @@ but that if JAX is moved forward to include
...
+The tool will mount a host system directory (under `--output-prefix`) into the
+container at `/triage-tool-output` and will make sure that this directory is unique for
+each container + commits that is tested. After the triage has converged, the tool will
+create `first-known-bad` and `last-known-good` symlinks under `--output-prefix` that
+identify the directories corresponding to immediately before and after the problematic
+commit.
+
+This can be useful to save metadata, such as HLO dump files or profile data.
+
+**Important**: if your test case launches multiple processes, it is your responsibility
```
gpupuck, line 305: Could you provide an example of this? Like how user should specify this and how it looks like in resulting dir.

.github/triage/jax_toolbox_triage/args.py
```diff
@@ -4,26 +4,28 @@
...
-compulsory_software = {"xla", "jax"}
-optional_software = {"flax", "maxtext"}
+# Note this is not a `set` for the sake of run-to-run determinism.
+compulsory_software = ["xla", "jax"]
+optional_software = ["flax", "maxtext"]
 
 
-def parse_commit_argument(s):
-    ret = {}
+def parse_commit_argument(s: str) -> typing.Dict[str, str]:
```
gspschmid, line 19: I think you should be able to use `dict[str, str]` / `list[X]` / `tuple[X, Y]` / ... these days, unless you're targeting older Python versions (do we run type checks with `<=3.8`?)

docs/triage-tool.md
olupton: Done

.github/triage/jax_toolbox_triage/logic.py
```diff
@@ -349,18 +359,72 @@ def commit_search(
...
+    Arguments:
+    commits: *ordered* dictionary of commit sequences for different software
+        packages, e.g. commits["jax"][0] is (hash, date) of the passing JAX
+        commit. The ordering of packages has implications for precisely how
+        the triage proceeds.
+    build_and_test: callable that tests if a given vector of commits passes
+    logger: instance to log output to
+    skip_precondition_checks: if True, some tests that should pass/fail by
+        construction are skipped
+    """
```
gspschmid, line 429: Would be useful to describe what's returned as well (ret, last_known_good, first_known_bad).

.github/workflows/ci.yaml
```diff
@@ -11,6 +11,7 @@ on:
       - synchronize
     paths-ignore:
       - '**.md'
+      - '.github/triage/**'
```
gspschmid, line 14: Why do we want to ignore these?

.github/triage/jax_toolbox_triage/main.py
```diff
@@ -67,20 +71,45 @@ def get_commits_and_dirs(
...
+        out_dirname = "-".join(
+            itertools.chain(
+                [urlhash], map(lambda t: f"{t[0]}-{t[1][:hash_chars]}", commits.items())
+            )
+        )
+        out_dir = args.output_prefix / out_dirname
+        assert not out_dir.exists(), (
+            f"{out_dir} should not already exist, maybe you are re-using {args.output_prefix}?"
+        )
+        out_dir.mkdir(mode=0o755)
```
gspschmid, line 99: Not sure if this would end up being useful in practice, but perhaps we should also dump the original container `url` into a textfile in `out_dir`?

.github/triage/jax_toolbox_triage/main.py
```diff
@@ -201,7 +240,48 @@ def check_container(date: datetime.date) -> TestResult:
...
+        result = worker.check_exec(
+            [
+                "git",
+                "log",
+                "--first-parent",
+                "--reverse",
+                "--format=%H %cI",
+                f"{start}^..{end}",
+            ],
+            policy="once",
```
gspschmid, line 269: Why should this one run `"once"` rather than `"once_per_container"` as the other commands above?

gspschmid (approved):
LGTM, minor comments

.github/triage/jax_toolbox_triage/args.py
olupton: Unfortunately `dict[str,str]` is a syntax error in python 3.8, so it's a problem even for running pytest with 3.8. We only actually run mypy with 3.12.

.github/workflows/ci.yaml
olupton: It's a heavy pipeline that doesn't touch these files.

.github/triage/jax_toolbox_triage/main.py
olupton: There's already a `summary.json` in the output prefix, which has a `"commits"` (because this is the commit-level search) list with elements that look like:
```
    {
      "build_time": 73.5948702134192,
      "container": "...",
      "output_directory": "/path/to/container-<hash>-xla-<hash>-jax-<hash>-flax-<hash>",
      "result": false,
      "test_time": 130.4865097105503,
      "xla": "<hash>",
      "jax": "<hash>",
      "flax": "<hash>"
    }
```
so it's possible to recover. Actually putting it inside the output directory that is written to by the container seems slightly less appealing, as that would mix together files written by the tool and by the test case.

.github/triage/jax_toolbox_triage/main.py
olupton: `git fetch` is being run for its side effects (making all N copies of the git checkout in N containers are up-to-date, so it needs to be `"once_per_container"`.

If `git log` was `"once_per_container"` we'd read N copies of the git history and would have to disentangle them, and we are not relying on any side effects; `"once"` is neater.

`git cat-file` could be `"once"` I think, but it's probably harmless.

.github/triage/jax_toolbox_triage/logic.py
olupton: Done

PR #1501, opened 2025-06-10
Add local runtime support for in-situ unit test bisection

camille-004:
This PR introduces a new `local` mode to the triage utility that allows developers to perform a commit-level bisection for unit tests without requiring container orchestration. This is designed for the case where a user is already inside a failing container and wants to find a culprit commit directly.

**Changes:**
1. Adds a new option, **--container-runtime=local**.
2. When this mode is active, all `git` operations and build/test commands are executed directly on the local machine using `subprocess`.
3. The `local` runtime skips the container search, so it requires the user to provide the bisection range explicitly using:
  - `--passing-commits="jax:<hash>,xla:<hash>"`
  - `--failing-commits="jax:<hash>,xla:<hash>"`
4. The argument parser has been updated to enforce these new rules and prevent mixing `local` mode with container-specific arguments (i.e, `--container`, `--start-date`).
5. Unit tests to the new argument parsing logic

Now, this command can be run inside a JAX development container:
```
jax-toolbox-triage \
    --container-runtime=local \
    --passing-commits="jax:<some_good_hash>,xla:<some_good_hash>" \
    --failing-commits="jax:<some_bad_hash>,xla:<some_bad_hash>" \
    /opt/jax-toolbox/.github/container/test-jax.sh "some_target"
```

### DIFF
--- a/.github/triage/jax_toolbox_triage/args.py
+++ b/.github/triage/jax_toolbox_triage/args.py
@@ -187,11 +187,25 @@ def parse_args(args=None) -> argparse.Namespace:
     parser.add_argument(
         "--container-runtime",
         default="docker",
-        help="Container runtime used, this can be either docker or pyxis.",
+        help="Container runtime used, can be docker, pyxis, or local.",
         type=lambda s: s.lower(),
     )
     args = parser.parse_args(args=args)
-    assert args.container_runtime in {"docker", "pyxis"}, args.container_runtime
+    assert args.container_runtime in {"docker", "pyxis", "local"}, args.container_runtime
+    
+    if args.container_runtime == "local":
+        assert (
+            args.passing_commits is not None and args.failing_commits is not None
+        ), "For local runtime, --passing-commits and --failing-commits must be provided."
+        assert (
+            args.container is None 
+            and args.start_date is None
+            and args.end_date is None
+            and args.passing_container is None
+            and args.failing_container is None
+        ), "Container-level search options are not applicable for local runtime."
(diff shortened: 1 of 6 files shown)


.github/triage/jax_toolbox_triage/args.py
```diff
@@ -187,11 +187,25 @@ def parse_args(args=None) -> argparse.Namespace:
...
+    if args.container_runtime == "local":
+        assert (
+            args.passing_commits is not None and args.failing_commits is not None
+        ), "For local runtime, --passing-commits and --failing-commits must be provided."
+        assert (
+            args.container is None 
+            and args.start_date is None
+            and args.end_date is None
+            and args.passing_container is None
+            and args.failing_container is None
```
olupton, line 207: `--container-mount` is also not going to do anything, plus `--container-url-template`

.github/triage/jax_toolbox_triage/local.py
```diff
@@ -0,0 +1,37 @@
...
+    def __exit__(self, *exc_info) -> None:
+        pass
+
+    def __repr__(self) -> str:
+        return "Local"
+
+    def exec(
+        self,
+        command: typing.List[str],
+        policy: typing.Literal["once"]
```
olupton: Python 3.8 doesn't like these type hints - the style used in the other backends should be fine.

olupton (approved):
LGTM, thanks! Some tiny nits from me and the CI.

PR #1505, opened 2025-06-12
triage tool: installXXX.sh-based triage

olupton:
This extends support to basic triage of other versioned components inside the container via external installation scripts. For example, if the cuBLAS version (as encoded by the `CUBLAS_VERSION` environment variable) changes from the last-known-good to the first-known-bad container, the tool can include those two versions in the triage algorithm and call an external `installCUBLAS.sh` (`--build-scripts-path`) script to change which one is active.

`--build-scripts-path` is a container *inside* the test environment/container, so it will likely be combined with `--container-mount` to allow the installation scripts to be maintained outside.

Also print a comparison of the environment variables in the last-known-good and first-known-bad containers if those are both available.

### DIFF
--- a/.github/triage/jax_toolbox_triage/logic.py
+++ b/.github/triage/jax_toolbox_triage/logic.py
@@ -318,7 +318,7 @@ def _version_search(
                     break
             indices[secondary] = index
             bisect_versions[secondary] = version
-            if len(version_list):
+            if len(version_list) > 1:
                 log_msg += f", {len(version_list)} remaining {secondary} versions"
                 log_msg += (
                     f" [{versions[secondary][0][0]}, {versions[secondary][-1][0]}]"
--- a/.github/triage/jax_toolbox_triage/main.py
+++ b/.github/triage/jax_toolbox_triage/main.py
@@ -419,7 +419,9 @@ def main() -> None:
             assert passing_versions[package] == package_versions[package][0][0]
             assert failing_versions[package] == package_versions[package][-1][0]
         # For the packages that just have one or two version numbers, associate those
-        # version numbers with the earliest and, if appropriate, latest XLA dates.
+        # version numbers with the earliest and, if appropriate, latest XLA dates. This
+        # is only relevant to packages that do not have git repositories checked out and
+        # are managed via installPACKAGE.sh scripts and PACKAGE_VERSION environment vars
         for package in packages:
             if package in package_versions:
                 continue
@@ -458,9 +460,10 @@ def main() -> None:
             packages_missing_scripts = packages_needing_scripts - packages_with_scripts
             if packages_missing_scripts:
                 logger.warning(
-                    f"No installation scripts found for: {packages_missing_scripts}, "
-                    "whose versions change across the bisection range. These will be "
-                    "excluded from the bisection, which may cause it not to converge!"
+                    "No installation scripts found for: "
+                    f"{' '.join(packages_missing_scripts)}, whose version(s) change "
+                    "across the bisection range. These will be excluded from the "
+                    "bisection, which may cause it not to converge!"
                 )
                 dynamic_packages -= packages_missing_scripts
 
--- a/docs/triage-tool.md
+++ b/docs/triage-tool.md
@@ -313,6 +313,11 @@ Only two versions of `COMPONENT` will be used, one from each end of the triage r
 If those versions are the same as each other, `installCOMPONENT.sh` will not be called
 and need not exist.
 
+Example: `--passing-container=contains_COMPONENT_v1` combined with
+`--failing-container=contains_COMPONENT_v3` might blame `v3` of `COMPONENT` for the
+regression, in which case `--passing-versions=COMPONENT:v2` could be used to re-run and
+see if the issue in fact appeared in `v2`.
+
 ## Other features
 The tool will mount a host system directory (under `--output-prefix`) into the
 container at `/triage-tool-output` and will make sure that this directory is unique for


docs/triage-tool.md
```diff
@@ -291,6 +291,24 @@ but that if JAX is moved forward to include
...
+triage algorithm.
+To enable this feature, `--build-scripts-path` must give a path **inside** the test
+environment (container) that contains executables named `installCOMPONENT.sh` that
+accept a version as the first argument.
+This is typically used in conjunction with `--container-mount` to allow those
+installation helpers to be injected from the host system.
+
+Versions of `COMPONENT` will be read from the last-known-good and first-known-bad
+container environment variable `COMPONENT_VERSION`, or the version of `COMPONENT` can
+be set explicitly via `--{passing,failing}-commits`.
```
nouiz: "or the version of `COMPONENT` can be set explicitly via `--{passing,failing}-commits`."

I don't understand that part.

docs/triage-tool.md
```diff
@@ -291,6 +291,24 @@ but that if JAX is moved forward to include
...
+
+Only two versions of `COMPONENT` will be read, one from each end of the triage range.
+If those versions are the same as each other, `installCOMPONENT.sh` will not be called
+and need not exist.
```
nouiz, line 314: What will happen if a component changes, but the script don't exist? Is it useful to mention it?

.github/triage/jax_toolbox_triage/logic.py
```diff
@@ -73,10 +71,21 @@ def adjust_date(
     return None
 
 
+class DateTester(typing.Protocol):
+    def __call__(
+        self, date: datetime.date, *, test_output_log_level: int = logging.DEBUG
+    ) -> TestResult:
+        """
+        Given an [unordered] set of {package_name: package_commit_sha}, build
```
gspschmid: This comment seems at odds with the signature -- i.e. we're given a single `date`.

gspschmid: (Copy-pasted from below?)

docs/triage-tool.md
gspschmid: Maybe add an example of what that looks like. I guess the confusion here comes from the mixing the notion of commits (for jax/xla/maxtext/flax these are indeed commit SHAs) and "versions" which may be arbitrary strings (for other software components).

docs/triage-tool.md
gspschmid: I guess the assumption here is that new components are only tracked once the installation script is available. Before that we'll be implicitly using what's installed in the container.

.github/triage/jax_toolbox_triage/main.py
```diff
@@ -23,59 +23,108 @@
...
+            map(
+                lambda kv: kv.split("=", 1),
+                worker.check_exec(
+                    ["env", "-0"],
+                    policy="once",
+                    stderr="separate",
+                )
+                .stdout[:-1]  # skip the trailing \0
+                .split("\0"),
+            )
```
gspschmid: Nothing wrong with this, but in Python I'd likely just use a comprehension instead of the map:
```python
def impl() -> typing.Dict[str, str]:
    kvs = worker.check_exec(["env", "-0"], policy="once", stderr="separate")
        .stdout[:-1]  # skip the trailing \0
        .split("\0")
    return dict(kv.split("=", 1) for kv in kvs)
```

.github/triage/jax_toolbox_triage/main.py
```diff
@@ -119,31 +169,48 @@ def Container(
...
+        If *only* overrides are given, those are returned verbatim and all other return
+        values are None.
 
-        If both arguments are non-None, the commits read from inside the given
-        container will be overwritten by explicit passed entries.
+        Otherwise, the software versions, git repository directories and runtime
+        environment are extracted from the given container. If overrides are *also* given,
+        these take precedence over the extracted values.
 
         If is an error for both of the arguments to be None.
```
gspschmid: ... for both `container_url` and `explicit_versions` to be None.

+ 

versions_from_env: Determines whether to extract additional versions from the environment.

.github/triage/jax_toolbox_triage/main.py
```diff
@@ -23,59 +23,108 @@
...
+    """
+    # Get the git repository paths and commits from the container.
+    versions, dirs = get_commits_and_dirs(worker)
+
+    # Get the environment variables from the container.
+    env = get_env(worker)
+
+    if versions_from_env:
+        # Promote any XXX_VERSION environment variables into `versions` if `XXX` is
+        # not already there.
```
gspschmid, line 102: Should there be a whitelist restricting which COMPONENTs we expect?

Iiuc including COMPONENT in the commit-level search will in general only work if we also have a `installCOMPONENT.sh`, so these should probably be opted into in some way. Perhaps `installCOMPONENT.sh` _is_ the way to opt into it?

.github/triage/jax_toolbox_triage/main.py
```diff
@@ -23,59 +23,108 @@
...
+        # TODO: the match might need to be fuzzier (case, hyphens vs underscores)
+        for k, v in env.items():
+            if not len(v):
+                continue
+            if k[-8:] != "_VERSION":
```
gspschmid: Nit: `k.endswith("_VERSION")`

.github/triage/jax_toolbox_triage/main.py
```diff
@@ -23,59 +23,108 @@
...
+        ]
+        if package in optional_software:
+            bits.append("true")
+        cmds.append(f"({' || '.join(bits)})")
+    result = worker.check_exec(
+        ["sh", "-c", " && ".join(cmds)], policy="once", stderr="separate"
+    )
+    versions, dirs = {}, {}
+    # Look over triplets of output lines
+    for commit, package, dirname in zip(*([iter(result.stdout.splitlines())] * 3)):
```
gspschmid, line 72: While I think there's a certain beauty to it, please consider using `itertools.batched(result.stdout.splitlines(), 3, strict=True)` instead ;-)

.github/triage/jax_toolbox_triage/main.py
```diff
@@ -225,34 +296,62 @@ def check_container(date: datetime.date) -> TestResult:
         passing_url = args.passing_container
         failing_url = args.failing_container
 
-    # Get the git commits in the endpoint containers, or the explicitly passed commits
-    passing_commits, passing_package_dirs = get_commits(
-        passing_url, args.passing_commits
+    # Get the commits/versions in the endpoint containers, or the explicitly passed
+    # commits, or the former overriden by the latter.
```
gspschmid: Is this different from saying "Get the commits/versions in the endpoint containers, overridden by any explicitly passed commits/versions"?

gspschmid: Also, maybe standardize language on
* "commits" -> "internal" (repo available) components
* "versions" -> internal+external (opaque install script available) components
so we'd only talk about versions here.

.github/triage/jax_toolbox_triage/main.py
```diff
@@ -320,27 +423,63 @@ def get_commit_history(worker, start, end, dir):
                 for a, b in zip(package_commits[package], package_commits[package][1:])
             )
             # Confirm the end values are included as expected
-            assert passing_commits[package] == package_commits[package][0][0]
-            assert failing_commits[package] == package_commits[package][-1][0]
+            assert passing_versions[package] == package_commits[package][0][0]
+            assert failing_versions[package] == package_commits[package][-1][0]
+        # For the packages that just have one or two version numbers, associate those
+        # version numbers with the earliest and, if appropriate, latest XLA dates.
```
gspschmid: I would call out here that this is only relevant for "external" components.

.github/triage/jax_toolbox_triage/main.py
```diff
@@ -320,27 +423,63 @@ def get_commit_history(worker, start, end, dir):
...
+                    "git stash",
+                    f"git checkout {commit}",
+                ]
+            else:
+                # Another software package, `commit` is probably a version number.
+                # Installation of this version is delegated to an installPACKAGE.sh
+                # script that is assumed to be available in `args.build_scripts_path`.
+                assert args.build_scripts_path is not None
+                git_commands += [
+                    f"DEVEL=1 {args.build_scripts_path}/install{package}.sh {commit}"
```
gspschmid: What's the significance of `DEVEL=1`?

gspschmid (changes requested):

.github/triage/jax_toolbox_triage/main.py
olupton: Added in 3.12 :(

.github/triage/jax_toolbox_triage/main.py
olupton: 1. I don't think that quite captures that endpoint container can be missing (i.e. you can pass a good container + a list of bad commits, without a bad container), but I can take your text with an extra "(if they exist)"
2. I agree this has become a mess and mostly commit should be renamed to version - i.e. 👍 to the second bullet. I think instead of keeping "commit" around with the meaning you give, I might just talk about versions of components that may (your internal) or may not (your external) be taken from git repository checkouts.

.github/triage/jax_toolbox_triage/main.py
olupton: It's a bit of detail of how to get our internal `installXXX.sh` scripts to do the right thing (in this context) that I couldn't keep out :/ I am now aware of a case where we need to set another variable in a `package`-dependent way though, so I might just admit defeat on making this quite so generic and be a bit more explicit.

docs/triage-tool.md
olupton: The meaning is that for, say, cuBLAS the version used for the bad/failing endpoint of the triage is the value of the `CUBLAS_VERSION` environment variable inside `--failing-container` (if it exists), but that you can override that value by passing `--failing-container=CUBLAS:x.y.z`.

It should also be possible to do something like
```
--passing-container=... # implicitly has CUBLAS_VERSION=1
--failing-commits=jax:foo,xla:bar,cublas:2
```
but admittedly there is no test coverage of this. (and commits might get renamed)

.github/triage/jax_toolbox_triage/logic.py
olupton: Yep, will fix.

docs/triage-tool.md
olupton: It was attempt to avoid enumerating the installation scripts and keep this implicit, but I should probably just explicitly list the available scripts at both endpoints and use that for an explicit check.

docs/triage-tool.md
olupton: Reworded
```
Versions of `COMPONENT` will be read from the last-known-good and first-known-bad
container environment variable `COMPONENT_VERSION`. These values can be overriden using
`--{passing,failing}-versions`, and **must** be set explicitly if only a single
container is used (*e.g.* if `--passing-container` and `--failing-versions` are used
without `--failing-container`).
```

docs/triage-tool.md
olupton: This is now explicit:
```
                logger.warning(
                    f"No installation scripts found for: {packages_missing_scripts}, "
                    "whose versions change across the bisection range. These will be "
                    "excluded from the bisection, which may cause it not to converge!"
                )
```

.github/triage/jax_toolbox_triage/main.py
olupton: Changed

.github/triage/jax_toolbox_triage/main.py
olupton: Now it:
- gets the list of installation scripts that do exist
- gets the list of packages that need installation scripts (i.e. versions change across the bisection range)
- warns if the latter is not a subset of the former, and excludes those packages from the bisection (which might make it fail to converge, but might be fine -- the results should be correct **if** it converges)

.github/triage/jax_toolbox_triage/main.py
olupton: Did a lot of renaming...

docs/triage-tool.md
olupton: I also added an example:
> Example: `--passing-container=contains_COMPONENT_v1` combined with
`--failing-container=contains_COMPONENT_v3` might blame `v3` of `COMPONENT` for the
regression, in which case `--passing-versions=COMPONENT:v2` could be used to re-run and
see if the issue in fact appeared in `v2`.

.github/triage/jax_toolbox_triage/main.py
olupton: Done

.github/triage/jax_toolbox_triage/main.py
gspschmid: 🙏

PR #1527, opened 2025-06-26
Update `jax` build `bazel` options (`25.07-devel`)

aybchan:


aybchan:
Not necessary according to @DwarKapex

PR #1536, opened 2025-07-02
nsys-jax: re-map module IDs

olupton:
This adds support for multi-process profiling when the different processes do not agree on the global numbering of modules.


.github/container/nsys_jax/nsys_jax/data_loaders.py
```diff
@@ -440,33 +445,43 @@ def _load_nvtx_gpu_proj_trace(
...
         assert not meta_path.is_dir()
         filenames = [path]
+        replica_slugs = [None]
         meta_filenames = [meta_path]
 
     if len(filenames) > 1:
         tmp = defaultdict(list)
         with multiprocessing.Pool(processes=_enough_processes(len(filenames))) as pool:
-            for single_trace in pool.starmap(
+            for single_trace_tup in pool.starmap(
```
gspschmid: `for single_trace, hlo_cache in pool.starmap(...):`?

.github/container/nsys_jax/nsys_jax/data_loaders.py
```diff
@@ -440,33 +445,43 @@ def _load_nvtx_gpu_proj_trace(
...
                     meta_filenames,
                     itertools.repeat(frames),
                 ),
             ):
+                single_trace, hlo_cache = single_trace_tup
                 for k, v in single_trace.items():
                     tmp[k].append(v)
+                # Merge the caches from the pool worker processes into the main one.
+                for k, v in hlo_cache.items():
+                    _hlo_cache[k] |= v
```
gspschmid: Where do you actually hit this cache again in the main process?

.github/container/nsys_jax/nsys_jax/protobuf.py
```diff
@@ -237,6 +252,83 @@ def unique_result(self, callable):
...
+        return "unknown"
+    old_id = int(old_id_str)
+
+    dump_dir = prefix / "dump"
+    candidates = [
+        candidate
+        for candidate in dump_dir.glob("*after_optimizations.hlo.pb.xz")
+        if old_id
+        == int(candidate.name.split(".", maxsplit=1)[0].split("_", maxsplit=1)[1])
+        and candidate.name.split(".", maxsplit=2)[1] == name
```
gspschmid: Maybe add something like
```python
RE_MODULE = re.compile(rf"^module_(\d+)\.jit_(.+)\.(.+)\.hlo\.pb\.xz$")

def match_module(name) -> tuple[int, str, str] | None:
    if m := RE_MODULE.match(name):
        return int(m.group(1)), m.group(2), m.group(3)
```
and use it here via
```python
    ...
    if match_module(candidate.name) == (old_id, name, "after_optimizations")
```
and similarly for the test helpers (`get_module_id` and `module_exists`) below?

.github/container/nsys_jax/tests/nsys_jax_test_helpers/__init__.py
```diff
@@ -29,23 +30,49 @@ def nsys_jax(command):
...
 
 
-def nsys_jax_archive(command):
-    """
-    Helper to run nsys-jax and automatically extract the output, yielding a directory.
-    """
-    tmpdir = tempfile.TemporaryDirectory()
-    archive = nsys_jax(command)
+def extract(archive):
+    tmpdir = tempfile.TemporaryDirectory(delete=False)
```
gspschmid: Do you actually need`delete=False` (since the directory should remain alive as long as `tmpdir`)?

.github/container/nsys_jax/tests/test_multi_process_program.py
```diff
@@ -0,0 +1,279 @@
...
+    assert rank0_spmd1 == rank1_spmd1, (rank0_spmd1, rank1_spmd1)
+    # This should only be executed in the 0th process
+    rank0_only1 = get_module_id(rank0, "only_in_process_zero")
+    with pytest.raises(Exception, match="Did not find only_in_process_zero"):
+        get_module_id(rank1, "only_in_process_zero")
+    # This should be executed in both processes, but the IDs should have diverged
+    rank0_spmd2 = get_module_id(rank0, "another_distinctively_named_function")
+    rank1_spmd2 = get_module_id(rank1, "another_distinctively_named_function")
+    assert rank0_spmd2 > rank1_spmd2
+    assert rank1_spmd2 == rank0_only1  # might turn out to be too flaky?
```
gspschmid, line 108: Seems okay to give a try

gspschmid (approved):
LGTM, minor comments

.github/container/nsys_jax/nsys_jax/data_loaders.py
olupton: Done - not sure what the origin story of that one was...

.github/container/nsys_jax/nsys_jax/data_loaders.py
olupton: `xla_module_metadata` in protobuf.py - that's where we need to go from a re-mapped ProgramId (= HLO fingerprint) back to the list of dumped metadata files on disk.

.github/container/nsys_jax/nsys_jax/protobuf.py
olupton: Done something similar, thanks (after_optimizations is only the suffix, so we can't use an exact match there).

.github/container/nsys_jax/tests/nsys_jax_test_helpers/__init__.py
olupton: Right, this was a leftover from debugging (when it's useful to have the directories outlive the test process)

gspschmid (approved):

PR #1538, opened 2025-07-03
Triage tool deals with non-linear history

Steboss:
If commits do not lie on the same linear git history. The tool would use `git fetch` because it couldn't find a direct path between the passing and failing commits. 
Here the changes introduce a new logic to find the correct bisection range: 
- apply `git merge-base` to identify common ancestors between the passing and failing commit
- at build-and-run stage we're checking out the base commit from the main branch, and we apply cherry-pick for all new relevant features


.github/triage/jax_toolbox_triage/versions.py
```diff
@@ -0,0 +1,89 @@
+import typing
+from .container import Container
+from .args import compulsory_software, optional_software
+
+
+def get_env(worker: Container) -> typing.Dict[str, str]:
```
Steboss, line 6: [Original function from main](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/triage/jax_toolbox_triage/main.py#L26-L45)

No changes applied

.github/triage/jax_toolbox_triage/versions.py
```diff
@@ -0,0 +1,89 @@
...
+        return dict(kv.split("=", 1) for kv in kvs)
+
+    # Remove any environment variables that differ between consecutive `env` calls, for
+    # example some step-specific Slurm variables.
+    env1, env2 = impl(), impl()
+    # sorted(...) for run-to-run determinism
+    return {k: env1[k] for k in sorted(env1.keys() & env2.keys()) if env1[k] == env2[k]}
+
+
+def get_commits_and_dirs(
```
Steboss, line 28: [Original function from main](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/triage/jax_toolbox_triage/main.py#L48-L75) 

No changes applied

.github/triage/jax_toolbox_triage/versions.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+    versions, dirs = {}, {}
+    # Look over triplets of output lines
+    for commit, package, dirname in zip(*([iter(result.stdout.splitlines())] * 3)):
+        dirs[package] = dirname
+        versions[package] = commit
+    return versions, dirs
+
+
+def get_versions_dirs_env(
```
Steboss, line 58: [Original function from main](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/triage/jax_toolbox_triage/main.py#L78-L109)

No changes applied

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,588 @@
...
+        )
+        result_str = "pass" if test_result.returncode == 0 else "fail"
+        self.logger.info(f"Test completed in {test_time:.1f}s ({result_str})")
+        return TestResult(
+            host_output_directory=out_dir,
+            result=test_result.returncode == 0,
+            stdouterr=test_result.stdout,
+        )
+
+    def prepare(self):
```
Steboss: As [before](https://github.com/NVIDIA/JAX-Toolbox/blob/efb11b79004e74a2a889c955d451861af2ad5425/.github/triage/jax_toolbox_triage/main.py#L114) the first step is to prepare the triage, so here we're loading the bazel cache

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,588 @@
...
+        self.bazel_cache_mounts = []
+        self.bisection_url = None
+        self.bisection_versions = {}
+        self.package_dirs = None
+        self.dynamic_packages = set()
+        self.packages_with_scripts = set()
+        # the cherry-pick gets populated only for non-linear cases
+        self.args.cherry_pick_commits = {}
+
+    def _test_output_directory(
```
Steboss, line 39: [Original function from main](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/triage/jax_toolbox_triage/main.py#L124-L147)

No changes applied to the function. Now the function is a private method of the triage tool.

.github/triage/jax_toolbox_triage/container_factory.py
```diff
@@ -0,0 +1,28 @@
+import logging
+from .container import Container
+from .docker import DockerContainer
+from .pyxis import PyxisContainer
+from .local import LocalContainer
+
+
+def make_container(
```
Steboss, line 8: [Original function from main](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/triage/jax_toolbox_triage/main.py#L150-L161) 

The mount is still handled as before in the overall code

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,588 @@
...
+        )
+
+        out_dir = self.args.output_prefix / out_dirname
+        assert not out_dir.exists(), (
+            f"{out_dir} should not already exist, maybe you are re-using {self.args.output_prefix}?"
+        )
+        out_dir.mkdir(mode=0o755)
+        return out_dir.resolve()
+
+    def _get_versions(
```
Steboss, line 84: [Original function from main](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/triage/jax_toolbox_triage/main.py#L163-L204) 

No changes applied, we're now using `with make_container` in place of ` with Container(container_url) as worker:`

.github/triage/jax_toolbox_triage/summary.py
```diff
@@ -0,0 +1,78 @@
+import json
+import logging
+import pathlib
+import typing
+from .logic import TestResult
+
+
+def add_summary_record(
```
Steboss, line 8: [Original function from main](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/triage/jax_toolbox_triage/main.py#L206-L230)

No changes

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,588 @@
...
+            self.logger.info(f"\tOnly in {url1}: {key}={env1[key]}")
+        for key in env2.keys() - env1.keys():
+            self.logger.info(f"\tOnly in {url2}: {key}={env2[key]}")
+        for key in env1.keys() & env2.keys():
+            if env1[key] != env2[key]:
+                self.logger.info(
+                    f"{key}: {env1[key]} ({url1}) vs. {env2[key]} ({url2})"
+                )
+
+    def _check_container_by_date(
```
Steboss, line 208: This function was originally `check_container` (here)[https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/triage/jax_toolbox_triage/main.py#L234-L265] 

no changes applied to the logic

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,588 @@
...
+        )
+
+    def prepare(self):
+        """
+        Function to prepare the triage tool for execution.
+        At the moment, we're adding the bazel cache mounts to the tool.
+        """
+        self.bazel_cache_mounts = prepare_bazel_cache_mounts(self.args.bazel_cache)
+
+    def find_container_range(self) -> Tuple[str, str]:
```
Steboss, line 436: This function re-implements [the following lines for searching dates in containers from main](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/triage/jax_toolbox_triage/main.py#L267-L288), so that we have a function

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,588 @@
...
+                logger=self.logger,
+                skip_precondition_checks=self.args.skip_precondition_checks,
+                threshold_days=self.args.threshold_days,
+            )
+
+            return container_url_func(range_start), container_url_func(range_end)
+        else:
+            return self.args.passing_container, self.args.failing_container
+
+    def gather_version_info(self, passing_url: str, failing_url: str):
```
Steboss, line 471: This function re-implements the [version search](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/triage/jax_toolbox_triage/main.py#L302-L345) 
no changes

.github/triage/jax_toolbox_triage/bisect.py
```diff
@@ -0,0 +1,109 @@
+import datetime
+import subprocess
+
+
+def get_commit_history(
```
Steboss, line 5: [Original function from main](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/triage/jax_toolbox_triage/main.py#L348-L385) 

Here I've added the following: 
1. 
```python 
if commits_known.returncode != 0:
        if worker.exec(["git", "remote"]).stout.strip():
            worker.check_exec(
                ["git", "fetch"], policy="once_per_container", workdir=dir
            )
        else:
            logger.warning("No remote found, skipping fetch.")
```
This is mainly for testing purposes, so we can create local repos and test directly on those. 


2. 
```python 
 # detect non-linear history
    is_ancestor_cmd = f"git merge-base --is-ancestor {start} {end}"
    is_ancestor_result = worker.exec(
        ["sh", "-c", is_ancestor_cmd],
        workdir=dir,
    )
    is_linear = is_ancestor_result.returncode == 0
```
For detecting possible common ancestors in case on non-linear history 

3. The non-linear history works only for `jax` and `xla` packages

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,588 @@
...
+        if packages_missing_scripts:
+            self.logger.warning(
+                "No installation scripts found for: "
+                f"{' '.join(packages_missing_scripts)}, whose version(s) change "
+                "across the bisection range. These will be excluded from the "
+                "bisection, which may cause it not to converge!"
+            )
+            self.dynamic_packages -= packages_missing_scripts
+
+    def _build_and_test(
```
Steboss, line 313: The `build_and_test` is [as the same as it was before](https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/triage/jax_toolbox_triage/main.py#L470-L563)

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,588 @@
...
+            assert passing_url is not None
+            self.bisection_url = passing_url
+            self.bisection_versions = original_passing_versions
+            self.package_dirs = passing_package_dirs
+        assert self.package_dirs is not None
+        # This is the set of versions that are already installed
+        assert self.bisection_versions is not None
+        return passing_versions, failing_versions
+
+    def run_version_bisection(
```
Steboss, line 537: This is [final part of the main script](http://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/triage/jax_toolbox_triage/main.py#L569-L591) + I've moved `symlink` to `summary.py` as `create_output_symlink`

Steboss:
I've also added to the doc the following scheme: 
```
+--------------------------------+
|           main.py              |
|             main()             |
+--------------------------------+
                 |
                 v
+--------------------------------+
|    tool = TriageTool(args)     |
|         tool.prepare()         |
+--------------------------------+
                 |
                 v
+--------------------------------+
| tool.find_container_range()    |
+--------------------------------+
                 |
                 v
+--------------------------------+
| tool.gather_version_info()     |
| # extract commit hashes,       |
| # compare environments         |
+--------------------------------+
                 |
                 v
+--------------------------------+
| tool.run_version_bisection()   |
| # main part of the logic       |
| # it runs all the steps below  |
+--------------------------------+
                 |
                 v
+------------------------------------------------+
|           tool._gather_histories()             |
| (Calls bisect.py -> get_commit_history)        |
| # get the list of all the commits between p&f  |
+------------------------------------------------+
                 |
                 | <--- Inside get_commit_history()
                 |
+----------------------------------------------------------------------------+
|               Is history linear? (git merge-base --is-ancestor)            | 
+--------------------------+-------------------------------------------------+
|           YES            |                       NO                        |
|   (Linear History Path)  |            (Non-Linear History Path)            |
+--------------------------+-------------------------------------------------+
                 |                                |
                 v                                v
+--------------------------+     +-------------------------------------------------+
| get_commit_history()     |     | get_commit_history()                            |
| returns commits from     |     | - Uses 'git merge-base' to find linear range    |
| main branch directly.    |     | - Finds cherry-picks to apply to args           |
|                          |     | - Returns commits from the *main* branch        |
+--------------------------+     +-------------------------------------------------+
                 |                                |
                 |                                |
                 +--------------------------------+
                                 |
                                 v
+------------------------------------------------+
|      logic.py -> version_search() loop         |
| (Repeatedly calls tool._build_and_test)        |
+------------------------------------------------+
                                 |
                                 | <--- Inside _build_and_test()
                                 |
+----------------------------------------------------------------------------+
|                Does args.cherry_pick_commits exist? -->                    |
+--------------------------+-------------------------------------------------+
|           NO             |                      YES                        |
|   (Linear History Path)  |            (Non-Linear History Path)            |
+--------------------------+-------------------------------------------------+
                 |                                |
                 v                                v
+--------------------------+     +-------------------------------------------------+
| _build_and_test()        |     | _build_and_test()                               |
| - git checkout <commit>  |     | - git checkout <main_branch_commit>             |
| - build & test           |     | - git cherry-pick <feature_commits>             |
|                          |     | - build & test                                  |
+--------------------------+     +-------------------------------------------------+
```
to simplify the logic in the code

.github/triage/jax_toolbox_triage/bisect.py
```diff
@@ -0,0 +1,109 @@
...
+            )
+        else:
+            logger.warning("No remote found, skipping fetch.")
+
+    # detect non-linear history
+    is_ancestor_cmd = f"git merge-base --is-ancestor {start} {end}"
+    is_ancestor_result = worker.exec(
+        ["sh", "-c", is_ancestor_cmd],
+        workdir=dir,
+    )
```
olupton:
```suggestion
    is_ancestor_result = worker.exec(
        ["git", "merge-base", "--is-ancestor", start, end],
        workdir=dir,
    )
```

.github/triage/jax_toolbox_triage/bisect.py
```diff
@@ -0,0 +1,109 @@
...
+        [
+            "sh",
+            "-c",
+            f"git cat-file commit {start} && git cat-file commit {end}",
+        ],
+        policy="once_per_container",
+        workdir=dir,
+    )
+    if commits_known.returncode != 0:
+        if worker.exec(["git", "remote"]).stout.strip():
```
olupton: `check_exec`? But I am not sure I understood why this is needed? How do you end up with a local remote-less repository that doesn't know about all of the relevant commits?

.github/triage/jax_toolbox_triage/bisect.py
```diff
@@ -0,0 +1,109 @@
...
+            worker.check_exec(
+                ["git", "fetch"], policy="once_per_container", workdir=dir
```
olupton: I think that to actually work for the case we're targeting, this needs to explicitly list `start` and `end` - they are likely to be dangling commits from the perspective of the remote, and are therefore will not be fetched by a simple `git fetch`.

I don't know how to avoid specifying the remote name explicitly, but we need something like `["git", "fetch", "origin", start, end]`.

.github/triage/jax_toolbox_triage/bisect.py
```diff
@@ -0,0 +1,109 @@
...
+
+    # detect non-linear history
+    is_ancestor_cmd = f"git merge-base --is-ancestor {start} {end}"
+    is_ancestor_result = worker.exec(
+        ["sh", "-c", is_ancestor_cmd],
+        workdir=dir,
+    )
+    is_linear = is_ancestor_result.returncode == 0
+
+    if not is_linear and package in ["jax", "xla"]:
```
olupton: Any particular reason to list out JAX and XLA explicitly? Why is this logic not generic?

.github/triage/jax_toolbox_triage/bisect.py
```diff
@@ -0,0 +1,109 @@
...
+        # 1. find the linear range on the main branch
+        passing_main_commit_cmd = f"git merge-base {start} {end}"
+        failing_main_commit_cmd = f"git merge-base {end} {main_branch}"
+
+        passing_main_commit = worker.check_exec(
+            ["sh", "-c", passing_main_commit_cmd], workdir=dir
+        ).stdout.strip()
+        failing_main_commit = worker.check_exec(
+            ["sh", "-c", failing_main_commit_cmd], workdir=dir
+        ).stdout.strip()
```
olupton: As-is the indirection via `sh -c` isn't achieving anything - you can just run `["git", "merge-base", start, end]` and so on.

You could save a command launch overhead by running `["sh", "-c", f"git merge-base {start} {end} && git merge-base {end} {main_branch}"]` and splitting the result in two.

.github/triage/jax_toolbox_triage/bisect.py
```diff
@@ -0,0 +1,109 @@
...
+        # 2. find commits to cherry-pick from the failing branch
+        cherry_pick_cmd = f"git rev-list --reverse {failing_main_commit}..{end}"
+        cherry_pick_commits_list = (
+            worker.check_exec(["sh", "-c", cherry_pick_cmd], workdir=dir)
+            .stdout.strip()
+            .splitlines()
+        )
+        if cherry_pick_commits_list:
+            args.cherry_pick_commits[package] = cherry_pick_commits_list
+        logger.info(f"Cherry-pick commits: {cherry_pick_commits_list}")
```
olupton: Do we have to expand the range into a list of hashes? Can't we just `cherry-pick {failing_main_commit}..{end}` later?

.github/triage/jax_toolbox_triage/bisect.py
```diff
@@ -0,0 +1,109 @@
...
+
+        # 2. find commits to cherry-pick from the failing branch
+        cherry_pick_cmd = f"git rev-list --reverse {failing_main_commit}..{end}"
+        cherry_pick_commits_list = (
+            worker.check_exec(["sh", "-c", cherry_pick_cmd], workdir=dir)
+            .stdout.strip()
+            .splitlines()
+        )
+        if cherry_pick_commits_list:
+            args.cherry_pick_commits[package] = cherry_pick_commits_list
```
olupton: Not sure why this is in `args`? There's no CLI arg, right?

.github/triage/jax_toolbox_triage/main.py
```diff
@@ -1,591 +1,26 @@
...
-    )
 
-    def Container(
-        url, test_output_host_directory: typing.Optional[pathlib.Path] = None
-    ):
-        if args.container_runtime == "local":
-            return LocalContainer(logger=logger)
+    try:
+        tool = TriageTool(args, logger)
+        tool.prepare()
```
olupton: Why have the separate `.prepare` method?

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,588 @@
...
+                args=self.args,
+            )
+
+            if not self.args.cherry_pick_commits.get(package):
+                assert all(
+                    b[1] >= a[1]
+                    for a, b in zip(
+                        package_versions[package], package_versions[package][1:]
+                    )
+                )
```
olupton: I expect this to pass even in the non-linear case.

.github/triage/jax_toolbox_triage/bisect.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    Get the commit history for a given package between two commits.
+
+    Args:
+        worker (Container): The container worker to execute commands.
+        package (str): The name of the package.
+        start (str): The starting commit hash.
+        end (str): The ending commit hash.
+        dir (str): The directory where the git repository is located.
+        main_branch (str, optional): The main branch name. Defaults to None.
```
olupton: I don't think it would work if you actually passed `None`.

You could just declare it compulsory. IMO `None` should mean "the default branch of the repository in question".

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,588 @@
...
+                package_cherry_picks = self.args.cherry_pick_commits.get(package, [])
+                if package_cherry_picks:
+                    self.logger.info("Working on a non-linear history")
+                    git_commands.append(f"cd {self.package_dirs[package]}")
+                    git_commands.append("git stash")
+                    # this is a checkout on the main branch
+                    git_commands.append(f"git checkout {version}")
+                    cherry_pick_str = " ".join(package_cherry_picks)
+                    git_commands.append(
+                        f"git cherry-pick {cherry_pick_str} || (echo 'Cherry-pick failed' && exit 1)"
```
olupton: I would put some extra brackets in to avoid having to think about operator precedence in `" && ".join(git_commands)`.

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,588 @@
...
+                    )
+                else:
+                    # Linear history
+                    # A git repository that exists in the container.
+                    git_commands += [
```
olupton: This is the same in both branches, no need to duplicate.

.github/triage/jax_toolbox_triage/triage_tool.py
olupton: Can't this just go in `__init__`?

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,588 @@
...
+
+        Returns:
+            Tuple[dict, TestResult]: The final versions and the test result.
+        """
+        self.logger.info("Running version-level bisection...")
+        # Prepare the container for the bisection
+        with make_container(
+            self.args.container_runtime,
+            self.bisection_url,
+            self.bazel_cache_mounts,
```
olupton: I think this used to respect `args.container_mount` - it should.

.github/triage/tests/test_triage_tool_class.py
```diff
@@ -0,0 +1,244 @@
...
+        result = subprocess.run(
+            command, cwd=cwd, env=env, check=True, capture_output=True, text=True
+        )
+        return result.stdout.strip()
+    except subprocess.CalledProcessError as e:
+        logging.error(f"Command '{' '.join(command)}' failed with error: {e}")
+        raise e
+
+
+class MockContainer(Container):
```
olupton: Can we just use `LocalContainer`?

.github/triage/tests/test_triage_tool_class.py
```diff
@@ -0,0 +1,244 @@
...
+    """A mock container class for testing purposes."""
+
+    def __init__(self, mock_scripts_path, logger):
+        super().__init__(logger=logger)
+        self.mock_scripts_path = mock_scripts_path
+        self._env = os.environ.copy()
+        self._env["PATH"] = f"{self.mock_scripts_path}:{self._env['PATH']}"
```
olupton: Can we just use absolute paths?

.github/triage/tests/test_triage_tool_class.py
```diff
@@ -0,0 +1,244 @@
...
+        (mock_scripts_path / "build-jax.sh").write_text(build_script_content)
+        os.chmod(mock_scripts_path / "build-jax.sh", 0o755)
+
+        test_case_content = (source_scripts_dir / "test-case.sh").read_text()
+        (mock_scripts_path / "test-case.sh").write_text(test_case_content)
+        os.chmod(mock_scripts_path / "test-case.sh", 0o755)
+
+        # Create a fake bazel executable
+        (mock_scripts_path / "bazel").write_text("#!/bin/sh\necho bazel")
+        os.chmod(mock_scripts_path / "bazel", 0o755)
```
olupton: Because of `bazel clean`?

.github/triage/tests/test_triage_tool_class.py
```diff
@@ -0,0 +1,244 @@
...
+        mock_scripts_path.mkdir()
+
+        source_scripts_dir = pathlib.Path(__file__).parent / "mock_scripts"
+        build_script_content = (source_scripts_dir / "build-jax.sh").read_text()
+        (mock_scripts_path / "build-jax.sh").write_text(build_script_content)
+        os.chmod(mock_scripts_path / "build-jax.sh", 0o755)
+
+        test_case_content = (source_scripts_dir / "test-case.sh").read_text()
+        (mock_scripts_path / "test-case.sh").write_text(test_case_content)
+        os.chmod(mock_scripts_path / "test-case.sh", 0o755)
```
olupton: We can just use the fully-qualified test path as the test case?

.github/triage/tests/test_triage_tool_class.py
```diff
@@ -0,0 +1,244 @@
...
+
+        # setup the jax repo path
+        jax_repo_path = repo_path / "jax"
+        jax_repo_path.mkdir()
+
+        def git_cmd(command, *args):
+            return run_command(["git", command, *args], cwd=jax_repo_path)
+
+        git_cmd("init", "-b", "main")
+        git_cmd("remote", "add", "origin", str(jax_repo_path))
```
olupton: Needed?

.github/triage/tests/test_triage_tool_class.py
```diff
@@ -0,0 +1,244 @@
...
+        git_cmd("commit", "-m", "F1")
+        f1 = git_cmd("rev-parse", "HEAD")
+
+        git_cmd("checkout", "-b", "passing_nonlinear", m2)
+        git_cmd("cherry-pick", f1)
+        passing_nonlinear = git_cmd("rev-parse", "HEAD")
+
+        git_cmd("checkout", "-b", "failing_nonlinear", m3)
+        git_cmd("cherry-pick", f1)
+        failing_nonlinear = git_cmd("rev-parse", "HEAD")
```
olupton: Just use `f1`?

.github/triage/tests/test_triage_tool_class.py
```diff
@@ -0,0 +1,244 @@
...
+
+    class MockArgs:
+        main_branch = "main"
+        bazel_cache = ""
+        build_scripts_path = None
+        test_command = ["test-case.sh", str(jax_repo_path), all_commits["bad_main"]]
+        cherry_pick_commits = {}
+        output_prefix = paths["output"]
+        container_runtime = "mock"  # Use a mock runtime
+        container_mount = []
```
olupton: Can we just use the existing args class?

.github/triage/jax_toolbox_triage/triage_tool.py
Steboss: Yes, I'll move it to `__init__`. I did it for following the original `main.py` but it makes more sense to have it in the constructor

.github/triage/jax_toolbox_triage/bisect.py
Steboss: This is my fault, I wrongly thought we wanted to stop the check for `jax` and `xla` only

.github/triage/tests/test_triage_tool_class.py
Steboss: Good shout, this would allow me to use `LocalContainer` actually. Let me give it a try

.github/triage/tests/test_triage_tool_class.py
Steboss: Correct, I had to add this because I had an error due to bazel

Steboss:
@olupton 
I've fixed the refactoring. There's still an error in the testing, I am working on it. 
I think it would be great to have another review from you, see also if the initial setup to solve non-linear history may be robust enough, otherwise we can further dig into that for a new PR :) 
thank you

.github/triage/jax_toolbox_triage/args.py
```diff
@@ -208,6 +225,13 @@ def parse_args(args=None) -> argparse.Namespace:
             in question has different versions at the endpoints of the bisection range.
         """,
     )
+    version_search_args.add_argument(
+        "--override-remotes",
```
olupton, line 229: `default={}`?

.github/triage/jax_toolbox_triage/bisect.py
```diff
@@ -0,0 +1,116 @@
...
+            )
+            if remote_url_result.returncode == 0:
+                remote_url = remote_url_result.stdout().strip()
+                fetch_cmd = ["git", "fetch", remote_url, start, end]
+                worker.check_exec(fetch_cmd, workdir=dir)
+            else:
+                logger.error(
+                    "No remote 'origin' found and no override provided. Cannot fetch missing commits"
+                )
+                raise Exception("Cannot find commits and no remote is configured")
```
olupton: `worker.check_exec(["git", "fetch", args.override_remotes.get(package, "origin"), start, end], workdir=dir)`?

.github/triage/jax_toolbox_triage/bisect.py
```diff
@@ -0,0 +1,116 @@
...
+            [
+                "sh",
+                "-c",
+                f"git merge-base {start} {end} && git merge-base {end} {main_branch}",
+            ],
+            workdir=dir,
+        ).stdout.strip()
+        passing_main_commit, failing_main_commit = passing_and_failing_cmd.splitlines()
+
+        # 2. find commits to cherry-pick from the failing branch
```
olupton, line 71: Maybe add a TODO that we might like to try, in the future, `{passing_main_commit}..{start}` as an alternative if the first attempt does not apply cleanly?

.github/triage/tests/test_triage_history_bisection.py
olupton, line 1: Which test file should be reviewed?

Steboss: @olupton 
the `test_triage_history_bisection.py` is the test file to be reviewed

.github/triage/tests/test_triage_history_bisection.py
```diff
@@ -0,0 +1,199 @@
...
+        str(paths["scripts"] / "test-case.sh"),
+        str(jax_repo_path),
+        all_commits["bad_main"],
+    ]
+    args = parse_args(arg_list)
+    logger = logging.getLogger(f"Scenario-{scenario}")
+    logging.basicConfig(level=logging.INFO)
+    # mp to mock bazel
+    original_path = os.environ.get("PATH", "")
+    monkeypatch.setenv("PATH", f"{paths['scripts']}:{original_path}")
```
olupton:
```suggestion
    # Ensure bazel and build-jax.sh can be found.
    monkeypatch.setenv("PATH", paths['scripts'], prepend=':')
```

.github/triage/tests/test_triage_history_bisection.py
```diff
@@ -0,0 +1,199 @@
...
+        jax_repo_path.mkdir()
+
+        def git_cmd(command, *args):
+            return run_command(["git", command, *args], cwd=jax_repo_path)
+
+        # main
+        # why don't we push the scripts and use them in repo
+        git_cmd("init", "-b", "main")
+        git_cmd("config", "user.name", "Test User")
+        git_cmd("config", "user.email", "test@user.it")
```
olupton:
```suggestion
        git_cmd("config", "user.email", "test@example.com")
```

.github/triage/tests/test_triage_history_bisection.py
```diff
@@ -0,0 +1,199 @@
...
+        # create a feature branch
+        git_cmd("checkout", "-b", "feature", m1)
+        (jax_repo_path / "feature_file.txt").write_text("feature")
+        git_cmd("add", "feature_file.txt")
+        git_cmd("commit", "-m", "F1")
+        f1 = git_cmd("rev-parse", "HEAD")
+        # here we're applying a feature to the good f1 commit
+        git_cmd("checkout", "-b", "passing_nonlinear", m2)
+        git_cmd("cherry-pick", f1)
+        passing_nonlinear = git_cmd("rev-parse", "HEAD")
```
olupton:
```suggestion
        passing_nonlinear = git_cmd("rev-parse", "HEAD")
```

.github/triage/tests/test_triage_history_bisection.py
```diff
@@ -0,0 +1,199 @@
...
+        # and then we apply the feature to the bad commit
+        # this simulated the rebase scenario
+        git_cmd("checkout", "-b", "failing_nonlinear", m3)
+        git_cmd("cherry-pick", f1)
```
olupton:
```suggestion
        git_cmd("cherry-pick", passing_nonlinear)
```

.github/triage/tests/test_triage_history_bisection.py
```diff
@@ -0,0 +1,199 @@
...
+        build_script_content = (source_scripts_dir / "build-jax.sh").read_text()
+        (mock_scripts_path / "build-jax.sh").write_text(build_script_content)
+        os.chmod(mock_scripts_path / "build-jax.sh", 0o755)
+        # test-case.sh helper test script
+        test_case_content = (source_scripts_dir / "test-case.sh").read_text()
+        (mock_scripts_path / "test-case.sh").write_text(test_case_content)
+        os.chmod(mock_scripts_path / "test-case.sh", 0o755)
+        # fake bazel
+        (mock_scripts_path / "bazel").write_text("#!/bin/sh\nexit 0")
+        os.chmod(mock_scripts_path / "bazel", 0o755)
```
olupton: Why can we not use `source_scripts_dir` as `mock_scripts_path`?

.github/triage/tests/test_triage_history_bisection.py
```diff
@@ -0,0 +1,199 @@
...
+
+        # yield all the info
+        yield {
+            "paths": {
+                "repo": repo_path,
+                "output": output_path,
+                "scripts": mock_scripts_path,
+            },
+            "commits": {
+                "good_main": m2,
```
olupton:
```suggestion
                "good_main": m1,
```

.github/triage/tests/test_triage_history_bisection.py
```diff
@@ -0,0 +1,199 @@
...
+        # main
+        # why don't we push the scripts and use them in repo
+        git_cmd("init", "-b", "main")
+        git_cmd("config", "user.name", "Test User")
+        git_cmd("config", "user.email", "test@user.it")
+        # Create a linear commit history
+        git_cmd("commit", "--allow-empty", "-m", "M1")
+        m1 = git_cmd("rev-parse", "HEAD")
+        git_cmd("commit", "--allow-empty", "-m", "M2")  # good commit
+        m2 = git_cmd("rev-parse", "HEAD")
```
olupton, line 65:
```suggestion
```

.github/triage/tests/test_triage_history_bisection.py
```diff
@@ -0,0 +1,199 @@
...
+        git_cmd("commit", "--allow-empty", "-m", "M3")  # bad commit
+        m3 = git_cmd("rev-parse", "HEAD")
+        # create a feature branch
```
olupton:
```suggestion
        # create a feature branch; feature_file.txt must exist for the mock build-jax.sh to return true
```

.github/triage/tests/test_triage_history_bisection.py
```diff
@@ -0,0 +1,199 @@
...
+        # here we're applying a feature to the good f1 commit
+        git_cmd("checkout", "-b", "passing_nonlinear", m2)
+        git_cmd("cherry-pick", f1)
+        passing_nonlinear = git_cmd("rev-parse", "HEAD")
+        # and then we apply the feature to the bad commit
+        # this simulated the rebase scenario
+        git_cmd("checkout", "-b", "failing_nonlinear", m3)
+        git_cmd("cherry-pick", f1)
+        failing_nonlinear = git_cmd("rev-parse", "HEAD")
+        git_cmd("checkout", "main")
```
olupton: Needed?

.github/triage/tests/test_triage_history_bisection.py
```diff
@@ -0,0 +1,199 @@
...
+        yield {
+            "paths": {
+                "repo": repo_path,
+                "output": output_path,
+                "scripts": mock_scripts_path,
+            },
+            "commits": {
+                "good_main": m2,
+                "bad_main": m3,
+                "feature": f1,
```
olupton:
```suggestion
```

.github/triage/tests/test_triage_history_bisection.py
```diff
@@ -0,0 +1,199 @@
...
+    tool = TriageTool(args, logger)
+    tool.package_dirs = {"jax": str(jax_repo_path)}
+    tool.dynamic_packages = {"jax"}
+    tool.bisection_url = "local"
+    if scenario == "Linear History":
+        linear_build_script_path = paths["scripts"] / "build-jax.sh"
+        linear_build_script_path.write_text(
+            "#!/bin/sh\necho 'Mock linear build successful.'\nexit 0"
+        )
+        os.chmod(linear_build_script_path, 0o755)
```
olupton: How about instead of doing this you just add an extra couple of commits on top of `F1` (or a cherry-pick of it) and use those for the linear case? i.e. `F1 -> L1 -> L2 -> L3` and the linear triage is between L1 and L3.

.github/triage/tests/test_triage_history_bisection.py
```diff
@@ -0,0 +1,199 @@
...
+
+    # run the bisection
+    tool.run_version_bisection(passing_versions, failing_versions)
```
olupton: Maybe this should just return `summary_data`?

.github/triage/tests/test_triage_history_bisection.py
Steboss: We could, but we would prevent from running these tests in isolation: 
- if we keep the `mock_scripts_path` we can run `non-linear` and `linear` tests, without having problems for overwriting files. At the moment, we're not use tons of files, but it's safer to keep the two scenarios separated
- if we want to run tests locally, we're not writing anything in dirs and folders

.github/triage/tests/test_triage_history_bisection.py
Steboss: I love this suggestion!

Steboss:
@olupton 
re-ready for review 
thank you 🙏

.github/triage/jax_toolbox_triage/bisect.py
```diff
@@ -0,0 +1,122 @@
...
+            )
+            if remote_url_result.returncode == 0:
+                remote_url = remote_url_result.stdout().strip()
+                fetch_cmd = ["git", "fetch", remote_url, start, end]
+                worker.check_exec(fetch_cmd, workdir=dir)
+            else:
+                logger.error(
+                    "No remote 'origin' found and no override provided. Cannot fetch missing commits"
+                )
+                raise Exception("Cannot find commits and no remote is configured")
```
olupton: As I suggested in https://github.com/NVIDIA/JAX-Toolbox/pull/1538#discussion_r2201181033 I think this can be made way shorter and more direct
```suggestion
        worker.check_exec(["git", "fetch", args.override_remotes.get(package, "origin"), start, end], workdir=dir)```

File: .github/triage/jax_toolbox_triage/bisect.py [hunk 51]
```diff
@@ -0,0 +1,122 @@
...
+        # 2. find commits to cherry-pick from the failing branch
+        # TODO: as an alternative approach we may need to consider `{passing_main_commit}..{start}`
+        cherry_pick_range[package] = f"{failing_main_commit}..{end}"
+
+        # 3. now we can use the main branch  commits for bisection
+        start = passing_main_commit
+        end = failing_main_commit
+
+    logger.info(
+        f"INFO: cherry_pick_range {cherry_pick_range}, start: {start} and end {end}"
```
olupton:
```suggestion
        f"cherry_pick_range: {cherry_pick_range}, start: {start}, end: {end}"
```

.github/triage/jax_toolbox_triage/bisect.py
```diff
@@ -0,0 +1,122 @@
...
+    )
+    # check if the start is the root commit. We may have to deal with the very start of the repo
+    # so we need to handle this case too
+    parent_check_result = worker.check_exec(
+        ["git", "rev-list", "--parents", "-n", "1", start], workdir=dir
+    )
+    is_root_commit = len(parent_check_result.stdout.strip().split()) == 1
+    log_range = f"{start}..{end}" if is_root_commit else f"{start}^..{end}"
```
olupton: I doubt this is ever relevant in the real world, I would be happy to just error out in that case and avoid paying for an extra `check_exec`.

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,583 @@
...
+        assert container_url is not None, (
+            "Container URL must be provided if explicit versions are not set."
+        )
+
+        with make_container(
+            self.args.container_runtime,
+            container_url,
+            self.bazel_cache_mounts,
+            self.logger,
+        ) as worker:
```
olupton: Maybe make a `_make_container(self, url, extra_mounts=[])` wrapper for this? It seems inconsistent in different places, in particular regarding the mounts.

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,583 @@
...
+
+            assert all(
+                b[1] >= a[1]
+                for a, b in zip(
+                    package_versions[package], package_versions[package][1:]
+                )
+            )
+
+            # this check works only for linera-case
+            if not self.cherry_pick_commits.get(package):
```
olupton:
```suggestion
            if package not in self.cherry_pick_commits:
```

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,583 @@
...
+            test_output_log_level (int): The log level for test output.
+        Returns:
+            TestResult: The result of the test, including whether it passed and the output.
+        """
+        container_url_func = functools.partial(
+            container_url_base,
+            container=self.args.container,
+            template=self.args.container_url_template,
+        )
+        container_url = container_url_func(date)
```
olupton:
```suggestion
        container_url = container_url_base(date, container=self.args.container, template=self.args.container_url_template)
```

.github/triage/jax_toolbox_triage/summary.py
```diff
@@ -0,0 +1,78 @@
...
+        if section in data:
+            logging.warning(f"Overwriting summary data in section {section}")
+        data[section] = record
+    else:
+        if section not in data:
+            data[section] = []
+        data[section].append(record)
+
+    with open(summary_filename, "w") as ofile:
+        json.dump(data, ofile)
```
olupton, line 45:
```suggestion
    with open(summary_filename, "w") as ofile:
        json.dump(data, ofile)
    return data      
```

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -0,0 +1,583 @@
...
+        )
+        result["container"] = self.bisection_url
+        add_summary_record(self.args.output_prefix, "result", result, scalar=True)
+        self.logger.info("Version-level bisection completed")
+
+        summary_file = self.args.output_prefix / "summary.json"
+        with open(summary_file, "r") as ifile:
+            summary_data = json.load(ifile)
+
+        return summary_data
```
olupton:
```suggestion
        self.logger.info("Version-level bisection completed")
        return add_summary_record(self.args.output_prefix, "result", result, scalar=True)
```

.github/triage/tests/test_triage_history_bisection.py
```diff
@@ -0,0 +1,214 @@
...
+        # this simulated the rebase scenario
+        git_cmd("checkout", "-b", "failing_nonlinear", m3)
+        git_cmd("cherry-pick", passing_nonlinear)
+        failing_nonlinear = git_cmd("rev-parse", "HEAD")  # F1'
+        # so now we have:
+        # M1 --- M2 --- M3
+        # |             |
+        # F1           F1'
+        # where F1 = passing
+        # and F2 = failing
```
olupton:
```suggestion
        # and F1' = failing
```
?

.github/triage/tests/test_triage_history_bisection.py
olupton: I don't follow, what is overwritten? Seems like all that's needed is that the scripts in `source_scripts_dir` are committed with exec permissions and that `source_scripts_dir` is in `PATH` when the test runs.

Steboss: you're actually right 
I am going to try with this suggestion, this would make the code more readable too

.github/triage/jax_toolbox_triage/bisect.py
Steboss: True, I had to introduce this because of the test.

Steboss: I'll modify this :)

.github/triage/jax_toolbox_triage/triage_tool.py
Steboss: Great idea

Steboss:
@olupton 
re-re-ready for a new review

PR #1549, opened 2025-07-10
Add Experimental Library for Calling Cute/Cutlass DSL Kernels in Jax

mgoldfarb-nvidia:
This PR contains the first implementation of our integration of Cute DSL into Jax.

### DIFF
--- a/.github/container/Dockerfile.base
+++ b/.github/container/Dockerfile.base
@@ -18,6 +18,7 @@ ARG CLANG_VERSION
 ARG JAX_TOOLBOX_REF
 ARG NSIGHT_SYSTEMS_VERSION_OVERRIDE
 ENV CUDA_BASE_IMAGE=${BASE_IMAGE}
+ENV JAX_TOOLBOX_REF=${JAX_TOOLBOX_REF}
 
 ###############################################################################
 ## Install Python and essential tools
--- a/.github/container/cutlass_dsl_jax/LICENSE
+++ b/.github/container/cutlass_dsl_jax/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
+
+   2. Grant of Copyright License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      copyright license to reproduce, prepare Derivative Works of,
+      publicly display, publicly perform, sublicense, and distribute the
+      Work and such Derivative Works in Source or Object form.
+
+   3. Grant of Patent License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      (except as stated in this section) patent license to make, have made,
+      use, offer to sell, sell, import, and otherwise transfer the Work,
+      where such license applies only to those patent claims licensable
+      by such Contributor that are necessarily infringed by their
+      Contribution(s) alone or by combination of their Contribution(s)
+      with the Work to which such Contribution(s) was submitted. If You
+      institute patent litigation against any entity (including a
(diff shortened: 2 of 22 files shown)


gpupuck (dismissed):
Looks good to me!

.github/container/jax_cutlass/README.md
```diff
@@ -0,0 +1,165 @@
...
+    constval=1.0,
+    ...
+)
+```
+
+## Layouts and Modes
+
+`cutlass_call` accepts two optional parameters to aid in converting `jax.Array` into a `cute.Tensor`:
+
+_Layout_: Describes the physical order of axis strides for the source `jax.Array`. If omitted the array is assumed to have row-major layout.
```
jaro-sevcik: Are these always tuples? It also seems that you specify the layout and/or mode for inputs and outputs, could you spell that out here?

.github/container/jax_cutlass/README.md
```diff
@@ -0,0 +1,165 @@
...
+A common example of when to use modes is to implement logical layouts for gemm operations. For example we want the `cute.Tensor`s to follow a consistent layout specification of `[M][K][L]` for A, `[N][K][L]` for B and `[M][N][L]` for C. Strides of each dimension are set to properly offset into physical memory. 
+
+
+### Complex Layouts and Modes
+
+If your kernel requires more complex layouts at the function boundary e.g. tiled, composed or hierarchical it is recommended that the kernel be wrapped in a `cute.jit` function. Once you are outside of Jax you can built arbitrary layouts as needed using CuTe DSL.
+
+## Compilation Cache
+
+Like Jax, we maintain a cache of the compiled functions. When a kernel is compiled we check if it was previously seen for the given shapes, dtypes and constant values. For example if your kernel is called in a series of homogeneous layer (e.g a transformer model) it will only need to compile once and that instance can be reused as needed.
```
jaro-sevcik: Is there a user-configurable directory that can be reused across ranks? If not, then I would perhaps not imply that it is "Like JAX".

.github/container/jax_cutlass/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Limitations
+
+There are several limitations to highlight to avoid unexpected errors or behavior. Over time we hope to improve these as CuTe DSL matures.
+
+### Jit Function Argument Types
+
+`cutlass_call` does not allow for arbitrary python types to be passed between Jax and the DSL kernel. If the kernel interface depends on a complex Python type it is recommended that a wrapper function be used to bind together `jax.Array` and other compile time constants that can be passed.
```
jaro-sevcik: Please spell out which types are allowed. Only `jax.Array`s? Or also lists and tuples of `jax.Arrays`?

.github/container/jax_cutlass/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+@cute.jit
+def wrapper(stream: cuda.CUstream, a: cute.Tensor, b: cute.Tensor, *, constval: float):
+    x = CustomType(a, b, constval)
+    launch(stream, x)
+
+out = cutlass_call(wrapper, ..., constval=1.0)(a, b)
+```
+
+A useful trick can be to flatten a `PyTree` structure outside the call then unflatten inside the call by passing the `PyTree` as a constexpr argument.
```
jaro-sevcik: Is there a reason why `cutlass_call` does not do that automatically? (Similarly to `buffer_callback`.)

.github/container/jax_cutlass/README.md
```diff
@@ -0,0 +1,165 @@
...
+    @cute.jit
+    def __call__(self, stream, ...):
+        # use self.v0, self.v1
+        ...
+
+def make_function(v0, v1):
+    return MyFunction(v0, v1)
+```
+
+One common exception is if you need to
```
jaro-sevcik: Incomplete sentence.

.github/container/jax_cutlass/jax_cutlass/primitive.py
```diff
@@ -0,0 +1,282 @@
...
+    input_mode_flat: Any,
+    output_mode_flat: Any,
+    input_output_aliases,
+    convert_tensors,
+    allow_cuda_graph,
+    **kwargs,
+):
+    input_output_aliases = dict(input_output_aliases)
+
+    # TODO: Need to support device-less compilation and deffer module load
```
jaro-sevcik: Nit: s/deffer/defer/

.github/container/jax_cutlass/jax_cutlass/primitive.py
```diff
@@ -0,0 +1,282 @@
...
+
+def _cutlass_call_vmap_rule(*args, **kwargs):
+    del args, kwargs
+    raise NotImplementedError(
+        "cutlass_call does not support batching with jax.vmap. Please "
+        "use jax.custom_batching.custom_vmap for applying vmap. "
+    )
+
+
+batching.primitive_batchers[cutlass_call_inner_p] = _cutlass_call_vmap_rule
```
jaro-sevcik: Suggestion for future: Perhaps you could wire up vmap_method from JAX FFI/buffer_callback.

.github/container/jax_cutlass/tests/test_args.py
```diff
@@ -0,0 +1,209 @@
...
+    for ci, ci_ref in zip(c, c_ref):
+        assert jnp.allclose(ci, ci_ref)
+
+
+def test_list_args_alias():
+
+    @cute.kernel
+    def kernel(
+        a: cute.Tensor,
+        b: list[cute.Tensor],
```
jaro-sevcik: How about also testing tuple args?

.github/container/jax_cutlass/README.md
mgoldfarb-nvidia: Yes these are tuples currently. More complex layouts/orders could require custom types. I think we could also accept `Layout` from jax and normalize/convert it. Tuples are at least safe choice for now I think.

Do you have any thoughts on `Layout` and its relation here?

I will update the wording to reflect these are specified individually for both input and output even when there is aliasing.

.github/container/jax_cutlass/README.md
mgoldfarb-nvidia: The cache is in memor- we don't save it to disk although that is a feature I'd like to add support for in a follow up.

I'll remove `Like Jax` to avoid confusion.

.github/container/jax_cutlass/README.md
mgoldfarb-nvidia: Ooops will finish this one :)

.github/container/jax_cutlass/tests/test_args.py
mgoldfarb-nvidia: Good suggestion, I will make one of the `list` a `tuple`.

.github/container/jax_cutlass/README.md
mgoldfarb-nvidia: We actually do automatically flatten and unflatten to support list types. But I left more complicated mapping to the user. Perhaps this last sentence will just cause confusion?

There are a few challenges that make it tricky and why I left it as a future feature:
1. There are limitations to what types can pass the jit boundry and a custom protocol is used to augment user defined types. Primitive types like dict are not allowed although it may be possible to wrap/inherit. We might be able to provide utilities to augment dataclasses and provide documentation for adventurous users.

2. Input / Output representation in a Pytree gets tricky since you may need to alias arrays in the tree if the tree combines both inputs and outputs (e.g. an update operation). In this case it might be better to fully alias the entire tree than try to return a new tree with updates that get merged in somehow.

The current simplified approach basically says if you have some complex pytree structure you'd need to do the work to flatten it prior to calling cutlass kernel. But I agree it would be nice to support more natively arbitrary Pytrees but its going to take some more work.

.github/container/jax_cutlass/jax_cutlass/primitive.py
mgoldfarb-nvidia: Agree will look at this for next update.

.github/container/jax_cutlass/README.md
mgoldfarb-nvidia: Done. Let me know if you fell this still needs to be made more clear.

.github/container/jax_cutlass/README.md
jaro-sevcik: Makes sense, thank you for the clarification.

.github/container/jax_cutlass/README.md
jaro-sevcik: I see - I did not mean if the individual layouts are tuples, I meant if the structure of layouts is also tuples.

For example, if the call takes a 2-element list of arrays and one other array, would the input_layout be something like `input_layout = ([(0, 1), (2, 1, 0)], (0, 1, 2))` or rather `input_layout = (((0, 1), (2, 1, 0)), (0, 1, 2))` (or does not it matter)?

Also, it would be nice to have some tests for the layouts, if only for documentation purposes.

.github/container/jax_cutlass/README.md
```diff
@@ -0,0 +1,174 @@
...
+            block=[a.shape[-1], 1, 1],
+            stream=stream)
+
+call = cutlass_call(launch, output_shape_dtype=jax.ShapeDtypeStruct((128, 64), jnp.float16))
+out = call(input)
+```
+
+## Host Function Signature
+
+`cutlass_call` requires a specific function signature to bind arrays and constant values provided by Jax. It is recommend that you annotate the signature with the appropriate or expected types.
```
jaro-sevcik: s/recommend/recommended/

.github/container/jax_cutlass/README.md
jaro-sevcik: I was wondering if you should also say that each argument can also be a list of tensors or a tuple of tensors, but perhaps that is too much information for the docs.

jaro-sevcik (dismissed):

.github/container/jax_cutlass/README.md
mgoldfarb-nvidia: Oh I see, let me double check, I think today it may assume a flattened list.

PR #1561, opened 2025-07-16
Add Experimental Library for Calling Cute/Cutlass DSL Kernels in Jax

mgoldfarb-nvidia:
This PR contains the first implementation of our integration of Cute DSL into Jax.

### DIFF
--- a/.github/eks-workflow-files/jax-cutlass/scripts/unittest.sh
+++ b/.github/eks-workflow-files/jax-cutlass/scripts/unittest.sh
@@ -1,28 +1,28 @@
-set -xu -o pipefail
+          set -xu -o pipefail
 
-LOG_DIR=${LOG_DIR:-/opt/output}
+          LOG_DIR=${LOG_DIR:-/opt/output}
 
-SRC_ROOT=${SRC_ROOT:-$PWD/jax-cutlass-src}
-SRC_ROOT=$(realpath $SRC_ROOT)
+          SRC_ROOT=${SRC_ROOT:-$PWD/jax-cutlass-src}
+          SRC_ROOT=$(realpath $SRC_ROOT)
 
-pip install pytest-reportlog pytest-xdist
+          pip install pytest-reportlog pytest-xdist
 
-# nvidia-cutlass-dsl-jax is not yet installed to the container by default.
-# Clone if not already present locally as indicated by SRC_ROOT
-if [[ ! -d ${SRC_ROOT} ]]; then
-  git clone https://github.com/NVIDIA/JAX-Toolbox.git --branch ${JAX_TOOLBOX_REF} ${SRC_ROOT}
-  PIP_SRC=${SRC_ROOT}/.github/container/cutlass_dsl_jax
-else
-  PIP_SRC=${SRC_ROOT}
-fi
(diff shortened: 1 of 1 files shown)


.github/workflows/_jax_cutlass_eks.yaml
```diff
@@ -0,0 +1,151 @@
...
+    - name: Download and process logs from S3
+      id: s3-logs-process
+      run: |
+        LOCAL_DIR=${{ inputs.CI_NAME }}-output
+        REPORT_JSON=pytest-report.jsonl
+
+        mkdir -p $LOCAL_DIR
+        aws s3 cp s3://${{ inputs.S3_BUCKET }}/${{ inputs.CI_NAME }}/${{ env.RUN_NAME }}/ $LOCAL_DIR/ --recursive 
+
+        passed_tests=$(cat $LOCAL_DIR/$REPORT_JSON |  jq --slurp | jq '.[] | select(.outcome == "passed") | .outcome' | wc -l || true)
```
olupton, line 91: Happened to notice this is returning 48 for your test suite of 16 tests. Looks like it includes setup/teardown/execution separately.

.github/eks-workflow-files/jax-cutlass/scripts/unittest.sh
```diff
@@ -0,0 +1,28 @@
...
+          fi
+
+          pip install ${PIP_SRC}
+
+          NGPUS=$(nvidia-smi --list-gpus | wc -l)
+
+          # Start MPS daemon
+          nvidia-cuda-mps-control -d
+
+          pytest-xdist.sh ${NGPUS} 1 ${LOG_DIR}/pytest-report.jsonl pytest -xsv --log-file=${LOG_DIR}/pytest_log.log --log-file-level=INFO ${PIP_SRC}/tests/ | tee -a ${LOG_DIR}/pytest_stdout_dist.log
```
olupton, line 26: It shouldn't cause any problems, but MPS/xdist aren't needed if you're just running 1 worker that can see all of the GPUs.

.github/workflows/_jax_cutlass_eks.yaml
mgoldfarb-nvidia: Its also possible some tests are run multiple times and that inflates the count. We do this to detect some interesting issues .

.github/eks-workflow-files/jax-cutlass/scripts/unittest.sh
mgoldfarb-nvidia: Thanks! The naming was a bit confusing to me. If we were to do `pytest-xdist.sh ${NGPUS} 2` would that launch two process each with `NGPU/2` visible devices?

olupton: No, it's `GPUS_PER_WORKER WORKERS_PER_GPU`.
So `${NGPUS} 2` is two workers that can both see `NGPUS` devices, which in your case is all eight of them (i.e. oversubscribing the GPUs)
If you want 2 processes that see disjoint sets of 4 GPUs, you'd want `4 1`.

PR #1593, opened 2025-07-31
Update AXLearn performance script 

Steboss:
Add the metrics to the output log file

### DIFF
--- a/.github/container/fuji-train-perf.py
+++ b/.github/container/fuji-train-perf.py
@@ -202,6 +202,16 @@ def extract_metrics(
     # tflops/s/n_devices, take the very last number
     tflops_s_devices = model_flop_per_iteration / times_arr[-1] / world_size / 10**12
 
+    # re-add to the input file the final metrics
+    with open(input_file, "a") as f:
+        f.write(
+            f"\n=== Final metrics ===\n"
+            f"Tokens/s/device: {tokens_per_sec_gpu.mean()} +/- {tokens_per_sec_gpu.std()}\n"
+            f"Seqs/s/device: {seqs_per_sec_gpu.mean()} +/- {seqs_per_sec_gpu.std()}\n"
+            f"AvgTimestep: {times_arr.mean()} +/- {times_arr.std()}\n"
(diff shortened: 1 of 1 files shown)


.github/container/fuji-train-perf.py
```diff
@@ -202,6 +202,16 @@ def extract_metrics(
     # tflops/s/n_devices, take the very last number
     tflops_s_devices = model_flop_per_iteration / times_arr[-1] / world_size / 10**12
 
+    # re-add to the input file the final metrics
+    with open(input_file, "a") as f:
+        f.write(
+            f"\n=== Final metrics ===\n"
+            f"Tokens/s/device: {tokens_per_sec_gpu.mean()} +/- {tokens_per_sec_gpu.std()}\n"
+            f"Seqs/s/device: {seqs_per_sec_gpu.mean()} +/- {seqs_per_sec_gpu.std()}\n"
+            f"AvgTimestep: {times_arr.mean()} +/- {times_arr.std()}\n"
```
gpupuck, line 214: A bit confused, what's `Timestep`?

Steboss: Hey @gpupuck 
`Timestep` the name is a bit misleading, it comes directly from AXLearn. It's how many seconds we take to perform a single training step. 
BTW I was thinking of saving those metrics into a specific `metric.log` file, rather than writing them directly in the output log. WDYT?

gpupuck: I think printing the metrics to stdout is good enough. It's easier to process the log if all the metrics are there.

Steboss: That's true 
However, I am now running some tests on cloud and I need to save those metrics in a storage, to check them after the model has run

olupton: Just do both? Print human-readable to stdout, dump machine-readable (json?) values to a separate file.

Steboss: You're right Olli
I ended up to that  :) no need to change the script

PR #1597, opened 2025-08-01
war: Pin pip-tools and patch ARM build

gpupuck:
This PR applies two WARs to fix the build issues in current CI:
1. pip-tools was upgraded from 7.4.1 to 7.5.0 on July 31. However, pip-tools 7.5.0 will cause `ValueError: '/opt/maxtext/requirements.txt (line 1)' is not in the subpath of '/opt/pip-tools.d'`. I'm guessing something is not quite compatible with the Python 3.12 we current have in the base container. Theoretically, -r ../maxtext/requirements.txt should work, but since we are using a specific version of pip. Let's play safe at this point and use 7.4.1
2. ARM build has been failing for days, this is due to the issue https://github.com/openxla/xla/issues/28256 from Hermetic C++ changes and no fix yet. `--config=ci_linux_aarch64_cuda` is necessary but the `HERMETIC_CUDA_COMPUTE_CAPABILITIES` override is a hot fix which would be adjusted later once we have a good fix from Google


Steboss (approved):

.github/container/Dockerfile.jax
```diff
@@ -6,7 +6,7 @@ ARG BUILD_PATH_JAXLIB=/opt/jaxlibs
 ARG EXTRA_BAZEL_TARGETS=""
 ARG URLREF_JAX=https://github.com/google/jax.git#main
 ARG URLREF_XLA=https://github.com/openxla/xla.git#main
-ARG URLREF_FLAX=https://github.com/google/flax.git#main
```
Steboss: Do we need to update also the manifest.yaml? e.g. https://github.com/NVIDIA/JAX-Toolbox/blob/b4bf8c075fedcdec6014de6a37feab4bfc66a751/.github/container/manifest.yaml#L14

Steboss (approved):

Steboss (approved):

PR #1602, opened 2025-08-04
flax: pin v0.10.6

olupton:
Should fix the MaxText build.


.github/container/create-distribution.sh
```diff
@@ -139,7 +139,7 @@ cd ${INSTALLED_DIR}
 
 for git_command in cherry-pick rebase merge revert; do
   RESERVED_INPROGRESS_REF=$(echo $git_command | tr 'a-z' 'A-Z' | sed 's/-/_/g')_HEAD
-  if git rev-parse --verfiy $RESERVED_INPROGRESS_REF >/dev/null 2>&1; then
+  if git rev-parse --verify $RESERVED_INPROGRESS_REF >/dev/null 2>&1; then
```
gpupuck, line 142: nice catch

PR #1609, opened 2025-08-07
Update `xpk` `GKE` cluster create workflow

aybchan:
Enable the `xpk` `GKE` cluster create workflow to trigger on workflow dispatch events, and upgrade `xpk` to `v0.10.1`

The included patch file on  `xpk@v0.10.1`is necessary for:
1. Addressing an error on GPU node pool creation (comment out placement policy)
```
Google Compute Engine: Not all instances running in IGM after 22.157464014s. Expected 2, running 0, transitioning 2. 
Current errors: [UNSUPPORTED_OPERATION]: Instance 'gke-jtb-2025-08-06-jtb-2025-08-06-a3--954bb82f-1rsz' 
creation failed: To consume reservation with placement policy, the instance has to specify the same resource policy as the
 reservation.
```
2. Enabling [GPU Operator](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/google-gke.html) install (requires node label, and disabling of the GKE managed driver)

A service account to enable pod interactivity with GCS buckets is also added

Installation of GPU operator is added

n.b This workflow is not operational in the CI, it is added as reference recipe for `GKE` cluster creation with `xpk`

### DIFF
--- a/.github/container/Dockerfile.maxtext
+++ b/.github/container/Dockerfile.maxtext
@@ -37,8 +37,7 @@ for pattern in \
     "s|absl-py|absl-py>=2.1.0|g" \
     "s|protobuf==3.20.3|protobuf>=3.19.0|g" \
     "s|tensorflow-datasets|tensorflow-datasets>=4.8.0|g" \
-    "s|sentencepiece==0.1.97|sentencepiece>=0.2|g" \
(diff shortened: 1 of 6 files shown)


olupton (dismissed):
LGTM, though I don't see the GPU Operator installation anywhere?

olupton (approved):

PR #1622, opened 2025-08-13
Support custom MAXTEXT_DIR and --quantization option

jberchtold-nvidia:
This PR updates `test-maxtext.sh` with two features.

1. If a user specifies `MAXTEXT_DIR` in their environment, this dir will be used instead of `/opt/maxtext`. This makes it easier for development. Default behavior of using /opt/maxtext is unchanged unless that environment variable is specified.
2. Support for more quantization recipes via MaxText's `quantization` config via a new `--quantization` arg. 

Before this PR, this script uses `--dtype=fp8` to specify MaxText FP8 DelayedScaling. However, when developing new quantization recipes, support for different types of quantization beyond "fp8" is useful. For backwards compatibility, I've kept the behavior of `--dtype=fp8` unchanged. But you can now specify `--quantization="..."` with an arbitrary string and it will be passed to MaxText's `quantization` config unfiltered. An error is raised when both `--dtype=fp8` and `--quantization="..."` are specified at the same time.

### DIFF
--- a/.github/container/test-maxtext.sh
+++ b/.github/container/test-maxtext.sh
@@ -47,7 +47,7 @@ usage() {
     exit $1
 }
 
-args=$(getopt -o a:b:s:o:n:h --long additional-args:,mem-fraction:,model-name:,decoder-block:,attn-type:,remat-policy:,batch-per-gpu:,dtype:,steps:,help,multiprocess,output:,data-parallel:,fsdp:,tensor-parallel:,pipeline-parallel:,nodes: -- "$@")
+args=$(getopt -o a:b:s:o:n:h --long additional-args:,mem-fraction:,model-name:,decoder-block:,attn-type:,remat-policy:,batch-per-gpu:,dtype:,quantization:,steps:,help,multiprocess,output:,data-parallel:,fsdp:,tensor-parallel:,pipeline-parallel:,nodes: -- "$@")
 if [[ $? -ne 0 ]]; then
     exit $1
 fi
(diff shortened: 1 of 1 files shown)


.github/container/test-maxtext.sh
```diff
@@ -107,6 +108,10 @@ while [ : ]; do
         DTYPE="$2"
         shift 2
         ;;
+    --quantization)
+        QUANTIZATION="$2"
+        shift 2
+        ;;
```
olupton, line 114: What's the distinction between this and `--dtype`?

jberchtold-nvidia: The existing behavior of `--dtype` is that it is hardcoded to only work with `fp8` for MaxText's DelayedScaling FP8 in `test-maxtext.sh`.
```
if [ $DTYPE == "fp8" ]; then
    ADDITIONAL_ARGS+=" quantization=$DTYPE"
fi
```

We are in the process of adding new TE quantization recipes, such as `te_fp8_delayedscaling`, `te_mxfp8`, etc.

Since these are not really unique `dtype`s and more so recipes, I've exposed this quantization option more directly instead of extending `dtype`.

To avoid breaking any existing workflows relying on `--dtype=fp8`, I've left dtype unchanged (with an error message if both dtype and quantization are specified). Long term, I think just `--quantization` would be sufficient as `--quantization=fp8` will still work to enable MaxText's DelayedScaling FP8.

If you have a different preference on naming of these args lmk and I can update the PR

olupton (approved):

PR #1643, opened 2025-08-26
Update default GKE cluster settings

aybchan:


olupton (approved):
LGTM, we can look at the failing MaxText job in a follow-up

PR #1658, opened 2025-09-05
Check whether the EKS cluster is under maintenance and skip steps

Steboss:
Little modification to the CI to avoid running on EKS for the moment, as the cluster is under maintenance

### DIFF
--- a/.github/actions/submit-delete-k8s-job/action.yml
+++ b/.github/actions/submit-delete-k8s-job/action.yml
@@ -16,6 +16,12 @@ runs:
       uses: ./.github/actions/with-post-step
       with:
         main: |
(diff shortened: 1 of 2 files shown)


aybchan (changes requested):
```
Error from server (Forbidden): nodes is forbidden: User "arn:aws:sts::941377147396:assumed-role/runner-role/i-0fd068c4a6deac5bf" cannot list resource "nodes" in API group "" at the cluster scope
```
https://github.com/NVIDIA/JAX-Toolbox/actions/runs/17496944800/job/49702868098#step:7:18

Looks like an uncaught failure, and the job was still submitted

Steboss:
@aybchan fixed

PR #1663, opened 2025-09-08
Modify installation of MaxText using editable mode

Steboss:


gpupuck (approved):
Let's move it forward first

PR #1698, opened 2025-09-29
triage-tool: two bugfixes

olupton:
Bug one: if a package FOO has different versions in the two endpoint containers of a triage, as given by the FOO_VERSION environment variable inside those containers, but no installFOO.sh script is available, previously it was not excluded properly from the triage.

Bug two: manually interrupting a triage with Ctrl-C did not kill the child processes, leading to orphaned `srun` commands etc.

Tests added covering both issues.


Steboss (approved):
mamma mia! 🚀

PR #2035, opened 2026-04-13
Add multiprocess as part of our JAX tests

Steboss:


.github/container/test-jax.sh
```diff
@@ -181,6 +181,9 @@ case "${BATTERY}" in
             "//tests:gpu_tests"
             "//tests/pallas:gpu_tests"
             "//tests/mosaic:gpu_tests"
+            "//tests/multiprocess:all"
+            # avoid this test that runs on tpu, that's likely to fail
+            "-//tests/multiprocess:tpu_device_test_tpu"
```
olupton:
```suggestion
            "//tests/multiprocess:gpu_tests"
```
I would try this to match upstream:
https://github.com/jax-ml/jax/blob/6165fcf4e95f2f96ec691f85428f06e8dea84a50/.github/workflows/bazel_cuda_h100_b200.yml#L147

olupton (changes requested):

olupton (approved):

PR #2037, opened 2026-04-15
Add EKS JobSet action

aybchan:
- [x] Add EKS JobSet action
- [x] Port A100 tests (JAX, nsys, TE, MaxText) to EKS H100


.github/workflows/_ci.yaml
```diff
@@ -246,83 +246,121 @@ jobs:
...
+          COMMAND: |
+            set -xu -o pipefail;
+
+            mkdir -p /opt/output;
+
+            pip install pytest-reportlog pytest-xdist;
+            nvidia-cuda-mps-control -d;
+            export TE_PATH=\${SRC_PATH_TRANSFORMER_ENGINE};
+
+            [[ "${{ matrix.test }}" == *unittest* ]] && xdist_args=(4 1) || xdist_args=(8 1);
```
olupton: This looks a bit weird. I think the point of the old script was to have `8 1` for the `distributed` tests and `1 4` for the non-distributed ones.

.github/workflows/_ci.yaml
```diff
@@ -246,83 +246,121 @@ jobs:
...
-        test-single-gpu.log
-    secrets: inherit
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - test: backend-independent
+            gpus: 1
+          - test: single-gpu
+            gpus: 1
```
olupton, line 256: I am not sure this is really doing anything? At least `nvidia-smi` in the job log shows it can see all 8 GPUs.

Either way, I think seeing all 8 is fine because `test-jax.sh` / `bazel` know how to take advantage of them and parallelise the single-gpu tests across all available devices.

aybchan: This was due to mis-set securityContext of containers in the JobSet

.github/workflows/_ci.yaml
```diff
@@ -246,83 +246,121 @@ jobs:
...
-    secrets: inherit
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - test: backend-independent
+            gpus: 4
+          - test: single-gpu
+            gpus: 4
+          - test: multi-gpu
```
olupton, line 257: Why 4?

.github/workflows/_ci.yaml
```diff
@@ -246,83 +246,121 @@ jobs:
...
+          COMMAND: |
+            set -xu -o pipefail;
+
+            mkdir -p /opt/output;
+
+            pip install pytest-reportlog pytest-xdist;
+            nvidia-cuda-mps-control -d;
+            export TE_PATH=\${SRC_PATH_TRANSFORMER_ENGINE};
+
+            [[ "${{ matrix.test }}" == *distributed* ]] && xdist_args=(8 1) || xdist_args=(4 1);
```
olupton: I think this is still inverted. We used to have:
```
 # 1 GPU per worker, 4 workers per GPU
pytest-xdist.sh 1 4 ${LOG_DIR}/pytest-report-L0-unittest.jsonl bash ${TE_PATH}/qa/L0_jax_unittest/test.sh | tee -a ${LOG_DIR}/pytest_stdout.log
# 8 GPUs per worker, 1 worker per GPU. pytest-xdist.sh allows aggregation
# into a single .jsonl file of results from multiple pytest invocations
# inside the test.sh script, so it's useful even with a single worker per
# device.
pytest-xdist.sh 8 1 ${LOG_DIR}/pytest-report-L0-distributed-unittest.jsonl bash ${TE_PATH}/qa/L0_jax_distributed_unittest/test.sh | tee -a ${LOG_DIR}/pytest_stdout.log
```

```suggestion
            [[ "${{ matrix.test }}" == *distributed* ]] && xdist_args=(8 1) || xdist_args=(1 4);
```

.github/workflows/_ci.yaml
```diff
@@ -525,10 +493,63 @@ jobs:
...
+            gpus: 8
+            pipeline-parallel: 1
+            data-parallel: 1
+            fsdp: 2
+            tensor-parallel: 4
+            envs: |-
+              NCCL_IB_DISABLE=1
+              NCCL_P2P_DISABLE=1
+              NCCL_NET_PLUGIN=none
+              NCCL_NET=Socket
```
olupton: Is this because of https://github.com/aws/aws-ofi-nccl/issues/1134 ?
I would hope that `NCCL_NET_PLUGIN=none` is sufficient ?
Are these definitely needed?

.github/workflows/_ci.yaml
```diff
@@ -525,10 +493,63 @@ jobs:
...
+          - test: multi-node
+            nodes: 2
+            gpus: 8
+            pipeline-parallel: 1
+            data-parallel: 4
+            fsdp: 2
+            tensor-parallel: 2
+            envs: |-
+              OFI_NCCL_PROTOCOL=SENDRECV
```
olupton, line 518: Why?

olupton (dismissed):
A few small things.

.github/workflows/_ci.yaml
aybchan:
```
NCCL_NET_PLUGIN=none
NCCL_NET=Socket
```
Is the minimum set necessary - it still tries to load the EFA plugin without `NCCL_NET=Socket` leading to seg fault on single node pretrain

.github/workflows/_ci.yaml
aybchan: I think this may be related to https://github.com/aws/aws-ofi-nccl/issues/1134 - without setting this, multi-node pretrain seg faults

olupton (approved):

PR #2047, opened 2026-04-20
Make sure scale-training runs every 2 weeks

Steboss:
- We want to run the scale-training every 2 weeks
- We won't target Sevin's XLA, but upstream XLA. In particular, there will be a `nv-staging/YYYY-MM-DD` branch
- Since we can't have in the CI a crontab simple enough to say:  "every 2 weeks", we anchor last Friday as the initial date to compare against.


gpupuck (approved):
The XLA repo shift is great, much simpler!

PR #2048, opened 2026-04-21
nsys-jax: fix collective metadata extraction.

olupton:
Also make jax-nccl-test more useful.


.github/container/nsys_jax/nsys_jax/scripts/jax_nccl_test.py
olupton, line 1: This is not really new, but it is a significant rewrite to `.github/container/jax-nccl-test`

.github/container/nsys_jax/pyproject.toml
```diff
@@ -11,7 +13,7 @@ dependencies = [
   "uncertainties", # communication analysis recipe
   "prettytable",
 ]
-requires-python = ">= 3.10"
+requires-python = ">= 3.12"
```
Steboss, line 16: if we're requiring python >=3.12, should we change also the `venv` in this script https://github.com/NVIDIA/JAX-Toolbox/blob/main/.github/container/nsys_jax/nsys_jax/scripts/nsys_jax.py#L69C3-L69C64 ?

.github/container/nsys_jax/nsys_jax/scripts/jax_nccl_test.py
```diff
@@ -0,0 +1,629 @@
...
+            "Divide the GPUs into this many parallel groups that do not communicate with one another. For example with 8 GPUs and a replica count of 2, the emitted collectives will be for 2 groups of 4 GPUs."
+        ),
+        type=int_or_env,
+        default=1,
+    )
+    parser.add_argument(
+        "--collectives",
+        help="Which collectives to profile.",
+        action="append",
+        default=["all_gather", "all_reduce", "broadcast", "permute", "reduce_scatter"],
```
Steboss: I think we should have this list empty if we have `append`, otherwise running this with `--collectives all_gather`, will run the test with all the default list + `all_gather` again - am I wrong?

Steboss (dismissed):

.github/container/nsys_jax/pyproject.toml
olupton: Good spot; changed.

.github/container/nsys_jax/nsys_jax/scripts/jax_nccl_test.py
olupton: Good spot; changed.

Steboss (approved):

PR #2053, opened 2026-04-23
Nvidia-runners S3 cache implementation

Steboss:


Steboss:
@olupton 
Here is an initial setup for using nvidia-enterprise-runners + S3 bucket. 
At the moment I've used a test bucket within our JAX AWS space. 

Here is a comparison between this implementation and OCI one ( I took the very last `main` CI for the comparison):

Process AMD64 | OCI (s) | S3 cache (s)
-- | -- | --
Jaxlib wheel | 1017 | 721
CUDA plugin wheel | 47 | 50
CUDA PJRT wheel | 488 | 461
Extra XLA tools | 26 | 25


Process ARM64 | OCI (s) | S3 cache (s)
-- | -- | --
Jaxlib wheel | 931 | 824
CUDA plugin wheel | 104 | 70
CUDA PJRT wheel | 771 | 543
Extra XLA tools | 90 | 36


I moved also our tests to EKS, but we need to discuss what tests we need to run, if the tests I've implemented make sense/preserve what we have before, and what to do with nsys-jax tests. If you could have a look at this please. 
thank you :)

.github/actions/build-container/action.yml
```diff
@@ -88,10 +96,20 @@ runs:
         password: ${{ inputs.github-token }}
 
     - name: Set up Docker Buildx
+      if: ${{ inputs.docker-build-network != 'host' }}
+      uses: docker/setup-buildx-action@v4
+      with:
+        driver-opts: |
+          image=moby/buildkit:v0.12.1
+        version: v0.30.1
```
olupton: IIRC these pins were related to the old builders being old. Hopefully they can be removed now?

.github/actions/build-container/action.yml
```diff
@@ -88,10 +96,20 @@ runs:
...
+
+    - name: Set up Docker Buildx with host network
+      if: ${{ inputs.docker-build-network == 'host' }}
       uses: docker/setup-buildx-action@v4
       with:
         driver-opts: |
           image=moby/buildkit:v0.12.1
+          network=host
```
olupton: https://docs.docker.com/reference/cli/docker/buildx/create/#buildkit-daemon-network-mode (not 100% sure this is the right page) suggests the default is `auto` -- if that's right, we could just plumb through `inputs.docker-build-network` here, make that the default, and avoid the duplication.

.github/eks-workflow-files/maxtext/test.yml
```diff
@@ -0,0 +1,69 @@
...
+                          --attn-type dot_product \
+                          --batch-per-gpu 2 \
+                          --steps 10 \
+                          --pipeline-parallel 1 \
+                          --data-parallel 1 \
+                          --fsdp 2 \
+                          --tensor-parallel 4 \
+                          --nodes 1
+
+                      # multi-process: PP=1, DP=2, FSDP=2, TP=2
```
olupton: I'm surprised if this works...

.github/workflows/_ci.yaml
```diff
@@ -53,20 +54,101 @@ jobs:
...
+          uses: aws-actions/configure-aws-credentials@v4
+          with:
+            role-to-assume: ${{ secrets.TEST_ROLE_FOR_AWS_S3 }}
+            aws-region: us-east-1
+        - name: Start bazel-remote sidecar
+          env:
+            BAZEL_REMOTE_BUCKET: jax-toolbox-bazel-cache-test
+            BAZEL_REMOTE_AWS_DIR: /tmp/bazel-remote-aws
+            BAZEL_REMOTE_CONTAINER: bazel-remote-cache-${{ inputs.ARCHITECTURE }}
+            BAZEL_REMOTE_DISK_DIR: /tmp/bazel-remote-cache
```
olupton, line 83: Does this need to be a host directory? I assumed it could just be container-local storage.

.github/workflows/_ci.yaml
```diff
@@ -53,20 +54,101 @@ jobs:
...
-    runs-on: [self-hosted, "${{ inputs.ARCHITECTURE }}", "large"]
+    runs-on: ${{ inputs.ARCHITECTURE == 'amd64' && 'linux-amd64-cpu32m' || 'linux-arm64-cpu32m' }}
     steps:
         - name: Checkout repository
           uses: actions/checkout@v6
+        - name: Configure AWS credentials via OIDC
+          uses: aws-actions/configure-aws-credentials@v4
+          with:
+            role-to-assume: ${{ secrets.TEST_ROLE_FOR_AWS_S3 }}
+            aws-region: us-east-1
```
olupton: Maybe this should be a repository variable/secret instead?

.github/workflows/_ci.yaml
```diff
@@ -53,20 +54,101 @@ jobs:
...
+              --http_address=0.0.0.0:8080 \
+              --s3.auth_method=aws_credentials_file \
+              --s3.aws_profile=default \
+              --s3.aws_shared_credentials_file=/aws-config/credentials \
+              --s3.bucket="${BAZEL_REMOTE_BUCKET}" \
+              --s3.endpoint="s3.${BAZEL_REMOTE_REGION}.amazonaws.com" \
+              --s3.prefix="${BAZEL_REMOTE_PREFIX}" \
+              --s3.region="${BAZEL_REMOTE_REGION}" \
+              --s3.update_timestamps
+            sleep 2
```
olupton: Seems like it could be flaky. Loop? Or fold into the next step.

.github/workflows/_ci.yaml
```diff
@@ -82,19 +164,43 @@ jobs:
...
+          if: always()
+          uses: actions/upload-artifact@v6
+          with:
+            name: bazel-remote-diagnostics-${{ inputs.ARCHITECTURE }}
+            path: /tmp/bazel-remote-diagnostics
+        - name: Stop bazel-remote sidecar
+          if: always()
+          run: |
+            docker rm -f "bazel-remote-cache-${{ inputs.ARCHITECTURE }}" || true
+            rm -rf /tmp/bazel-remote-aws
```
olupton, line 198: Is this actually needed?

.github/actions/build-container/action.yml
```diff
@@ -54,6 +54,14 @@ inputs:
...
     required: true
     default: ""
+  docker-build-allow:
+    description: "Additional BuildKit entitlements to allow during docker build"
+    required: false
+    default: ""
+  docker-build-network:
+    description: "Networking mode for RUN instructions during docker build"
+    required: false
+    default: default
```
olupton: `auto` ?

olupton (commented):
https://github.com/NVIDIA/JAX-Toolbox/actions/runs/24884720555/job/72862175533?pr=2053#step:7:5245 looks like we still run through the whole build again for `final` having done it for `mealkit`. Maybe the runners' local cache directories are too small?

The logic to spin up the remote cache server proxying to S3 looks good.

.github/eks-workflow-files/maxtext/test.yml
Steboss: ah, this is just an initial attempt, I need to fix this

.github/workflows/_ci.yaml
Steboss: It doesn't have to be a host directory, a docker volume would work as well. I want to keep this bare minimal changes, without modifying at the docker level.

.github/workflows/_ci.yaml
Steboss: Yes and no
With this we can stop having port 9090 occupied and delete temporary credential files. I would keep it or move it to a post-cache action

.github/actions/build-container/action.yml
Steboss: Good shout

Steboss:
@olupton 
A little update on the caching of `mealkit` for building `final` JAX image. 
After testing some unsuccessful solutions proposed by the nv-gha-runners team, I came up with [this strategy](https://github.com/NVIDIA/JAX-Toolbox/compare/sbosisio/custom-s3-cache-with-local-aws...sbosisio/test-local-host-for-mealkit?expand=1). Here we are using a local-registry that pushes `mealkit` (here to `ghcr` and `localhost:5000` but we could push it to `localhost:5000` only). 
This enables to saved time, avoiding the whole rebuild of `final`. [Here](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/25554140396/job/75009254514) is the example with `amd64`. 
The approach is minimal, doesn't introduce more tech debt or breaking changes. 
If you're ok with it, I can clean up this repo, introduce this localhost registry, and we can have a first solution for substituting OCI in build-jax steps.

olupton:
> A little update on the caching of `mealkit` for building `final` JAX image. After testing some unsuccessful solutions proposed by the nv-gha-runners team, I came up with [this strategy](https://github.com/NVIDIA/JAX-Toolbox/compare/sbosisio/custom-s3-cache-with-local-aws...sbosisio/test-local-host-for-mealkit?expand=1).

Great!

> Here we are using a local-registry that pushes `mealkit` (here to `ghcr` and `localhost:5000` but we could push it to `localhost:5000` only).

We have to keep pushing it to somewhere non-local because the JAX `mealkit` is the input to, for example, the MaxText `mealkit`.

> This enables to saved time, avoiding the whole rebuild of `final`. [Here](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/25554140396/job/75009254514) is the example with `amd64`. The approach is minimal, doesn't introduce more tech debt or breaking changes. If you're ok with it, I can clean up this repo, introduce this localhost registry, and we can have a first solution for substituting OCI in build-jax steps.

Let's try it!

We don't need to do this now, but we can speed this up more by enabling build caching for TE (already implemented internally).

Steboss:
@olupton 
I cleaned the repo and fixed the process + created secrets and the terraform side on AWS. 
There’s one part of this that I’m not especially excited about, but I think it may be the only practical option: adding a CI dispatch step.
We need this to trigger the CI for any pull requests, given the [doc from nvidia runners](https://docs.gha-runners.nvidia.com/platform/onboarding/pull-request-testing/#example). 
The cleanest alternative would be to adopt a strict branch naming convention, for example `pull-request/sbosisio-something`, so that CI can continue to trigger through push. But if we want to keep normal feature branch names, a dispatch-based approach seems like the most reliable solution.
Let me know what you think of it. Thank you :)

.github/actions/build-container/action.yml
olupton: Did this not work?

.github/workflows/_ci.yaml
```diff
@@ -55,20 +56,101 @@ jobs:
...
+            umask 077
+            {
+              printf '[default]\n'
+              printf 'aws_access_key_id=%s\n' "${AWS_ACCESS_KEY_ID}"
+              printf 'aws_secret_access_key=%s\n' "${AWS_SECRET_ACCESS_KEY}"
+              printf 'aws_session_token=%s\n' "${AWS_SESSION_TOKEN}"
+            } > "${BAZEL_REMOTE_AWS_CREDENTIALS}"
+            set -x
+            docker rm -f "${BAZEL_REMOTE_CONTAINER}" >/dev/null 2>&1 || true
+            docker pull buchgr/bazel-remote-cache
```
olupton: Needed? Don't the runners always start clean?

.github/workflows/pr-dispatch-ci.yaml
Steboss, line 1: I didn't see it 
That's great i'll open another PR for setting up the copy-pr-bot configurations

.github/actions/build-container/action.yml
Steboss: Sorry I forgot to correct this. 
Actually `auto` doesn't work, e.g. [here](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/25803932631/job/75812854666#step:4:269), as per `docker buildx create --driver-opt network=...`  Docker expects an actual network choice it can attach the builder container to, like `host`, `bridge`, or a named Docker network. 
However, we can fix this and avoid repetition by using an empty string for all the other builds, and `host` for JAX (e.g. [here](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/25808106738/job/75826867628) )

.github/workflows/_ci.yaml
Steboss: Good shout

.github/workflows/pr-dispatch-ci.yaml
Steboss: Opened a PR #2102

Steboss:
> This pull request requires additional validation before any workflows can run on NVIDIA's runners.
> 
> Pull request vetters can view their responsibilities [here](https://docs.gha-runners.nvidia.com/cpr/vetters).
> 
> Contributors can view more details about this message [here](https://docs.gha-runners.nvidia.com/cpr/contributors).

@olupton it works

.github/workflows/pr-dispatch-ci.yaml
olupton, line 1: This can go away, right?

Steboss:
/ok to test [fde4e51](https://github.com/NVIDIA/JAX-Toolbox/pull/2053/commits/fde4e51bae9b8b4f2cb0560f92b85dfe70ed7fe7)

.github/workflows/_ci.yaml
```diff
@@ -55,20 +56,100 @@ jobs:
...
-    with:
-      CONTAINER: ${{ needs.build-base.outputs.DOCKER_TAG }}
-    secrets: inherit
+  # test-nccl:
+  #   if: inputs.ARCHITECTURE == 'amd64' # build only amd64
+  #   needs: build-base
+  #   uses: ./.github/workflows/_test_nccl.yaml
+  #   with:
+  #     CONTAINER: ${{ needs.build-base.outputs.DOCKER_TAG }}
+  #   secrets: inherit
```
olupton: Needs to be reverted

olupton (dismissed):

.github/workflows/_ci.yaml
Steboss: thanks for spotting this. reverted

olupton (approved):

PR #2066, opened 2026-04-30
Start scaffolding AF

Steboss:


Steboss:
This PR adds Alphafold3 as part of our containers: 
- added `alphafold_files` with the following inputs: 
    - `input_sequences`. These are input files that can be used for benchmark on different tokens' lengths 512-5120 
    - `af3_inference_benchmark.py` script for running benchmarks
    - `generate_af_parameters.py` script to generate AF's models (fake) weights
    - `model_parameters.md` model parameters description for generating weights - this is given by AF repo itself
    - `README.md` to explain the content 
- `Dockerfile.alphafold` installs Alphafold with upstream JAX + we run AF's `build_data` to generate needed chemical data for inference

.github/container/Dockerfile.alphafold
```diff
@@ -0,0 +1,46 @@
...
+RUN cat ${SRC_PATH_AF}/pyproject.toml
+
+###############################################################################
+## Add all the needed scripts
+###############################################################################
+
+ADD alphafold_files/generate_af_parameters.py /usr/local/bin
+ADD alphafold_files/input_sequences /opt/alphafold/input_sequences
+ADD alphafold_files/af3_inference_benchmark.py /usr/local/bin
+ADD alphafold_files/model_parameters.md /opt/alphafold/model_parameters.md
```
olupton: This gives an `/opt/alphafold` that is a checkout of the upstream repo with some extra files copied on top. Is that desirable? Maybe we should keep the two sets of files under separate prefixes?

.github/container/Dockerfile.alphafold
```diff
@@ -0,0 +1,46 @@
...
+###############################################################################
+## Install accumulated packages from the base image and the previous stage
+###############################################################################
+
+FROM mealkit AS final
+
+RUN pip-finalize.sh
+
+WORKDIR ${SRC_PATH_AF}
+RUN build_data
```
olupton: What does this do? Does the result really want to be baked into the container?

.github/workflows/_ci.yaml
```diff
@@ -306,7 +334,7 @@ jobs:
 
             mkdir -p /opt/output;
 
-            pip install pytest-reportlog pytest-xdist triton;
```
olupton, line 309: Why? Last I checked this was a real test-time dependency of TE.

.github/container/Dockerfile.alphafold
Steboss: Good shout, yes I think we can move them on another location, there's no need to have those with the upstream repo

.github/container/Dockerfile.alphafold
Steboss: Let's say it's the easiest thing. 
Ideally we would have to execute [this script](https://github.com/google-deepmind/alphafold3/blob/main/src/alphafold3/build_data.py). `RUN build_data` is also what AF people are doing. I can avoid having this, and run on top of any benchmark run .

.github/workflows/_ci.yaml
Steboss: I am 100% sure this is me, that have clear something when commented the other ci steps for testing.

.github/container/Dockerfile.alphafold
Steboss: Ok i think that in our benchmarking script we can have something like: 
```
python -c "from alphafold3.build_data import build_data; build_data()"
```

.github/container/alphafold_files/af3_inference_benchmark.py
```diff
@@ -0,0 +1,307 @@
...
+    ap.add_argument(
+        "--warm-runs", type=int, default=3
+    )  # we are going to run 3 times the main benchmark and report stats from there. There's an additional run on top to avoid compilation time impacting the bench.
+    ap.add_argument("--jax-compilation-cache-dir", default=None)
+    ap.add_argument(
+        "--gpu-index",
+        type=int,
+        default=0,
+        help="Index of the GPU to use for benchmarking (default: 0)",
+    )  # AF3 runs on process per GPU. We can run multiple different jobs on different gpu-indexes
```
olupton: I don't see any `jax.distributed.initialize`, so it's less 'process per GPU mode' (which usually means multiple processes each driving one GPU collectively running a multi-GPU program) and just "this is a single-GPU benchmark", right?

In that case, I don't think there's a need to put any explicit logic in the script. You can just use CUDA_VISIBLE_DEVICES if you really want to run it twice in parallel.

.github/container/alphafold_files/af3_inference_benchmark.py
```diff
@@ -0,0 +1,307 @@
...
+    )  # this is the number of samples AF3 generates from teh diffusion head. For benchmarking 1 is a good starting point
+    ap.add_argument(
+        "--flash-attention-implementation",
+        default="triton",
+        choices=["triton", "cudnn", "xla"],
+    )  # here we can chose what FA to use, ideally Triton > CuDNN > XLA
+    ap.add_argument(
+        "--warm-runs", type=int, default=3
+    )  # we are going to run 3 times the main benchmark and report stats from there. There's an additional run on top to avoid compilation time impacting the bench.
+    ap.add_argument("--jax-compilation-cache-dir", default=None)
```
olupton: Any need for an explicit option here? JAX natively supports setting this from the environment if needed.

.github/workflows/_ci.yaml
```diff
@@ -707,3 +736,43 @@ jobs:
       with:
         job-config-file:  ".github/eks-workflow-files/axlearn/axlearn-fuji-model.yml"
         job-name: ${{ env.JOB_NAME }}
+
+  test-alphafold-inference-eks:
```
olupton, line 740: @aybchan added a cleaner wrapper for this, e.g. https://github.com/NVIDIA/JAX-Toolbox/blob/cc5e9abfee6ab7e27810715ed479787767ab45dd/.github/workflows/_ci.yaml#L241-L277 -- can we use that here?

.github/workflows/_ci.yaml
```diff
@@ -204,6 +204,34 @@ jobs:
...
+          BASE_IMAGE: ${{ needs.build-jax.outputs.DOCKER_TAG_MEALKIT }}
+          CONTAINER_NAME: alphafold
+          DOCKERFILE: .github/container/Dockerfile.alphafold
+          RUNNER_SIZE: large
+          ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
+          ssh-known-hosts: ${{ vars.SSH_KNOWN_HOSTS }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          bazel-remote-cache-url: ${{ vars.BAZEL_REMOTE_CACHE_URL }}
+          EXTRA_BUILD_ARGS: |
+            URLREF_ALPHAFOLD=${{ fromJson(inputs.SOURCE_URLREFS).ALPHAFOLD }}
```
olupton, line 233:
```suggestion
            URLREF_AF=${{ fromJson(inputs.SOURCE_URLREFS).ALPHAFOLD }}
```
or update the Dockerfile

.github/workflows/_ci.yaml
```diff
@@ -204,6 +204,34 @@ jobs:
...
+          BASE_IMAGE: ${{ needs.build-jax.outputs.DOCKER_TAG_MEALKIT }}
+          CONTAINER_NAME: alphafold
+          DOCKERFILE: .github/container/Dockerfile.alphafold
+          RUNNER_SIZE: large
+          ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
+          ssh-known-hosts: ${{ vars.SSH_KNOWN_HOSTS }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          bazel-remote-cache-url: ${{ vars.BAZEL_REMOTE_CACHE_URL }}
+          EXTRA_BUILD_ARGS: |
+            URLREF_AF=${{ fromJson(inputs.SOURCE_URLREFS).ALPHAFOLD }}
```
olupton: Sorry, should `.ALPHAFOLD` be `.AF` or is it OK?

olupton (dismissed):

.github/workflows/_ci.yaml
Steboss: It's correct `ALPHAFOLD`
Should I change `URLREF_AF` to `UREFLREF_ALPHAFOLD` for consistency?

olupton: I'll leave that to you

Steboss: Ok moved everything to `ALPHAFOLD`, I prefer consistency. 
Can you re-check and re-approve please? thank you

olupton (approved):

PR #2113, opened 2026-05-18
Restart triage from interrupted run

Steboss:


.github/triage/jax_toolbox_triage/args.py
```diff
@@ -100,14 +100,31 @@ def parse_args(args=None) -> argparse.Namespace:
...
+        """,
+    )
+    parser.add_argument(
+        "--restart-folder",
+        type=pathlib.Path,
+        help="""
+            Output folder from a previous triage run. Must contain summary.json and is
+            used as --output-prefix during --restart.
+        """,
+    )
```
olupton, line 119: Do we need both of these?

.github/triage/jax_toolbox_triage/utils.py
```diff
@@ -25,16 +25,20 @@ def container_url(
...
 
-def get_logger(output_prefix: pathlib.Path) -> logging.Logger:
-    output_prefix.mkdir()
+def get_logger(output_prefix: pathlib.Path, append: bool = False) -> logging.Logger:
+    output_prefix.mkdir(exist_ok=append)
     logger = logging.getLogger("triage")
     logger.setLevel(logging.DEBUG)
+    for handler in logger.handlers:
+        handler.close()
+    logger.handlers.clear()
```
olupton: Why do we need this?

.github/triage/jax_toolbox_triage/logic.py
```diff
@@ -330,6 +341,7 @@ def _version_search(
     logger: logging.Logger,
     skip_precondition_checks: bool,
     result_cache: typing.Dict[FlatVersionDict, TestResult],
+    preloaded_cache_keys: typing.Set[FlatVersionDict],
```
olupton: I think we could just delete this? It seems like a lot of plumbing to keep a sanity-check assertion (unexpected cache hit) alive.

.github/triage/jax_toolbox_triage/logic.py
```diff
@@ -701,6 +705,8 @@ def version_search(
     skip_precondition_checks: bool,
     check_success_before_failure: bool = True,
     confirmation_iterations: int = 1,
+    result_cache: typing.Optional[typing.Dict[FlatVersionDict, TestResult]] = None,
```
olupton:
```suggestion
    result_cache: typing.Dict[FlatVersionDict, TestResult] = {},
```

.github/triage/jax_toolbox_triage/summary.py
```diff
@@ -46,6 +51,105 @@ def add_summary_record(
...
+    if isinstance(value, bool):
+        return (
+            TestExecutionOutcome.TEST_SUCCESS
+            if value
+            else TestExecutionOutcome.TEST_FAILURE
+        )
+    if isinstance(value, str):
+        name = value.rsplit(".", 1)[-1]
+        return TestExecutionOutcome[name]
+    raise ValueError(f"Cannot parse test result from {value!r}")
```
olupton: Why is this needed? How do we end up with heterogeneous data?

.github/triage/jax_toolbox_triage/summary.py
```diff
@@ -46,6 +51,105 @@ def add_summary_record(
...
+        cache[key] = TestResult(
+            build_stdouterr=None,
+            host_output_directory=_record_output_directory(output_prefix, record),
+            result=_parse_result(record["result"]),
+            stdouterr=None,
+        )
+    return cache
+
+
+def container_result_cache_from_summary(
```
olupton: Do we need to have separate methods here? It would seem simpler to me to just load one summary.json into one cache that covers both version-level and container-level searches?

.github/triage/jax_toolbox_triage/summary.py
```diff
@@ -67,13 +171,19 @@ def create_output_symlinks(
     def symlink(result: typing.Optional[TestResult], symlink_name: str) -> None:
         if result is None:
             return
-        symlink_path = (output_prefix / symlink_name).resolve()
-        assert not symlink_path.exists(), symlink_path
-        assert symlink_path.parent == result.host_output_directory.parent, (
-            symlink_path,
+        symlink_path = output_prefix / symlink_name
+        if symlink_path.exists() or symlink_path.is_symlink():
```
olupton: Not sure I follow this. I think the only time there are symlinks is if the triage converged. If it converged, why is anyone running the tool again with restart?

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -51,6 +57,16 @@ def __init__(self, args, logger):
         self.packages_with_scripts = set()
         self.bazel_cache_mounts = prepare_bazel_cache_mounts(self.args.bazel_cache)
         self.check_success_before_failure = True
+        self.restart_summary = (
+            load_summary(self.args.restart_folder) if args.restart else {}
+        )
+        self.container_result_cache = (
```
olupton: Do we actually need this? Can't we just optionally load the cache here?

.github/triage/jax_toolbox_triage/args.py
Steboss: Fixed, keeping `--restart-folder` as it's the main thing we need here.

.github/triage/jax_toolbox_triage/utils.py
Steboss: 100% no

.github/triage/jax_toolbox_triage/logic.py
Steboss: 100% agree

.github/triage/jax_toolbox_triage/args.py
```diff
@@ -308,6 +317,27 @@ def parse_args(args=None) -> argparse.Namespace:
...
+        if not summary_file.exists():
+            raise Exception(
+                f"--restart-folder must contain summary.json: {summary_file}"
+            )
+        if (
+            args.output_prefix is not None
+            and args.output_prefix.resolve() != args.restart_folder
+        ):
+            raise Exception(
+                "--output-prefix must match --restart-folder when restarting"
```
olupton: With these semantics, isn't a boolean `--restart` sufficient?

.github/triage/jax_toolbox_triage/logic.py
olupton: As well as this, we might not even need `_version_search` anymore? I think it only existed because of `result_cache` not being in the top-level API.

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -686,6 +714,24 @@ def run_version_bisection(
 
         # Run the version-level bisection
         self.logger.info("Running version-level bisection...")
+        result_cache = {}
+        if self.args.restart_folder is not None:
+            self.restart_cache.update(
+                result_cache_from_summary(
```
olupton: Why do we have to do this? Why isn't it enough to just populate the in-memory cache from JSON at startup and pass that into the two searches?

Steboss:
@olupton I made some changes: 
- to perform the triage restart we just need `jax_toolbox_triage ... --restart --output-prefix OLD_FOLDER`
- collapsed `_version_search` into `version_search`
- simplified the restart cache loading: 
    - load `summary.json` once at startup
    - use that for container-level cache hits
    - derive the version-level  cache locally after package histories are known

.github/triage/jax_toolbox_triage/args.py
```diff
@@ -100,14 +100,24 @@ def parse_args(args=None) -> argparse.Namespace:
...
         type=pathlib.Path,
     )
+    parser.add_argument(
+        "--restart",
+        action="store_true",
+        help="""
+            Restart a previous triage run from --output-prefix. The output directory
+            must contain summary.json, which is used as the source of truth for
+            completed records. Incomplete output directories without summary records
+            are ignored.
```
olupton:
```suggestion
            completed records.
```

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -697,6 +725,22 @@ def run_version_bisection(
...
         # Run the version-level bisection
         self.logger.info("Running version-level bisection...")
+        result_cache = {}
+        if self.args.restart:
+            summary_cache = result_cache_from_summary(
+                self.args.output_prefix,
+                package_versions.keys(),
+                self.restart_summary,
+            )
+            result_cache = {
```
olupton: I am still missing why this needs to be (re?)loaded late instead of at startup?

Steboss: Sorry Olli I forgot to answer here!
I'll fix this, as it was just used for testing with debug logs

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -51,6 +58,16 @@ def __init__(self, args, logger):
...
+        self.restart_summary = (
+            load_summary(self.args.output_prefix) if args.restart else {}
+        )
+        self.restart_cache = (
+            result_cache_from_summary(
+                self.args.output_prefix, summary=self.restart_summary
+            )
+            if args.restart
+            else {}
+        )
```
olupton:
```suggestion
        self.restart_cache = (
            result_cache_from_summary(
                self.args.output_prefix, summary=load_summary(self.args.output_prefix)
            )
            if args.restart
            else {}
        )
```

olupton (dismissed):

olupton (approved):

PR #2128, opened 2026-05-29
[jax-inference-offloading] bump vLLM version to 0.21.0

yhtang:


yhtang:
Superseded by #2141, which updates the JIO stack to vLLM 0.22/CUDA 13 and carries the CUDA package skiplist path.

PR #2141, opened 2026-06-03
[jax-inference-offloading] update CUDA 13 vLLM/PyTorch stack and replace stale CI with 1B smoke

yhtang:
Supersedes #2128.

## Summary

- update the JIO CUDA 13 dependency stack, including vLLM 0.22 and PyTorch CUDA 13 wheels
- add custom JIO Docker refs so image builds can select a specific JIO repository and commit
- add a CUDA package skiplist helper so the Docker build does not reinstall CUDA payloads already provided by the base image, while still installing missing runtime packages
- keep the JIO examples compatible with the updated JAX/vLLM surface
- replace stale JIO CI coverage with an 8B decoupled synchronous smoke that uses the CI-accessible `meta-llama/Llama-3.1-8B-Instruct` model
- pin `huggingface-hub` for the smoke download path and use the HF CLI with the CI HF token
- preserve the GKE TCPXO NCCL path update, source the TCPXO NCCL env profile, pass the tuner and guest-checker configs, and let that profile provide enforced NCCL values such as `NCCL_BUFFSIZE`
- keep vLLM logs visible and fail fast when any decoupled smoke component exits non-zero, so CI reports real failures instead of silent passes

## Validation

- `python3 -m py_compile jax-inference-offloading/dockerfile/cuda_package_skiplist.py`
- `git diff --check`
- CUDA 13 image build completed successfully
- smoke-checked the rebuilt container:
  - PyTorch reports CUDA 13
  - vLLM 0.22 imports successfully
  - JAX sees CUDA devices
  - `pip check` passes
- #2147 validation before stack merge:
  - JAX-vLLM offloading workflow run `26973090959` passed
  - `jio-decoupled-sync-8b-gke-smoke` passed with `EXIT_CODE=0`
  - smoke sitrep reported `1/1 passed`

## CI note

The active JIO runtime smoke is now the 8B decoupled synchronous path. This avoids the CI HF-token access issue seen with the 1B gated model and exercises the current split JAX/vLLM offloading flow.

The legacy 8B/70B transfer, GRPO, and EKS jobs still target removed flat example entrypoints, so they are preserved but disabled pending a follow-up port.

### DIFF
--- a/jax-inference-offloading/examples/single_controller/run_single_controller.sh
+++ b/jax-inference-offloading/examples/single_controller/run_single_controller.sh
@@ -175,7 +175,7 @@ done
 
 MODEL_NAME=${MODEL_NAME:-"meta-llama/Llama-3.2-1B-Instruct"}
 
-trap "trap - SIGTERM && kill -- -$$" SIGINT SIGTERM EXIT
+trap "trap - SIGTERM EXIT && kill -- -$$ 2>/dev/null || true" SIGINT SIGTERM EXIT
 
 if [[ -f "${PWD}/.env" ]]; then
   echo "Loading ${PWD}/.env"
--- a/jax-inference-offloading/examples/single_controller/run_single_controller.sh
+++ b/jax-inference-offloading/examples/single_controller/run_single_controller.sh
@@ -175,7 +175,16 @@ done
 
 MODEL_NAME=${MODEL_NAME:-"meta-llama/Llama-3.2-1B-Instruct"}
 
-trap "trap - SIGTERM EXIT && kill -- -$$ 2>/dev/null || true" SIGINT SIGTERM EXIT
+cleanup() {
+  local status=$?
+  trap - SIGINT EXIT
+  trap '' SIGTERM
+  kill -- -$$ 2>/dev/null || true
+  wait 2>/dev/null || true
+  exit "${status}"
+}
+
+trap cleanup SIGINT SIGTERM EXIT
 
 if [[ -f "${PWD}/.env" ]]; then
   echo "Loading ${PWD}/.env"
--- a/.github/actions/build-container/action.yml
+++ b/.github/actions/build-container/action.yml
@@ -66,6 +66,10 @@ inputs:
     description: "Optional named build context to source from a runner registry for the final image handoff"
     required: false
     default: ""
+  require-nsys-jax-version-artifact:
+    description: "Whether this container build requires the nsys-jax version.py artifact"
+    required: false
+    default: "true"
 
 outputs:
   DOCKER_TAG_MEALKIT:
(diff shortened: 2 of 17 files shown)


yhtang:
CI update: I disabled the stale runtime jobs rather than deleting them. They still target removed flat example entrypoints.

The active runtime smoke now uses the current 1B `single_controller` example. The model-scope reduction was already present in earlier commits/codebase state; this change just points CI at the current supported smoke path. Larger-model and GRPO/EKS coverage should come back in a follow-up PR.

aybchan (commented):
Smoke test had a [silent failure](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/26934294017/job/79461761894#step:3:1375) in case you didn't see it. 

Added a couple of fixes in #2147 to address HuggingFace issues. I also requested access to the `meta-llama/Llama-3.2-1B-Instruct` repo which was missing for the token. The run from this branch could load the model from HF and has a new failure ([`EngineCore failed to start`](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/26945297199/job/79498460038#step:3:1280)).

jax-inference-offloading/examples/single_controller/run_single_controller.sh
```diff
@@ -175,7 +175,7 @@ done
 
 MODEL_NAME=${MODEL_NAME:-"meta-llama/Llama-3.2-1B-Instruct"}
 
-trap "trap - SIGTERM && kill -- -$$" SIGINT SIGTERM EXIT
+trap "trap - SIGTERM EXIT && kill -- -$$ 2>/dev/null || true" SIGINT SIGTERM EXIT
```
Steboss: Could we avoid killing the wrapper shell from the `EXIT` trap here? 
The new smoke runs this script under GNU `timeout`, which puts the child in its own process group. That means `kill -- -$$` can send SIGTERM to this Bash process during normal `EXIT`, turning a clean run into exit 143.

I think we can keep the process-group cleanup but preserve the original status and ignore SIGTERM inside cleanup, e.g.

```bash
cleanup() {
  local status=$?
  trap - SIGINT EXIT
  trap '' SIGTERM
  kill -- -$$ 2>/dev/null || true
  wait 2>/dev/null || true
  exit "${status}"
}

trap cleanup SIGINT SIGTERM EXIT

yhtang: Implemented in `869e17f7`. The cleanup now captures the original status, disables the `EXIT` trap, ignores `SIGTERM` while terminating the process group, waits for children, and exits with the captured status. Applied to both `single_controller` and the active `decoupled_synchronous` smoke script.

yhtang:
Silent-failure path is now handled by the 8B smoke and fail-fast propagation. The last completed smoke on `7904b68e` was run `26979827173` / job `79618351250`: logs include `Transferring weights to vLLM...`, `Weights transferred successfully!`, `Received 3 outputs`, and `EXIT_CODE=0`; sitrep was `1/1 passed`. Follow-up commit `869e17f7` only preserves cleanup exit status; CI is re-running for that head.

aybchan (approved):

PR #2143, opened 2026-06-03
JAX inference offloading update vLLM, update GKE action

aybchan:


.github/workflows/jax-vllm-offloading.yml
```diff
@@ -95,7 +95,7 @@ jobs:
             ARCHITECTURE: arm64
             ARTIFACT_NAME: artifact-jio-build
             BADGE_FILENAME: badge-jio-build
-            BASE_IMAGE: nvcr.io/nvidia/cuda-dl-base:25.06-cuda12.9-devel-ubuntu24.04
```
aybchan, line 98: This was overriding the dockerfile BASE_IMAGE arg

PR #2145, opened 2026-06-04
[jax-inference-offloading] add 8B decoupled sync smoke

yhtang:
## Summary
- Add the Llama 3.1 8B JIO parameter mapping used by the current examples.
- Let the current single-controller and decoupled examples use an explicit response topic instead of relying only on KV discovery.
- Add a GKE smoke job for the validated 8B decoupled synchronous example.

## Base and retargeting
This PR targets `jio-update-vllm-0.22` while #2141 is open, because the 8B smoke builds on that JIO/vLLM stack update. Once #2141 has merged, this PR shall be retargeted to `main`.

## Validation
- Ptyche allocation, image `gitlab-master.nvidia.com:5005/yuhangt/containers:20260603-jio-update-vllm-0.22`: `decoupled_synchronous/decoupled_sync.sh` completed one 8B iteration with 2 vLLM GPUs + 2 JAX GPUs, transferred weights, and returned 3 outputs.
- Confirmed the old legacy multihost entrypoints referenced by the disabled transfer/GRPO/EKS jobs are absent from the current image, so this reincarnates only the current supported example path.
- Did not add async CI: the async 8B probe dispatched a batch but did not receive rollout results before the 900s timeout.
- Local checks: `python3 -m py_compile` for touched example Python files, `python3 -m json.tool` for the 8B mapping, workflow YAML parse, and `git diff --check`.

### DIFF
--- a/.github/workflows/jax-vllm-offloading.yml
+++ b/.github/workflows/jax-vllm-offloading.yml
@@ -264,6 +264,40 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         NVCR_TOKEN: ${{ secrets.NVCR_TOKEN }}
 
+  jio-decoupled-sync-8b-gke-smoke:
+    needs: amd64
+    runs-on: gke-a3mega
+    steps:
+    - uses: actions/checkout@v6
+
+    - name: Run JIO decoupled sync 8B smoke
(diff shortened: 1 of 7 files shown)


yhtang:
Closing and replacing with #2146, which uses the branch pushed directly to NVIDIA/JAX-Toolbox rather than the fork branch.

PR #2147, opened 2026-06-04
JAX offloading - HuggingFace fixes

aybchan:


yhtang (approved):
Approved. CI is green and the PR cleanly stacks onto #2141; the 8B smoke now reports a real pass instead of a silent failure.

PR #2180, opened 2026-06-24
Revert "Temporary workaround for the nccl cxx error"

Steboss:
Reverts NVIDIA/JAX-Toolbox#2177


olupton (approved):
Why isn't the CI running?

PR #2182, opened 2026-06-25
Add Fern docs platform: scaffolding, CI/CD, ecosystem diagram, and new pages

zhenghax:
## Summary

- Adds [Fern](https://buildwithfern.com/) documentation scaffolding (`fern/docs.yml`, `fern/fern.config.json`) with custom domain support
- Adds interactive ecosystem stack diagram (`docs/_components/EcosystemDiagram.tsx`, `stack.ts`) showing the JAX-on-NVIDIA-GPU software stack
- Adds new documentation pages: introduction, getting-started guides, environment variables, cloud providers, container versions, FAQ, resources, staging containers, and JAX inference offloading
- Adds three CI/CD GitHub Actions workflows: Fern docs CI, preview deploys, and production publish
- Refines existing docs (GPU performance, AXLearn, MaxText, resiliency examples) and cleans up broken links
- Adds `Makefile`, `.nvmrc`, `.npmrc`, `package.json` for Node/Fern toolchain

## Test plan
- [ ] Fern docs build locally via `make docs` (see `fern/README.md`)
- [ ] CI workflow (`fern-docs-ci.yml`) passes on branch
- [ ] Preview deploy workflow triggers on PR
- [ ] Ecosystem diagram renders correctly in docs site
- [ ] All internal links resolve

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2200, opened 2026-07-02
Add Fern docs platform: scaffolding, CI/CD, ecosystem diagram, and new pages

aybchan:
https://github.com/NVIDIA/JAX-Toolbox/pull/2182


.github/workflows/fern-docs-ci.yml
```diff
@@ -0,0 +1,33 @@
...
+
+on:
+  push:
+    branches:
+      - "pull-request/[0-9]+"
+  pull_request:
+    paths:
+      - 'docs/**'
+      - 'fern/**'
+      - 'package.json'
```
olupton, line 11: This seems like a strange mix? Should the `paths` be under `push` so this works after `copy-pr-bot` has run and works on PRs from forks? Or should this one be just on `pull_request` because it doesn't need `FERN_TOKEN` and therefore should just work on PRs from forks?

.github/workflows/fern-docs-preview.yml
```diff
@@ -0,0 +1,55 @@
...
+
+on:
+  push:
+    branches:
+      - "pull-request/[0-9]+"
+  pull_request:
+    paths:
+      - 'docs/**'
+      - 'fern/**'
+      - 'package.json'
```
olupton: Won't this still fail when someone opens a PR from a fork?

.gitignore
```diff
@@ -0,0 +1,15 @@
...
+# Fern build output & cache
+.fern/
+.preview/
+
+# OS / editor
+.DS_Store
+
+# Local working files (not for commit)
+docs/ecosystem/*-outline.md
+guides/
```
olupton: This doesn't look like it belongs in `main`.

README.md
```diff
@@ -1,20 +1,91 @@
...
+- **Guides**: Practical guides for profiling, optimizing, scaling and third-party tool integrations in JAX.
+
+- **Resources**: Talks, presentations and ecosystem references
+
+- **Container images**: Nightly builds, monthly NGC releases, and staging containers for JAX on GPU projects
+
+- **Experimental projects**: Experimental projects for developers to evaluate and provide feedback on
+
+## Latest news
+- [**NGC JAX 26.05**](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/jax/tags?version=26.05-py3) container image released
```
olupton, line 39: 26.06 is out

olupton (dismissed):

aybchan:
/ok to test [b1f5c2e](https://github.com/NVIDIA/JAX-Toolbox/pull/2200/commits/b1f5c2ee9442fa8eb10254d6f7004b5ebad8d51d)

aybchan:
/ok to test [edde4eb](https://github.com/NVIDIA/JAX-Toolbox/pull/2200/commits/edde4eb0e583d23cddb5ca7bbb3ca05a4b2a7b3b)

.github/workflows/fern-docs-ci.yml
aybchan: Updated to run on arbitrary pull requests

.github/workflows/fern-docs-preview.yml
aybchan: Updated to run only on copy-pr-bot branches

aybchan:
/ok to test [251c7f8](https://github.com/NVIDIA/JAX-Toolbox/pull/2200/commits/251c7f8f30e6658d8e52eb16646cf089af504238)

.github/workflows/fern-docs-preview.yml
```diff
@@ -0,0 +1,86 @@
+# Generates a shareable Fern preview URL for every docs PR and posts it as a comment.
+# Requires the FERN_TOKEN repository secret.
+name: Fern Docs Preview
+
+on:
+  push:
+    branches:
```
olupton, line 7: Why do we need `PREVIEW_PATHS` instead of using `paths` here?

olupton (approved):
LGTM, just one part I didn't follow

.github/workflows/fern-docs-preview.yml
aybchan: There was a failure mode I encountered where on first creation of the copy-pr-bot branch from an existing commit, there is no push diff detected by the workflow to filter on. This works around that by enforcing a diff against main.

olupton: I guess that should be
> Pushes to new branches: A two-dot diff against the parent of the ancestor of the deepest commit pushed.

from https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#git-diff-comparisons -- I wonder why it didn't work. I don't think this is a blocker, but I wonder if we can drop the workaround once this is in.

PR #2205, opened 2026-07-06
triage: build plugins and metric-based triage

olupton:
- Add `plugin` backend that receives a container to test.
- Add metric-based triage, where a numeric value is extracted from each test and classified as belonging to a 'pass' or 'fail' population.

### DIFF
--- a/.github/triage/jax_toolbox_triage/args.py
+++ b/.github/triage/jax_toolbox_triage/args.py
@@ -98,6 +98,16 @@ def parse_args(args=None) -> argparse.Namespace:
             `date` will be substituted, e.g. ghcr.io/nvidia/jax:{container}-{date} for
             the JAX-Toolbox public nightlies.""",
     )
+    parser.add_argument("--metric-name", type=str, help="Example: tflops_per_sec")
+    parser.add_argument(
+        "--passing-metric",
+        action="append",
+        type=float,
+        help="Good value for the metric",
+    )
+    parser.add_argument(
+        "--failing-metric", action="append", type=float, help="Bad value for the metric"
+    )
     parser.add_argument(
         "--output-prefix",
         default=None,
@@ -316,6 +326,14 @@ def parse_args(args=None) -> argparse.Namespace:
         default="main",
         help="The name of the main branch (e.g. main) to derive cherry-picks from",
     )
+    parser.add_argument(
+        "--container-registry",
+        type=str,
+        help="""
+            Remote registry for use with the plugin backend. This can optionally
+            include a tag prefix, i.e. gitlab.com/USER/containers and
+            gitlab.com/USER/containers:PREFIX- are both valid.""",
+    )
     args = parser.parse_args(args=args)
     if args.restart:
         if args.output_prefix is None:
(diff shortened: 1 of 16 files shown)


.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -462,92 +559,165 @@ def _build_and_test(
...
+                    metrics={},
+                )
+        summary = {
+            "build_time": build_time,
+            "container": container_name,
+            "output_directory": out_dir.as_posix(),
+            "result": str(test_result.result),
+            "test_repetition": test_repetition,
+            "test_time": test_result.time,
+        }
```
Steboss, line 724: We need to add metrics here, as with `--restart` we're now expecting to have `metrics` in the `summary`. So this should be: 
```
summary = {
    "build_time": build_time,
    "container": container_name,
    "output_directory": out_dir.as_posix(),
    "result": str(test_result.result),
    "test_repetition": test_repetition,
    "test_time": test_result.time,
    "metrics": test_result.metrics,
}
```

.github/triage/jax_toolbox_triage/args.py
```diff
@@ -331,12 +349,28 @@ def parse_args(args=None) -> argparse.Namespace:
...
+        args.container is not None
+        or args.start_date is not None
+        or args.end_date is not None
+    )
+    if args.metric_name is not None:
+        if container_search_options_passed:
+            raise Exception(
+                "--metric-name is only supported in version-level search; use --{passing,failing}-{container,versions}."
+            )
+
```
Steboss, line 383: Here we should continue the check, as, if I got it right, `--metric-name` requires at least `--passing-metric` and `--failing-metric`. So we could have: 
```python 
    if args.metric_name is not None:
        if container_search_options_passed:
            raise Exception(
                "--metric-name is only supported in version-level search; use --{passing,failing}-{container,versions}."
            )
        if not args.passing_metric or not args.failing_metric: 
            raise Expection("--metric-name requiest --passing-metric and --failing-metric")
```

Steboss: and likewise a check on `args.passing_metric` and `args.failing_metrics` to check whether `metric-name` was given

.github/triage/jax_toolbox_triage/args.py
```diff
@@ -331,12 +349,28 @@ def parse_args(args=None) -> argparse.Namespace:
...
         "local",
+        "plugin",
     }, args.container_runtime
+
+    # Metric-based triage checks
+    container_search_options_passed = (
+        args.container is not None
+        or args.start_date is not None
+        or args.end_date is not None
+    )
```
Steboss, line 367: Just a question. 
We're not allowing a similar run: 
```bash 
jax-toolbox-triage --container-runtime plugin --container ghcr.io:my_container ...
```
As in `container_factory.py` at the moment we're allowing only `docker` and `PyxisContainer` but not `plugin`. Is this right?

Steboss (dismissed):

.github/triage/jax_toolbox_triage/args.py
olupton: `--container` is a questionable name with hindsight; it's only used as part of the date-to-container-URL transformation in the container-level search. i.e. it is just `jax` or `maxtext` or whatever, to distinguish between the different nightly containers sharing a given date.

I think that today the `plugin` backend is probably supported everywhere, although I only tested it with version-level search:
- `--passing-container` and `--failing-container`
- `--passing-container` and `--failing-versions`
- `--failing-container` and `--passing-versions`

It isn't handled in `container_factory.py` because the `plugin` backend is defined as "use Docker locally to build, examine and push containers but defer to the plugin to launch tests using those containers and gather their results.

What isn't supported is metric-based triage in the container-level search. It would be quite easy to support this for closed search ranges (i.e. `--start-date` and `--end-date` both passed), but with an open range it would be harder; I didn't implement a classifier that would work without being seeded by both known-good and known-bad metric values.

.github/triage/jax_toolbox_triage/args.py
olupton: Added

.github/triage/jax_toolbox_triage/triage_tool.py
olupton: Done. Seems like this could do with being overhauled into a more straightforward TestResult <-> JSON pair of converters.

Steboss (approved):

PR #2245, opened 2026-07-30
Improve build-te for build-jax step in CI

Steboss:


Steboss:
/ok to test https://github.com/NVIDIA/JAX-Toolbox/pull/2245/commits/3448da0f5cd0afa276fc147f6bd50b0cb1b4c64f

Steboss:
/ok to test https://github.com/NVIDIA/JAX-Toolbox/pull/2245/commits/3448da0f5cd0afa276fc147f6bd50b0cb1b4c64f

Steboss:
/ok to test https://github.com/NVIDIA/JAX-Toolbox/pull/2245/commits/6285bfc758cf2022e6b8ef5d635a2ea54f12e9e9

Steboss:
/ok to test https://github.com/NVIDIA/JAX-Toolbox/pull/2245/commits/cb1317dbc55b6d4b9df830398cca432c6b511e76

.github/container/install-compiler-cache.sh
```diff
@@ -0,0 +1,160 @@
+#!/bin/bash
```
Steboss, line 1: References: 
- [PyTorch issue #190832](https://github.com/pytorch/pytorch/issues/190832)
- [sscache PR #2722](https://github.com/mozilla/sccache/pull/2722)
- [PyTorch commit 225ab0d](https://github.com/pytorch/pytorch/commit/225ab0df028a41b741a8c6f3c16a06fbdb55b14b)
- [PyTorch patch for sscache](https://raw.githubusercontent.com/pytorch/pytorch/225ab0df028a41b741a8c6f3c16a06fbdb55b14b/.ci/docker/common/patches/sccache-nvcc-13.3-dryrun-parsing.patch)

.github/container/build-te.sh
```diff
@@ -148,25 +162,107 @@ subprocess.run(
...
 )
 EOF
+# Transformer Engine adds NVTE_CCACHE_BIN as both its C++ and CUDA CMake
+# compiler launcher when NVTE_USE_CCACHE is set. Do not also wrap CXX: that
+# would produce a recursive "ccache ccache g++" invocation.
 if [[ "${CCACHE}" == "1" ]]; then
-    # Install ccache if not present (needs >= 4.1 for Redis remote storage support)
-    if ! command -v ccache &> /dev/null; then
-        apt-get update && apt-get install -y --no-install-recommends ccache
+    NVTE_CCACHE_BIN="${NVTE_CCACHE_BIN:-ccache}"
```
Steboss, line 155: If we want to use sscache, then we need the install script

.github/container/build-te.sh
```diff
@@ -148,25 +162,107 @@ subprocess.run(
...
+    if ! command -v "${NVTE_CCACHE_BIN}" &> /dev/null; then
+        echo "Compiler cache installation did not provide ${NVTE_CCACHE_BIN}"
+        exit 1
     fi
-    export CXX="ccache g++"
+
     export NVTE_USE_CCACHE=1
-    ccache --zero-stats
+    export NVTE_CCACHE_BIN
+    case "${CACHE_PROGRAM}" in
```
Steboss, line 178: Here we're initializing the compiler-cache backend.

 For sccache, we explicitly start the daemon, disable its idle shutdown for the
  duration of the long CUDA build (line 216)

.github/container/build-te-with-sccache.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
```
Steboss, line 1: `build-te-with-sccache.sh` is a thin BuildKit-specific wrapper around
  `build-te.sh`. It keeps Docker secret handling separate from the normal
  Transformer Engine build logic.

  When `ENABLE_NVTE_SCCACHE=1`, the wrapper:

  - Verifies that the temporary AWS and S3 BuildKit secret files are present.
  - Reads them without command tracing and exports them only to the current
    process tree.
  - Configures sccache to use TLS and an architecture-specific S3 prefix.
  - Adds `--ccache` to the `build-te.sh` invocation.

.github/workflows/_ci.yaml
```diff
@@ -67,14 +67,16 @@ jobs:
   build-jax:
     needs: build-base
     runs-on: ${{ inputs.ARCHITECTURE == 'amd64' && 'linux-amd64-cpu32m' || 'linux-arm64-cpu32m' }}
+    env:
+      CACHE_AWS_REGION: ${{ vars.AWS_REGION || secrets.AWS_REGION }}
```
olupton, line 71: Why?

.github/workflows/_ci.yaml
```diff
@@ -153,6 +155,8 @@ jobs:
         - name: Build JAX container
           id: build-jax
           uses: ./.github/actions/build-container
+          env:
+            SCCACHE_BUCKET: ${{ secrets.BAZEL_REMOTE_BUCKET }}
```
olupton, line 159: Why 4 entries in `mealkit-secret-envs` and 1 here?

.github/container/install-compiler-cache.sh
olupton: 0.17.0 is released, can this all be dropped?

olupton (commented):
I don't see a CI run from the latest diff? It seems like a lot of code and branches. Do we really need to support both ccache and sccache?

.github/container/install-compiler-cache.sh
Steboss: Not yet, I tried to get this script in a better shape though. 
The reason is that `sccache` 0.17.0 was released before the CUDA 13.3 fix, so there is no prebuilt `sccache` binary that contains the CUDA 13.3 fix yet

.github/workflows/_ci.yaml
Steboss: Old test, going to remove this bit

.github/workflows/_ci.yaml
Steboss: The env creates and env variable for `build-jax`, and it does not need the `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`, as the OIDC action that we have is already automatically providing these variables. 
The `mealkit-secret-envs` is an explicit list of which existing runner environment variables we have  that may cross into the Docker build. Since then we have Docker running, we do nee dto pass thes evariable somehow,and this is the only way I found.

Steboss:
/ok to test https://github.com/NVIDIA/JAX-Toolbox/pull/2245/commits/95ab9f9accd5b24120d5f92e0a67ae6e4c4f13db

PR #2248, opened 2026-07-31
Update CUDA base to 26.07 13.3.1

aybchan:


olupton (dismissed):
Diff LGTM, we should update https://docs.nvidia.com/jax-toolbox/container-versions too.

olupton (approved):

PR #2273, opened 2026-08-17
Add jax-cudnn-frontend agent skill

jessicadesilva:
Adds a discovery-first procedural skill for wrapping, integrating, and validating cuDNN Frontend / CuTe DSL kernels from pure JAX, without PyTorch.

**Example usage:**

Using the `jax-cudnn-frontend` skill: wrap the cuDNN Frontend CuTe DSL block sparse attention kernels for inference and training from pure JAX, exposing the kernel configuration knobs so we can autotune per problem shape.

**Lives at:** `docs/agent-skills/jax-cudnn-frontend/` so it's discoverable by any AI coding agent or human.

**Covers:** environment discovery, locating the vendor's own kernel call site as the source of truth for tensor contracts, minimal standalone repro scripts, JAX wrapper validation, and a symptom-indexed debugging guide distilled from real integration failures.

### DIFF
--- a/docs/agent-skills/jax-cudnn-frontend/SKILL.md
+++ b/docs/agent-skills/jax-cudnn-frontend/SKILL.md
@@ -1,0 +1,175 @@
+---
+name: jax-cudnn-frontend
+description: Implement, integrate, debug, and validate cuDNN Frontend / CuTe DSL kernels from pure JAX (no PyTorch). Use when asked to wrap, call, port, or autotune a cudnn-frontend kernel — attention variants, GEMM fusions, or experimental csrc kernel classes — from JAX. Discovery-first — derives APIs and tensor contracts from the installed environment instead of trusting memory or online examples.
+---
+
+# cuDNN Frontend kernels from pure JAX
+
+You are integrating experimental GPU kernels whose APIs, tensor contracts, and
+supported configurations change between package versions. **Treat everything
+you think you know about these APIs as stale.** Every claim about a signature,
+shape, dtype, layout, or flag must be re-derived from the installed
+environment before you build on it. This skill is the procedure for doing
+that, plus the failure modes that cost days when the procedure was skipped.
+
+## Non-negotiable rules
+
+1. **Never call a kernel whose contract you have not read from installed
+   source.** Kernel names, docs pages, GitHub examples, and your prior
+   knowledge describe *some* version — not necessarily the installed one.
+2. **Buffer shapes come from the vendor's own allocations**, not from
+   signature comments or docs. Find where the installed package allocates
+   each buffer before calling the kernel itself (see rule of the Rosetta
+   stone, Phase 2). An under-allocated output is a silent buffer overflow
+   that corrupts *neighboring* tensors and surfaces as unrelated crashes
+   later — memcheck cannot see it.
+3. **Compilation and execution are not correctness.** A kernel can launch,
+   return, and pass `block_until_ready()` while producing garbage, writing
+   nothing, or corrupting memory that only a later op trips over. Only value
+   checks against an independent reference count as passing.
+4. **No PyTorch** in the implementation path unless the user asks for it.
+   PyTorch is permitted only as an optional cross-check oracle, clearly
+   labeled.
+5. **Fresh process per meaningful measurement.** After any illegal memory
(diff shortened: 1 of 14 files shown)


docs/agent-skills/jax-cudnn-frontend/SKILL.md
```diff
@@ -0,0 +1,175 @@
...
+`cute.compile`). The A/B between them is your primary tool for separating
+bridge bugs from kernel bugs from your bugs. Mechanics, conventions, and
+version-sensitive behaviors to verify: `references/jax-integration.md`.
+
+### Phase 5 — Minimal standalone proof
+Before any notebook or abstraction: a single plain-Python script that runs
+the kernel once with fixed seeds, prints **value-based** evidence (NaN
+counts, checksums, fixed-position samples), and exposes every configuration
+choice as an env-var toggle. This script is simultaneously your repro for
+bug reports and your bisection harness. Allocate outputs with `jnp.empty` so
```
Steboss: A little correction, from JAX 0.11 `jnp.empty` may not be reliable as a NaN check - they say it may retrn genuinely uninitialized memory. Thus, an unwritten tile can contain any finite value and evade the `jnp.isnan`. 
We could say to use `jnp.full(..., jnp.nan)`, wdyt? could it be useful for this case?

docs/agent-skills/jax-cudnn-frontend/references/jax-integration.md
```diff
@@ -0,0 +1,137 @@
...
+
+**JIT.** Wrap the calling function in `jax.jit` (config choices that alter
+shapes go in `static_argnums`). Confirm integration by checking
+`jax.make_jaxpr(...)` shows the op, and — more importantly — by bit-comparing
+the jitted path's output against the direct path (below).
+
+## Keep the direct path alive
+
+Alongside the JAX-native wrapper, maintain a `cute.compile`-style direct
+invocation of the same kernel instance (concrete arrays via the DSL's
```
Steboss, line 77: I am not an expert on dlpack, but if I see from [JAX doc](https://docs.jax.dev/en/latest/_autosummary/jax.dlpack.from_dlpack.html) they say that external mutation of a DLPack-backed JAX array may lead to undefined behaviour.  Could this be a problem?

docs/agent-skills/jax-cudnn-frontend/references/jax-integration.md
```diff
@@ -0,0 +1,137 @@
...
+def op(x, meta):
+    return cjax.cutlass_call(
+        _launcher,
+        output_shape_dtype=[
+            jax.ShapeDtypeStruct(x.shape, x.dtype),          # out
+            jax.ShapeDtypeStruct(aux_shape, jnp.float32),    # aux — shape from
+        ],                                                   # vendor allocation!
+        softmax_scale=..., 
+    )(x, meta)
+```
```
Steboss, line 173: Was this example tested?

Steboss (commented):
I love this job, a few comments but overall ready for approval

docs/agent-skills/jax-cudnn-frontend/SKILL.md
jessicadesilva: Yes and thank you for that catch! I swapped for that and it also recommended a filler sentinel for the integer array case. We can remove that if we don't think it's necessary / wouldn't have a use case.

docs/agent-skills/jax-cudnn-frontend/references/jax-integration.md
jessicadesilva: Updated this example to be more "skeletal" and adjusted the wording to say instances of this structure have worked. (Most of this example came from the notebook I originally generated + tested.)

docs/agent-skills/jax-cudnn-frontend/references/jax-integration.md
jessicadesilva: Very handy that I was working on another project that touched on this! I'm not sure if the revision is the best, but I suggested changing the memory ownership to outside of JAX (export inputs and import outputs via dlpack). But that requires another dependency(?) like cupy, so if they don't want to do that then do the original thing with hints on how to avoid UB... but perhaps neither of these are good options?

PR #2275, opened 2026-08-18
Update the way we are reading exit code for jobset

Steboss:
[Here](https://github.com/NVIDIA/JAX-Toolbox/actions/runs/31776194253/job/94737553013) is an example of a false-success job on EKS. 
- The AF command fail because of Haiku version 
- This make an `exit 1` 
- However we do not have `-e` in the parent shell, which makes a non empty exit code
- The bash arithmetic treated this as a zero

To solve this error we can read directly the pod, and extract the exit code from there

### DIFF
--- a/.github/actions/eks-jobset/action.yml
+++ b/.github/actions/eks-jobset/action.yml
@@ -109,17 +109,20 @@ runs:
   - name: Set job command
     shell: bash -x -u {0}
     run: |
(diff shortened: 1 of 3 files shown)


Steboss:
/ok to test https://github.com/NVIDIA/JAX-Toolbox/pull/2275#:~:text=code%20for%20jobset-,fca06b2,-copy%2Dpr%2Dbot

olupton (approved):

PR #2288, opened 2026-08-25
docs: add Collective GEMMs in Tensor Sequence Parallelism blog post

shljessie:
## Summary

Adds a tech blog post on collective GEMMs (CGEMM) for tensor-sequence parallel training, and the MaxText + TransformerEngine integration that enables them.

The post covers why TPSP leaves reduce-scatter and all-gather collectives exposed at every layer boundary, how fusing the collective with the GEMM and chunking along the sequence dimension hides that latency behind compute, the `te_comm_gemm_overlap` flags (`MLP_ONLY` / `FULL` / `DISABLED`) and supported TE quantization recipes, and the results: 1008 → 1620 TFLOPS on Llama3-70B and 1201 → 1788 TFLOPS on Llama3-405B (61% and 49%) over an FSDP + TP baseline with overlap disabled.

- Adds `docs/blog/collective-gemms.md`, authored by Seonghee Lee, Deep Patel, Tejash Shah, Phuong Nguyen, and Abhinav Goel.
- Adds four figures to `docs/img/`: the benchmark chart, and diagrams for tensor parallelism, tensor-sequence parallelism, and the collective GEMM timelines.
- Adds a `Tech Blogs` section to `docs/index.yml` so the page is reachable in the Fern nav.
- Lists the post at the top of the `Tech blogs` section in the root `README.md`.

Placement note: existing tech blogs are hosted on developer.nvidia.com and only linked from the root `README.md`, so there was no in-repo precedent. This introduces `docs/blog/` rather than adding to `Performance & Profiling`, to keep blog-style writing separate from maintained reference docs. Happy to relocate if reviewers prefer it alongside the other performance pages.

The README entry points at `https://docs.nvidia.com/jax-toolbox/tech-blogs/collective-gemms`, which resolves once the docs site publishes from `main`. Unlike the other four entries, this post is hosted in the docs site rather than on developer.nvidia.com — if a developer.nvidia.com version is planned, the link should be repointed there.

## Still to do before publishing

- [ ] Measurement setup for the results table: GPU model and count, TP/SP degree, sequence length, per-device batch size, and container tag
- [ ] Resolve the precision discrepancy: the benchmark chart labels both runs "FP8 training", but the 1201 figure has also been described internally as MXFP8. These are different recipes (`te_fp8_currentscaling` vs `te_mxfp8`). The prose deliberately names neither until this is confirmed.
- [ ] Optional: re-export the benchmark chart at ~2400px wide. The current one is 824px and will look soft on HiDPI displays.
- [ ] Decide whether to keep the C2PA content-credential metadata embedded in the two SVGs (~7.7 KB each)
- [ ] Confirm whether DeepSeek-v3 (MoE) results should be included alongside the two dense models

## Test plan

- [x] `npm run docs:check` passes (0 errors; the 2 warnings are pre-existing and unrelated)
- [x] `fern docs dev` renders the page with no MDX parse errors
- [x] All four figures resolve and serve at full size on the hosted preview
- [x] All six external links verified reachable
- [x] README link matches the slug in the published sitemap
- [x] Baseline qualified as FSDP + TP in all four places the result is stated
- [x] Acknowledgements finalized
- [ ] Fern preview reviewed


shljessie:
@abhinavgoel95 @phu0ngng  Requesting final review before merge!

phu0ngng (dismissed):

phu0ngng (approved):

PR #2298, opened 2026-08-28
`--commit` feature to triage tool

Steboss:
When a `--commit` is provided, the tool:
- resolves the supplied commit and its first parent
- builds and tests the candidate from the failing endpoint
- tests the parent when the candidate reproduces the regression
- reports the candidate immediately when it fails and its parent passes
- falls back to standard version search in case the candidate is not confirmed and here we have: 
    - A failing candidate becomes the failing endpoint for a backward search
    - a passing candidate becomes the passing endpoint for a forward search
    - a build or test error retains the original bisection range

To use it: 
```bash 
jax-toolbox-triage --commit xla:<revision> --passing-container ghrc.io/something-pass --failing-container ghcr.io/something-fail -v ... --
```

### DIFF
--- a/.github/triage/jax_toolbox_triage/args.py
+++ b/.github/triage/jax_toolbox_triage/args.py
@@ -239,6 +239,18 @@ def parse_args(args=None) -> argparse.Namespace:
         type=parse_version_argument,
     )
     version_search_args.add_argument(
+        "--commit",
+        metavar="[PACKAGE:]REVISION",
+        help="""
+            Commit suspected of introducing the regression. The tool first rebuilds
+            and tests this commit and its first parent using the other package versions
+            from the failing endpoint. If the commit is not confirmed as the culprit,
+            its observed test result is used to narrow the normal version-level
+            bisection. A bare revision is auto-detected among the source repositories;
+            prefix it with a package name (for example jax:abc123) to disambiguate.
+        """,
+    )
+    version_search_args.add_argument(
         "--cherry-pick",
         default={},
         help="""
@@ -459,7 +471,7 @@ def parse_args(args=None) -> argparse.Namespace:
     else:
         # None of --{passing,failing}-{versions,container} were passed, make sure the
         # compulsory arguments for the container-level search were passed
-        assert args.container is not None, (
-            "--container must be passed for the container-level search"
-        )
+        assert (
+            args.container is not None
+        ), "--container must be passed for the container-level search"
     return args
--- a/.github/triage/jax_toolbox_triage/args.py
+++ b/.github/triage/jax_toolbox_triage/args.py
@@ -7,7 +7,6 @@ import os
 import pathlib
 import tempfile
 import warnings
-import typing
 
 # Software we know may exist in the containers that we might be able to triage
 # We know how to recompile JAX/XLA, so it's OK that they include C++ code
@@ -37,7 +36,7 @@ def parse_version_argument(s: str) -> dict[str, str]:
     return ret
 
 
-def parse_commit_argument(s: str) -> typing.Dict[str, str]:
+def parse_commit_argument(s: str) -> dict[str, str]:
     try:
         versions = parse_version_argument(s)
     except (AssertionError, ValueError) as error:
@@ -60,7 +59,7 @@ def parse_override_remotes(s: str) -> dict[str, str]:
         s: (str) e.g. https://<token>@host/repo.git
 
     Returns:
-        ret: (typing.Dict[str,str]) Dictionary with software as key and git-url as value.
+        ret: (dict[str, str]) Dictionary with software as key and git-url as value.
     """
     ret: dict[str, str] = {}
     for part in s.split(","):
--- a/.github/triage/jax_toolbox_triage/args.py
+++ b/.github/triage/jax_toolbox_triage/args.py
@@ -35,6 +35,26 @@ def parse_version_argument(s: str) -> typing.Dict[str, str]:
     return ret
 
 
+def parse_commit_argument(s: str) -> typing.Dict[str, str]:
+    # Most users of a bare --commit are triaging JAX itself. Keep that convenient,
+    # while reusing the existing package:version syntax for every other package.
+    try:
+        versions = parse_version_argument(s if ":" in s else f"jax:{s}")
+    except (AssertionError, ValueError) as error:
+        raise argparse.ArgumentTypeError("invalid --commit value") from error
+    if len(versions) != 1:
+        raise argparse.ArgumentTypeError("--commit accepts exactly one commit")
+    package, revision = next(iter(versions.items()))
+    if (
+        not package
+        or not revision
+        or revision.startswith("-")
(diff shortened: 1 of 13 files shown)


.github/triage/jax_toolbox_triage/args.py
```diff
@@ -35,6 +35,26 @@ def parse_version_argument(s: str) -> typing.Dict[str, str]:
...
+def parse_commit_argument(s: str) -> typing.Dict[str, str]:
+    # Most users of a bare --commit are triaging JAX itself. Keep that convenient,
+    # while reusing the existing package:version syntax for every other package.
+    try:
+        versions = parse_version_argument(s if ":" in s else f"jax:{s}")
+    except (AssertionError, ValueError) as error:
+        raise argparse.ArgumentTypeError("invalid --commit value") from error
+    if len(versions) != 1:
+        raise argparse.ArgumentTypeError("--commit accepts exactly one commit")
+    package, revision = next(iter(versions.items()))
```
olupton: Suggest not adding the JAX shortcut, I'm not sure it's a good heuristic (I think XLA is more common).

Also, I was anticipating that this would accept multiple versions and that it would be possible to pass it multiple times, because:
1. One use-case is passing in the outcome of a fuzzy top K guessing process, and maybe there are K>1 strong candidates
2. Another use-case is confirming whether one triage result (which is a vector of reference versions and one culprit) also explains a different failure. I realise that in this case there would need to be some extra information content, e.g. that `jax:A,xla:B` means that the hint is that A is the culprit and it and its parent should be tested in conjunction with B.

.github/triage/jax_toolbox_triage/args.py
```diff
@@ -238,6 +258,20 @@ def parse_args(args=None) -> argparse.Namespace:
...
         type=parse_version_argument,
     )
+    version_search_args.add_argument(
+        "--commit",
+        metavar="[PACKAGE:]REVISION",
+        type=parse_commit_argument,
+        help="""
+            Commit suspected of introducing the regression. The tool first rebuilds
+            and tests this commit, then its first parent if the failure reproduces,
+            using the other package versions from the failing endpoint. If the commit
```
olupton: I fear this will be fragile as it leads to larger version skews. Normally we only test JAX/XLA commits that are more or less contemporary (see bug 5978749), as failing to do so often leads to compilation failures.

I would suggest something more like plugging into the existing logic, where if `--commit` specifies a JAX commit then a contemporary XLA commit is taken. This also means that if the hint is wrong then we should be able to narrow the remaining search space in all packages.

.github/triage/jax_toolbox_triage/logic.py
```diff
@@ -1,19 +1,33 @@
...
 class CouldNotReproduceFailure(Exception):
-    pass
+    def __init__(
+        self,
+        message: str,
+        *,
+        outcome: typing.Optional["ClassifiedTestOutcome"] = None,
+    ) -> None:
+        super().__init__(message)
+        self.outcome = outcome
```
olupton: This seems redundant; we are putting the outcome both into a member variable and the type name? If we need to distinguish, maybe keep the member variable but fold the two classes into a single `CouldNotReproduceDesiredOutcome` exception?

.github/triage/jax_toolbox_triage/triage_tool.py
```diff
@@ -922,17 +1095,38 @@ def run_version_bisection(
...
                 classifier=classifier,
             )
+            if candidate_result is not None:
+                result, last_known_good, first_known_bad = candidate_result
+
+        # Run the ordinary version-level bisection unless direct validation converged.
+        if result is None:
+            self.logger.info("Running version-level bisection...")
+        try:
+            if result is None:
```
olupton: redundant?

.github/triage/jax_toolbox_triage/args.py
Steboss: Great point, I didn't consider it initially. 
Yes, I could make `--commit` with `append` action, so that we can pass: 
```bash 
--commit jax:A, xla:B, flax:C \
--commit xla:D, jax:E
```
where the first argument, `jax:A` and `xla:D` is the expected culprit commit, and the second argument, `xla:B` ,`flax:C`, and `jax:D` are the reference versions we want. We're then going to test for each `--commit` entry: 
```
jax:A with xla:B and flax:C
jax:A^ with xla:B and flax:C
```

.github/triage/jax_toolbox_triage/args.py
Steboss: 100% right

.github/triage/jax_toolbox_triage/args.py
```diff
@@ -239,6 +255,22 @@ def parse_args(args=None) -> argparse.Namespace:
         help="Deprecated alias for --passing-versions",
         type=parse_version_argument,
     )
+    version_search_args.add_argument(
+        "--commit",
+        metavar="PACKAGE:REVISION[,PACKAGE:REFERENCE...]",
+        action="append",
+        type=parse_commit_argument,
```
olupton:
```suggestion
        type=parse_version_argument,
```

.github/triage/jax_toolbox_triage/args.py
```diff
@@ -36,14 +36,30 @@ def parse_version_argument(s: str) -> dict[str, str]:
...
+        not package
+        or package.startswith("-")
+        or not revision
+        or revision.startswith("-")
+        for package, revision in versions.items()
+    ):
+        raise argparse.ArgumentTypeError("invalid --commit value")
+    return versions
+
+
```
olupton:
```suggestion
```
this seems unnecessary

docs/triage-tool.md
```diff
@@ -162,6 +169,12 @@ To use the tool, there are two compulsory inputs:
       * **Version Search between Containers**: `--passing-container` and `--failing-container`: a pair of URLs to containers to use in the version-level search; if these are passed then no container-level search is performed.
       * **Local Commit Search**: Use `--container-runtime=local` when you are already inside a JAX container. This mode skips all container orchestration and performs a version-level search directly in the local container. It requires you to specify the version range with `--passing-versions` and `--failing-versions`.
 
+Any of these scopes can be combined with `--commit REVISION` to check a suspected
+culprit first. A bare revision refers to JAX. For another source repository, use
+`--commit PACKAGE:REVISION` (for example, `--commit xla:abc123`). The ordinary scope
+arguments are still required because they provide the build environment and the
+fallback search range.
```
olupton:
```suggestion
Version-level search can be combined with `--commit PACKAGE:REVISION` to hint a probable culprit commit.
`--commit` may be passed multiple times to pass multiple hints.
If multiple packages are listed in the same instance (`--commit P1:R1,P2:R2`) then the 2nd and subsequent packages and revisions are interpreted as reference values; i.e. the hint is that `(R1, R2)` fails and `(parent of R1, R2)` passes.
The ordinary scope arguments are still required because they provide the build environment and the fallback search range.
```

